PR #3, opened 2018-06-07
Add a mechanism to lookup the AMI from current region

tanmng:
# 2018-06-07 9:45 EST

## Description

I added a feature to support using EKS optimized AMI by looking that up automatically using a variable which user can overwrite. Also, user can specify directly what AMI the user wants rather than having to use the mapping.

This added the benefit of forcing users to use this module only in EKS supported regions for now (`us-east-1` and `us-west-2`

I tested this using an AWS account and everything seemed to work as expected.

### Checklist


brandoconnor:
Going to close this one down as the implementation is [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/5/files). Your idea was important and influential; I appreciate your input!

PR #4, opened 2018-06-07
WIP: Tighter security group rules

tanmng:
# Improvement on security group

## Description

As per [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html), you can set up the Security groups for the control plane and worker node to have either minimum rules, or have more access according to _recommendation_ (ie. more relaxed rules and our instances more access).

I added 3 more variables to let the user specify which scenario they would want to use:

* `worker_node_allow_all_egress` (default `true`)
* `cp_to_wn_from_port` (default `1025`)
* `cp_to_wn_to_port` (default `65355`)

The default values of these correspond to the recommended case by AWS, but for any users who wish to set up a strict set of rules, they can provide different values to those variables, such as

* `worker_node_allow_all_egress = false`
* `cp_to_wn_from_port = 10250`
* `cp_to_wn_from_port = 10250`

Beside this, we also added the ability to provide a list of custom security groups that we would like to set onto the worker node. That will allow users to use other Terraform modules to set up the groups in the way that they like and then assign them onto the instance.

Since it's possible that user can set `worker_node_allow_all_egress = false`, we have to add one more security group rule resource to allow worker node egress to allow communication within the worker nodes themselves.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,25 @@ Full contributing [guidelines are covered here](https://github.com/terraform-aws
 Testing and using this repo requires a minimum set of IAM permissions. Test permissions
 are listed in the [eks_test_fixture README](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/eks_test_fixture/README.md).
 
+## Security group
+
+According to [AWS documentation on EKS security group](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html), you can set up the security group for worker nodes and control plane according to the minimum requirements or the recommended way.
+
+The module try to accommodate all possible settings that you might wish to have. Even including the ability to set additional security group onto the node and plane instances.
+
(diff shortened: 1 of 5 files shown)


brandoconnor:
I'd like this to be more flexible than just being able to control one or two specific variables of a rule or two. Giving the option for a module consumer to bring their own pre-defined security group to attach to either set of resources feels like a better approach. I'm working on this currently.

Thanks again on the idea even if the implementation isn't exactly what I think will be best for module consumers. The input is incredibly valuable! ⭐

tanmng:
@brandoconnor thanks for the feedback. I'll close this then

brandoconnor:
Thanks again @tanmng . I'll give you credit (because it's due!) in the changelog.

Issue #11, opened 2018-06-08
Bring your own security group


brandoconnor:
# I have issues...

## I'm submitting a

* [ ] bug report
* [x] feature request
* [ ] support request

## What is the current behavior

Currently the module only supports creation of the security groups for the cluster and workers from within the module itself. Some of the rules are 100% necessary and others are just commonplace and therefore useful. The rules rely on dynamic values but could be applied just the same to a security group passed to the module instead of created within the module. This would give flexibility to the module consumer to provide their own security group and a predefined set of rules that might be tighter than what the module currently prescribes.

## If this is a bug, how to reproduce? Please include a code sample

NA

## What's the expected behavior

The module should be able to accept a security group ID as input for both the cluster and workers with rules defined outside the module.

## Environment

* Affected module version: current (0.2.0)
* OS: All
* Terraform version: 0.11.x

brandoconnor:
This is complete in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/13

PR #13, opened 2018-06-11
Next developer team add worker groups

brandoconnor:
# PR o'clock

## Description

KOPS allows for creating multiple worker autoscaling groups each with different specifications (e.g. `instance_type`, asg size, userdata). This change allows for that same use case while still providing a default worker ASG if none are given. Security group IDs can also be provided for both workers and the cluster for better control over these aspects of the set up.

This finishes out the work of https://github.com/terraform-aws-modules/terraform-aws-eks/pull/12

While no more variable inputs are needed or any changed that are required, this is a breaking change as input for workers now takes on a new form.

### Checklist

- [x] Tests for the changes have been added and passing (for bug fixes/features) - I'd like to have 2 sets of tests here, one that tests a barebones instantiation of the module and another that provides the kitchen sink worth of extra options.


brandoconnor:
Also wraps up: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/11

PR #14, opened 2018-06-13
Add IAM role output

artursmet:
# PR o'clock

## Description

Added output for IAM role name that is automatically created for all workers within AutoScallingGoup.
That would allow granting additional permissions for the worker nodes, like using S3 buckets etc.

### Checklist

- [X] Docs have been added/updated (for bug fixes/features)
- [ ] Any breaking changes are noted in the description above


brandoconnor:
Hey @artursmet -

Many thanks for this. I'll be releasing again in the next day or so and this will become available.

PR #16, opened 2018-06-13
Add option to run worker nodes in a subset of eks subnets (e.g. only …

### LINKED ISSUE #15: Allow worker nodes to be created in private subnets if eks cluster has both private and public subnets
# I have issues

## I'm submitting a

* [ ] bug report
* [x ] feature request
* [ ] support request

## What is the current behavior
Based on [this guide](https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html) from aws, it is recommended that you specify both public and private subnets when creating your eks cluster, but that you only create your worker nodes in your private subnets. Current behaviour in this module will use the same subnets for creating the eks cluster as for placing the worker nodes within.

## What's the expected behavior
I believe it would be a good feature to add an additional (optional) list variable to the module called worker_subnets that will be used to create the worker nodes within. This means you can add private and public subnets to the `subnets` variable, but only add private subnets to the `worker_subnets` variable.

## Environment

* Affected module version:
* OS:
* Terraform version:

## Other relevant info
I have a branch with this feature on a fork, I will add a PR to be looked at.

MarkFreebairn:
…run workers in private subnets)

# PR o'clock

## Description

resolves #15 

The reason for this PR is to allow you to specify both private and public subnets when creating your eks cluster, but to only specify private subnets to create your worker nodes within (this is recommended in [this](https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html) AWS document).

The new variable added to achieve this is `worker_subnets`, it is an optional variable, current functionality will be unaffected if you do not use it. Below is an example of how using the new variable looks:
```
module "eks" {
  source       = "terraform-aws-modules/eks/aws"
  version        = "1.0.0"
  cluster_name   = "eks-test-1"
  subnets        = ["${module.vpc.public_subnets}", "${module.vpc.private_subnets}"]
  worker_subnets = ["${module.vpc.private_subnets}"]
  tags           = "${map("Managed", "terraform")}"
  vpc_id         = "${module.vpc.vpc_id}"
}
```

### Checklist

- [x ] `terraform fmt` and `terraform validate` both work from the root and `examples/eks_test_fixture` directories (look in CI for an example)
- [x ] Docs have been added/updated (for bug fixes/features)
- [ ] Any breaking changes are noted in the description above

I have not ran the test-kitchen tests as I am not currently able to, however I have created an eks cluster using this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ A full example leveraging other community modules is contained in the [examples/
 ```hcl
 module "eks" {
   source                = "terraform-aws-modules/eks/aws"
-  version               = "0.1.0"
+  version               = "1.0.0"
   cluster_name          = "test-eks-cluster"
   subnets               = ["subnet-abcde012", "subnet-bcde012a"]
   tags                  = "${map("Environment", "test")}"
@@ -100,6 +100,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | config_output_path        | Determines where config files are placed if using configure_kubectl_session and you want config files to land outside the current working directory.                                                                     | string |   `./`   |    no    |
 | configure_kubectl_session | Configure the current session's kubectl to use the instantiated EKS cluster.                                                                                                                                             | string |  `true`  |    no    |
 | subnets                   | A list of subnets to place the EKS cluster and workers within.                                                                                                                                                           |  list  |    -     |   yes    |
+| worker_subnets            | A list of subnets to place the EKS workers within. If empty then subnets variable will be used instead to place workers within.  |  list  |    []    |    no    |
 | tags                      | A map of tags to add to all resources.                                                                                                                                                                                   | string | `<map>`  |    no    |
 | vpc_id                    | VPC where the cluster and workers will be deployed.                                                                                                                                                                      | string |    -     |   yes    |
 | worker_groups             | A list of maps defining worker group configurations. See workers_group_defaults for valid keys.                                                                                                                          |  list  | `<list>` |    no    |
--- a/examples/eks_test_fixture/main.tf
+++ b/examples/eks_test_fixture/main.tf
@@ -48,10 +48,11 @@ module "vpc" {
 }
 
 module "eks" {
-  source        = "../.."
-  cluster_name  = "${local.cluster_name}"
-  subnets       = "${module.vpc.public_subnets}"
-  tags          = "${local.tags}"
-  vpc_id        = "${module.vpc.vpc_id}"
-  worker_groups = "${local.worker_groups}"
+  source         = "../.."
+  cluster_name   = "${local.cluster_name}"
+  subnets        = ["${module.vpc.public_subnets}", "${module.vpc.private_subnets}"]
+  worker_subnets = "${module.vpc.private_subnets}"
+  tags           = "${local.tags}"
+  vpc_id         = "${module.vpc.vpc_id}"
+  worker_groups  = "${local.worker_groups}"
 }
--- a/workers.tf
+++ b/workers.tf
@@ -4,7 +4,7 @@ resource "aws_autoscaling_group" "workers" {
   max_size             = "${lookup(var.worker_groups[count.index], "asg_max_size",lookup(var.workers_group_defaults, "asg_max_size"))}"
   min_size             = "${lookup(var.worker_groups[count.index], "asg_min_size",lookup(var.workers_group_defaults, "asg_min_size"))}"
   launch_configuration = "${element(aws_launch_configuration.workers.*.id, count.index)}"
-  vpc_zone_identifier  = ["${var.subnets}"]
+  vpc_zone_identifier  = ["${coalescelist(var.worker_subnets,var.subnets)}"]
   count                = "${length(var.worker_groups)}"
 
   tags = ["${concat(
--- a/variables.tf
+++ b/variables.tf
@@ -65,3 +65,9 @@ variable "worker_security_group_id" {
   description = "If provided, all workers will be attached to this security group. If not given, a security group will be created with necessary ingres/egress to work with the EKS cluster."
   default     = ""
 }
+
+variable "worker_subnets" {
+  description = "A list of subnets to place the EKS workers within. If empty then subnets variable will be used instead to place workers within."
+  type        = "list"
+  default     = []
+}


README.md
```diff
@@ -23,7 +23,7 @@ A full example leveraging other community modules is contained in the [examples/
 ```hcl
 module "eks" {
   source                = "terraform-aws-modules/eks/aws"
-  version               = "0.1.0"
+  version               = "1.0.0"
```
brandoconnor: Good catch! Hmm, does sourcing from the registry work without a version at all? kind of a burden to remember this on every release.

workers.tf
```diff
@@ -4,7 +4,7 @@ resource "aws_autoscaling_group" "workers" {
   max_size             = "${lookup(var.worker_groups[count.index], "asg_max_size",lookup(var.workers_group_defaults, "asg_max_size"))}"
   min_size             = "${lookup(var.worker_groups[count.index], "asg_min_size",lookup(var.workers_group_defaults, "asg_min_size"))}"
   launch_configuration = "${element(aws_launch_configuration.workers.*.id, count.index)}"
-  vpc_zone_identifier  = ["${var.subnets}"]
+  vpc_zone_identifier  = ["${coalescelist(var.worker_subnets,var.subnets)}"]
```
brandoconnor: IIRC, this is how the placement of autoscaling instances into subnets is determined. Wouldn't we want this to be just `var.worker_subnets` in that case?

README.md
MarkFreebairn: Yes it does, it will just grab the latest version if this is not included, I can remove this line altogether?

workers.tf
MarkFreebairn: The way that this works is that `coalescelist` will return the first non-empty list, therefore if the user passes in a `worker_subnets` list when creating the module it will not be empty so vpc_zone_identifier will be set to that list, otherwise, it will be set to the value of the `var.subnets` list. I wrote it like this to ensure that current functionality would not be broken (so anyone currently using this module can upgrade versions without it affecting them).

examples/eks_test_fixture/main.tf
```diff
@@ -48,10 +48,11 @@ module "vpc" {
...
-  source        = "../.."
-  cluster_name  = "${local.cluster_name}"
-  subnets       = "${module.vpc.public_subnets}"
-  tags          = "${local.tags}"
-  vpc_id        = "${module.vpc.vpc_id}"
-  worker_groups = "${local.worker_groups}"
+  source         = "../.."
+  cluster_name   = "${local.cluster_name}"
+  subnets        = ["${module.vpc.public_subnets}", "${module.vpc.private_subnets}"]
+  worker_subnets = "${module.vpc.private_subnets}"
```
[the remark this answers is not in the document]
MarkFreebairn: This can actually still be done using the above, `worker_subnets` is just a list so you could, for example, have:
`worker_subnets = ["var.my_public_subnet", "var.my_private_subnet", "subnet-abcde123"]` and all of these subnets would be used in the autoscaling group to place workers into.

README.md
brandoconnor: Yes, that'd be better. Thanks!

examples/eks_test_fixture/main.tf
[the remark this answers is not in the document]
brandoconnor: Hmm maybe I haven't explained myself well enough... Consider:

```
locals {
  worker_groups = "${list(
                  map("instance_type","t2.small",
                      "additional_userdata","echo foo bar",
                      "name", "foo",
                   ),
                  map("name", "bar",
                   ),
  )}"
```

The above arrangement of `worker_groups` gives me 2 distinct ASGs with workers of different size, name, and userdata, but with the code in your branch, there'd be no way to put `foo` ASG into strictly the public subnets while putting `bar` in strictly private subnets. The `worker_subnets` variable is currently shared amongst each of _n_  autoscaling groups if it's not split into a key/value like `name`, `instance_type`, etc. Does that make any better sense? Help me come around with an example if I'm missing something.

workers.tf
brandoconnor: Related to my comment above, if `worker_subnets` is a property of each `worker_group` list item, the `subnets` input would only apply to the cluster which means this could be yanked. In that case, the variable could probably stand to be renamed `cluster_subnets` for clarity. I'm less worried about making breaking changes right now as this module has only been live for a week. Bumping a major rev is sometimes necessary early on.

ozbillwang:
So should we call `node` more than `worker` to match kubernetes naming consistent?

reference: https://kubernetes.io/docs/concepts/architecture/nodes/

perryao:
@brandoconnor @MarkFreebairn I can help too if need be. Just lemme know

examples/eks_test_fixture/main.tf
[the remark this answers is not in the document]
ozbillwang: My option for this is, there should be no kubernetes worker nodes put in public subnets at all.  

Only public facing load balancers, Nat gateway, WAF (application firewalls), bastion/jump box or similars should be put in public subnets.

README.md
ozbillwang: since it will not possible to have this feature in version 1.0.0, we should update it to 1.2.0 now or others later.

Otherwise, the reader will be confused.

perryao:
@brandoconnor if I go ahead and add a public subnet to the cluster today, then later when this is merged restrict the workers to private subnets, do you think that would require a destroy/recreate of the entire eks cluster?

brandoconnor:
I'm not sure about that @perryao . In that case it'd probably be safer to create a second set of worker ASGs, balance the load across the entire cluster, and then destroy the original ASG.

ozbillwang:
@MarkFreebairn  && @brandoconnor 

I deploy a new EKS cluster with old version/release `v1.0.0` to private subnets, it works. 

So not sure if we still need this PR. Seems it makes the setting complex without big helps.

But I found another issue, when I apply eks cluster to private subnets, this module still applied `Public IPs` on nodes.

So I will recommend to close current PR and add feature (add a public_or_proviate = 1 or 0) to make sure, we don't apply EIP on nodes in private nodes. 

Let me know your thinking, and I can help to submit the PR.

brandoconnor:
@ozbillwang - I think your suggestion makes the most sense at this point. Promoting private subnets as the place for workers to land is the right choice for a k8s setup. You'll need to explain the `public_or_private` variable a little further (in a separate PR, please) to make that happen (naming is hard, we'll find something reasonable). Thanks again for all the ideas here but I'm going to close this down as we move forward on moving to only private subnets though other PRs.

Issue #27, opened 2018-06-22
Security group "workers_ingress_cluster" is very limiting


ilyasotkov:
Currently, in `workers.tf`, we have this security group:

```tf
resource "aws_security_group_rule" "workers_ingress_cluster" {
  description              = "Allow workers Kubelets and pods to receive communication from the cluster control plane."
  protocol                 = "tcp"
  security_group_id        = "${aws_security_group.workers.id}"
  source_security_group_id = "${local.cluster_security_group_id}"
  from_port                = 1025
  to_port                  = 65535
  type                     = "ingress"
  count                    = "${var.worker_security_group_id == "" ? 1 : 0}"
}
```

Basically, this setting makes it impossible for Kubernetes services to access pods that have have containerPort set to anything below 1025, which is a huge issue since so many of them use the 80 port (e.g. nginx). So, `from_port` should be set to 0, not 1025.

I realize this is copied from CloudFormation in the official EKS guide, so I'll also submit an issue there.

brandoconnor:
This is in discussion on #28 . Hopefully, you're good with what I've proposed there. Thanks for the contribution!

brandoconnor:
Shipping this change to a release now. Thanks again!

PR #28, opened 2018-06-22
Fix limiting behavior of workers_ingress_cluster security group

ilyasotkov:
* Changes workers_ingress_cluster security group to allow ingress connections on 0-65535 instead of 1025-65535
* See https://github.com/terraform-aws-modules/terraform-aws-eks/issues/27


hatemosphere:
to be honest, i don't think it's necessary since you can always attach your own SG with customized rules, current SG looks like an okay'ish container with sane defaults.

ilyasotkov:
@hatemosphere Is there a good security reason to not allow privileged ports?

Doing a quick search of https://github.com/kubernetes/charts for `containerPort` will show that significant number of applications use a port below 1025. Do we really want to make all of these application non-functional by default? That's the real question.

brandoconnor:
Hey @ilyasotkov . Thanks for the submission.

I tend to agree with @hatemosphere here - the project is adhering closer to docker defaults by only allowing this set of ports. Anything lower would require the `net_bind_service` capability or running the container as root - both recognized as loosening the security story of your container.

I would however be open to a variation on this submission. If the port lower bound were made to a variable that defaults to 1025, we'd have the best of both worlds: a sane/safe default and flexibility should we need to go outside of normal operation. Does that sound reasonable as an alternative @ilyasotkov ?

workers.tf
```diff
@@ -73,7 +73,7 @@ resource "aws_security_group_rule" "workers_ingress_cluster" {
   protocol                 = "tcp"
   security_group_id        = "${aws_security_group.workers.id}"
   source_security_group_id = "${local.cluster_security_group_id}"
-  from_port                = 1025
+  from_port                = 0
```
brandoconnor: Let's make this a variable instead and default it to 1025. Name it something like `worker_sg_ingress_from_port` and provide a suitable description.

ilyasotkov:
@brandoconnor Thanks for the explanation, you proposal is a totally reasonable alternative.

I made the suggested changes to the PR, please review and tell me if something out of place or missing.

brandoconnor:
Awesome! Merging now.

PR #44, opened 2018-06-27
adding capability to use non-default aws profile

### LINKED ISSUE #39: AWS Profile in kubeconfig template
## I'm submitting a

* [x] feature request

For my current delivery, the customer has credentials with multiple profiles and not only needs to specify different profiles per cluster, but has no `default` profile.

It would be great if the `kubeconfig.tpl` could be modified:

```yaml
...
users:
- name: aws
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: heptio-authenticator-aws
      args:
        - "token"
        - "-i"
        - "${cluster_name}"
      env:
        - name: AWS_PROFILE
          value: ${aws_profile}
```

where the default Terraform value to populate the template would be `default` to ensure no regression.

I wanted to start a discussion before a PR to ensure best path forward on this. Thanks!

mschuchard:
# PR o'clock

## Description

Fixes #39

### Checklist

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -51,6 +51,7 @@ data template_file kubeconfig {
     cluster_name        = "${var.cluster_name}"
     endpoint            = "${aws_eks_cluster.this.endpoint}"
     region              = "${data.aws_region.current.name}"
+    aws_profile         = "${var.aws_profile}"
     cluster_auth_base64 = "${aws_eks_cluster.this.certificate_authority.0.data}"
   }
 }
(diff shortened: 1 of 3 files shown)


sdavids13:
Would it make more sense to try to make the kubeconfig.tpl more generic or perhaps even allow clients to pass it in? For example, I would like to change the cluster name to be the name of the EKS cluster. Additionally, I assumed an IAM role to run Terraform and would need the `-r "roleArn"` added to the exec args list (I do know that this `AWS_PROFILE` would do the trick for that but would prefer to not have to set it up independently).

brandoconnor:
Even still I appreciate the contribution @mschuchard . Thanks for the submission. I merged the other PR which should fill your use case just as well. 👍

PR #49, opened 2018-06-30
Allow Kubeconfig to be more configurable

sdavids13:
# PR o'clock

## Description

Making Kubeconfig significantly more configurable in the following ways:
  * Setting the cluster name to the same value as the EKS cluster (so cluster name doesn't conflict with the old `kubernetes` naming)
  * Allowing the user name to be modified
  * Allowing the context name to be modified
  * Allowing additional heptio parameters to be specified (useful to be able to specify a role that should be assumed)
  * Allowed environment variables to be specified for the heptio execution, this provided a generic implementation compated to #44.

### Checklist


variables.tf
```diff
@@ -78,3 +78,28 @@ variable "worker_sg_ingress_from_port" {
   description = "Minimum port number from which pods will accept communication. Must be changed to a lower value if some pods in your cluster will expose a port lower than 1025 (e.g. 22, 80, or 443)."
   default     = "1025"
 }
+
+variable "kubeconfig_context_name" {
+  description = "Name of the kubeconfig context."
+  default     = "aws"
```
max-rocket-internet: I think this should default to the cluster name as `aws` is very uninformative.

sdavids13: Yea, I do agree (I configured my context name to be the cluster name as well). I was trying to limit the amount of changes made by default but if others think that is a more reasonable default I would be happy to provide another PR to fix it.

sdavids13: Also, I just found that if you are trying to merge various EKS kubeconfigs together via the `KUBECONFIG` environment variable the username needs to be unique across all of the config files. So, I will go ahead and make the change for both the context name and the user name.

brandoconnor: That'd all be appreciated @sdavids13 . Thanks for driving that

max-rocket-internet:
> username needs to be unique across all of the config files

Correct.

I've made a PR: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/52

PR #54, opened 2018-07-09
Splitting aws-auth and kubectl related resources as they are not related

max-rocket-internet:
Hi again,

Here I'm splitting the resources related to kubectl/kubeconfig and aws-auth as these are unrelated to each other. For example, the `aws-auth` config map has nothing to do with configuring a kubectl session. So rather than have a single variable `configure_kubectl_session` for both, I've added a variable for each.


brandoconnor:
Hey again! Back for more? 🥇 

I think we've managed to avoid a breaking change here though variable changes are often exactly that. Effectively, this splits out aws_auth from kubectl but also introduces a variable to control if we want to write the config file but not use it. I'll resolve the conflict and get this submitted. The value here is small but also given sane/non-disruptive defaults, the required interface doesn't actually grow, which is in part what I'm steering toward..

PR #55, opened 2018-07-09
Formatting and minor cosmetic fixes

max-rocket-internet:
Just some minor changes in README and resource naming.

1. The fields of `type` and `default` in `Inputs ` in the README are a little inconsistent. e.g. booleans are called strings and some types don't match their defaults. So I've made the type reflect the actual variable type and where the defaults are `[]`/`""`/`{}` I've just made the default as ``. Does that make sense?

2. Renamed `aws_eks_cluster.this` to `aws_eks_cluster.cluster`. `this` seems like an odd name for a resource, no? I think `cluster` matches the other resources better.

3. Rather than using `${var.cluster_name}` all over the place, just use it once on the `aws_eks_cluster.cluster` resource, then refer to the `name` attribute of that resource elsewhere. I think it's tidier?

Feel free to disagree with the above points, they are more cosmetic than anything else 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,3 +126,4 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | worker_iam_role_name | IAM role name attached to EKS workers |
 | worker_security_group_id | Security group ID attached to the EKS workers. |
 | workers_asg_arns | IDs of the autoscaling groups containing workers. |
+
--- a/README.md
+++ b/README.md
@@ -98,18 +98,18 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | cluster_security_group_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingres/egress to work with the workers and provide API access to your current IP/32. | string | `` | no |
 | cluster_version | Kubernetes version to use for the EKS cluster. | string | `1.10` | no |
 | config_output_path | Determines where config files are placed if using configure_kubectl_session and you want config files to land outside the current working directory. | string | `./` | no |
-| configure_kubectl_session | Configure the current session's kubectl to use the instantiated EKS cluster. | string | `true` | no |
-| kubeconfig_aws_authenticator_additional_args | Any additional arguments to pass to the authenticator such as the role to assume ["-r", "MyEksRole"] | string | `<list>` | no |
+| configure_kubectl_session | Configure the current session's kubectl to use the instantiated EKS cluster. | boolean | `true` | no |
+| kubeconfig_aws_authenticator_additional_args | Any additional arguments to pass to the authenticator such as the role to assume ["-r", "MyEksRole"] | list | `` | no |
 | kubeconfig_aws_authenticator_command | Command to use to to fetch AWS EKS credentials | string | `heptio-authenticator-aws` | no |
 | kubeconfig_aws_authenticator_env_variables | Environment variables that should be used when executing the authenticator i.e. { AWS_PROFILE = "eks"} | string | `<map>` | no |
 | kubeconfig_name | Override the default name used for items kubeconfig. | string | `` | no |
 | subnets | A list of subnets to place the EKS cluster and workers within. | list | - | yes |
-| tags | A map of tags to add to all resources. | string | `<map>` | no |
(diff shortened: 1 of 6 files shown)


bryantbiggs:
hey @max-rocket-internet - just a few questions/comments.

1. I see you changed the `configure_kubectl_session` in the docs to type boolean instead of string - is this wise since HCL does not have true (no pun intended) boolean types? They are interpreted as strings `"true"`|`"false"` which are further interpreted as `"1"`|`"0"` - I know this has tripped me up in the past as putting a `true` instead of `"true"` will throw an error

2. From the "official" modules I have seen supported, I think `this` is a convention that is used. See [here](https://github.com/terraform-aws-modules/terraform-aws-vpc)

3. (Personal opinion) I think its better to have it as it was with `${var.cluster_name}`. Seems odd to declare a variable only to use it once then use resource references everywhere else; you have a nested lookup as well (for developers, not for the interpreter)

brandoconnor:
Re: point 2 @bryantbiggs - I'm with you on this but while the EKS module here, the VPC module and my other ALB module all use `this`, each of these projects are community driven efforts and I don't think there's a formal Hashicorp standard around these names. I've seen the practice of `this`/`main` enough times that it seems to be the closest to convention. Also, changing the resource name would be a breaking change for all current users.

What I like about `this` or `main` is the dropping of the duplication of the resource name when referencing its outputs. `aws_eks_cluster.cluster.foo` feels like the worst of the options where `aws_eks_cluster.this.foo` or `aws_eks_cluster.main.foo` feels a little better if you come from the module philosophy of centering a module around a single core resource (with n ancillary resources surrounding it).

To the other bits, I'll have a closer look in a bit. Thanks for the continued efforts @max-rocket-internet !

brandoconnor:
Also, ditto on the first point by @bryantbiggs . Booleans aren't a valid variable input type iirc though strings of boolean values are interpreted as such.

max-rocket-internet:
1. I know *technically* there is no boolean type in TF but cosmetically there is.  e.g. `force_destroy = true` is valid but `force_destroy = yes` is not. The unquoted `true` is treated different to other unquoted strings and is shown as a different colour in syntax highlighting. But I checked and I see the convention is to call them all strings. I'll revert the boolean part for sure.

About other types, I checked a few modules and can't work out the convention. It seems mixed. If you have a non-required map variable, e.g. `workers_group_defaults`, what is the type? We have `map` but for `tags` variable we have string. That's inconsistent, right?

2. I honestly thought `this` was a typo when I saw it but OK, I see it's a convention, I'll revert that 🙂

3. I stand by this one. The relationship between many resources and the `aws_eks_cluster` is the `name` attribute of the `aws_eks_cluster`. And this relationship cannot change. e.g. in the `userdata` for workers or the `kubernetes.io/cluster/*` tags. In my mind it's like a tree relationship: `var.cluster_name` > `aws_eks_cluster` > many resources.

> Seems odd to declare a variable only to use it once then use resource references everywhere else

I would disagree with this completely. Resources references are one of the best things about TF and are also how it automatically manages dependencies.

> you have a nested lookup as well

Where?

But maybe this is all nitpicky, happy to leave this unchanged 🙂

bryantbiggs:
my apologies @max-rocket-internet  I should have been more clear. What I meant by the nested lookup is that the user of the module would lookup where the cluster name originates from to then find a variable and have to further lookup where that is being set. Its a personal preference of my own (can totally be disregarded), I just don't see a benefit of assigning the cluster name as a variable for the cluster resource itself and then using the cluster name lookup everywhere else. I know the Hashicorp docs/EKS demo use the prior method (`"${var.cluster_name}"`) as opposed to the latter method you have proposed. All a matter of opinion

brandoconnor:
Thanks both of you for the spirited discussion here. 😃  I certainly appreciate any cleanup people can give and typing the untyped just serves to make variables more explicit. That's well and good by me.

Regarding `var.cluster_name` and `aws_eks_cluster.cluster.name` I don't feel overly opinionated. It's not as if we're performing the same interpolation (via convention) over and over in several places - to me that's where devs don't fully leverage what terraform can do on your behalf. In this case, we're purely in the realm of philosophy. `aws_eks_cluster.cluster.name` makes good sense to me and I see the case for it.

Going to merge this as is. I'm happy with it, it breaks no one, and adds some clarity. 💯

Issue #50, opened 2018-06-30
Allow adding new users, roles, and accounts to the configmap/aws-auth


sdavids13:
# I have issues

Amazon's EKS access control is managed via the aws-auth configmap which allows multiple IAM users and roles (cross-account capable) to be granted group membership. The current implementation only allows worker node access, this should be configurable to allow more access control rules to be specified per the documentation: https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

## I'm submitting a

* [ ] bug report
* [x] feature request
* [ ] support request

## What is the current behavior

The current implementation only allows worker node access.

## What's the expected behavior

Ability to specify role/user/account mappings for group membership.

## Environment

* Affected module version: 1.1.0
* OS: Linux
* Terraform version: 0.11.7

max-rocket-internet:
Yes 100%! We need this ASAP. I was going to make a PR but haven't had time yet.

The problem is that also by default, the cluster writes `config_map_aws_auth` and then applies it immediately with `null_resource.configure_kubectl" and this overwrites any users or roles that have been added post cluster creation and would thus lock out those users.

Here's how I think it should look:

```
module "cluster_1" {
  source             = "git@github.com:terraform-aws-modules/terraform-aws-eks.git?ref=v1.2.0"
  cluster_name       = "cluster_1"
  subnets            = ["${aws_subnet.public.*.id}"]
  vpc_id             = "${aws_vpc.eks.id}"
  config_output_path = "kubeconfig"

  worker_groups = [
    {
      instance_type = "t2.medium"
      key_name      = "default"
    },
  ]

  map_roles             = [
    {
      role_arn = "${aws_iam_role.developers.arn}"
      username = "developers"
      groups [
        "cluster-admin"
      ]
    }
  ]

  map_users             = [
    {
      user_arn = "${aws_iam_user.dave.arn}"
      username = "${aws_iam_user.dave.name}"
      groups [
        "system:masters"
      ]
    },
    {
      user_arn = "${aws_iam_user.steve.arn}"
      username = "${aws_iam_user.steve.name}"
      groups [
        "system:basic-user"
      ]
    }
  ]

  map_accounts = [
    "666666666666",
    "999999999999"
  ]
}
```

brandoconnor:
Hey All - 

[...]

PR #56, opened 2018-07-09
Fully manage aws-auth configmap file

max-rocket-internet:
Related issue: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/50

The problem: Any users, roles or accounts that need access to the EKS cluster must have the relevant entries in the `aws-auth` configmap inside k8s. Currently this module only puts the ARN of the worker group in there, thus forcing people to manage this file manually, which is a pain.

Limitations:

1. It's a bit of a mess due to needing to loop over values in a template. Not sure if the new template syntax in [0.12](https://www.hashicorp.com/blog/terraform-0-1-2-preview) will help. Let me know if there's a better way to do this!

2. We can't have multiple groups per user or role due to this error: `lookup() may only be used with flat maps, this map contains elements of type list in`. No big deal I think at this stage.

3. We can't have interpolation in the items being passed to the module as it breaks the count option. Details [here](https://github.com/hashicorp/terraform/issues/10857). Annoying but also not a big deal.

Example configuration:

```hcl
module "cluster_1" {
  source                    = "git@github.com:terraform-aws-modules/terraform-aws-eks.git?ref=v1.4.0"
  cluster_name              = "cluster_1"
  subnets                   = ["${aws_subnet.public.*.id}"]
  vpc_id                    = "${aws_vpc.eks.id}"

  worker_groups = [
    {
      instance_type = "t2.medium"
      key_name      = "default"
    }
  ]

  map_roles = [
     {
       role_arn = "${aws_iam_role.k8s_users.arn}"
       username = "k8s_users"
       group = "system:masters"
     }
   ]

   map_users = [
     {
       user_arn = "arn:aws:iam::66666666666:user/user.name"
       username = "user.name"
       group  = "system:masters"
     }
   ]

   map_accounts = [
     "666666666666",
     "999999999999"
   ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Generate them like so:
 
 ```bash
 go get github.com/segmentio/terraform-docs
-terraform-docs md ./ | cat -s > README.md
+terraform-docs md ./ | cat -s | ghead -n -1 > README.md
 ```
 
 ## Contributing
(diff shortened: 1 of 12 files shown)


max-rocket-internet:
1. Done
2. Done but I'm having to guess the date it would be merged and also the tag name. Is that correct? Wouldn't this be done after creating a tag?
3. Done. Does it look OK?

README.md
```diff
@@ -63,7 +63,7 @@ Generate them like so:
 
 ```bash
 go get github.com/segmentio/terraform-docs
-terraform-docs md ./ | cat -s | ghead -n -1 > README.md
```
brandoconnor: This line didn't include the `ghead` bit just for giggles. 😉 Let's keep it UNIXy 'round these parts and preserve those end of file newlines, eh?

brandoconnor: If that means moving from `ghead` to `head`, so be it. Maybe it's my own bias but I assume most people are developing on Macs which don't come with `ghead` but it's the closest thing to our Linux tool cousin.

brandoconnor: Actually, scrap the above comment. Let's do something that works for both modern Linux and Mac flavors. Mind changing it to the following:

```bash
terraform-docs md ./ | cat -s | tail -r | tail -n +2 | tail -r > README.md
```

brandoconnor:
Can you figure out what's going on with the failed `terraform fmt` and fix up my comment around the `terraform-docs` command? We're sooooo close!

brandoconnor:
This is great btw. Thanks again!

README.md
brandoconnor: It's a utility that does the same thing as unix `head`. The problem is you have to install it via brew. `tail` should be shipped standard on any mac unless I'm off my rocker.

max-rocket-internet:
I'm still resolving conflicts.

max-rocket-internet:
OK I was a little confused after all my other PRs were merged and then I had `exec: "aws-iam-authenticator": executable file not found in $PATH` but now resolved.

1. terraform fmt done
2. docs change tested and done

Anything else?

PR #59, opened 2018-07-10
Updating heptio-authenticator-aws to aws-iam-authenticator

CharlieC3:
Updating heptio-authenticator-aws to aws-iam-authenticator since the heptio project was moved and renamed.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ kubeconfig
 config-map-aws-auth.yaml
 eks-admin-cluster-role-binding.yaml
 eks-admin-service-account.yaml
+.idea/
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Read the [AWS docs on EKS to get connected to the k8s dashboard](https://docs.aw
 * You want to create an EKS cluster and an autoscaling group of workers for the cluster.
 * You want these resources to exist within security groups that allow communication and coordination. These can be user provided or created within the module.
 * You've created a Virtual Private Cloud (VPC) and subnets where you intend to put the EKS resources.
-* If using the default variable value (`true`) for `configure_kubectl_session`, it's required that both [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl) (>=1.10) and [`heptio-authenticator-aws`](https://github.com/heptio/authenticator#4-set-up-kubectl-to-use-heptio-authenticator-for-aws-tokens) are installed and on your shell's PATH.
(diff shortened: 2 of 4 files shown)


sdavids13:
Do we know when this binary name change will be incorporated into the "EKS Optimized" AMI? I would think we should try to synchronize the release of this PR with the release of an updated AMI with the name change.

brandoconnor:
@sdavids13 - I don't know that we have a timeline on when this binary will be included in an AWS AMI or even a promise that they will. Is the previous heptio authenticator binary included currently? To me I don't know that it makes sense to wait on an unknown like that. Either way, if a user doesn't feel inclined to use the new command, they can always set the variable until their heart's desire.

@CharlieC3 thanks for the keeping your eyes out for this and picking up the change. 👍

sdavids13:
@brandoconnor Yes, they currently do include it in their AMI. That particular AMI is specified as the default here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/data.tf#L22

You can see that Amazon is still using the old binary in their packer scripts here: https://github.com/awslabs/amazon-eks-ami/blob/19279be9fa95dce0057474609fc28f094fa4cf6d/install-worker.sh#L89

The current EKS documentation says to use the heptio distribution as well: https://docs.aws.amazon.com/eks/latest/userguide/configure-kubectl.html

So, it is important to note that with this change by default none of the worker nodes will be able to communicate with the control plane/master nodes. Everyone who wants to use the other default values in this project will need to set this particular variable to the old value (until Amazon starts using the new binary).

CharlieC3:
@sdavids13 
> So, it is important to note that with this change by default none of the worker nodes will be able to communicate with the control plane/master nodes.

I don't believe that is true. Correct me if I'm wrong, but I believe the `aws-iam-authenticator/heptio` command is only used in this module to generate a kubeconfig file so the user can easily connect to the cluster via `kubectl`. It isn't used to dictate how the worker nodes authenticate with the master node(s).

I created a new EKS cluster while overriding the `kubeconfig_aws_authenticator_command` default value with the new `aws-iam-authenticator` command prior to submitting this PR, and the worker nodes are able to communicate with the master(s) just fine.

>The current EKS documentation says to use the heptio distribution as well

The docs include the following snippet a bit further down:

_The open source version of this binary has been renamed to aws-iam-authenticator. If you download this binary using go get, make sure to follow the steps below that use the updated name._

CharlieC3:
No worries! Glad you spoke your concern anyways.

brandoconnor:
Whew! Always feels good when I'm not the one taking crazy pills! 😉 

Thanks both of you for helping out here. Sanity check always appreciated!

PR #61, opened 2018-07-12
Optionally define worker subnets

bshelton229:
# PR o'clock

## Description

Adds a `subnets` option to `worker_groups`. The primary use-case is to address #15 . This will allow the eks cluster to be passed both private and public subnets while the workers only run in the private subnets.

Thanks!

### Checklist


brandoconnor:
Thanks again for the work here and discussion on #15 !

Issue #57, opened 2018-07-09
root_block_device  missed


ozbillwang:
# I have issues

running EKS cluster for a week, get disk space issue, need feature to control and extend with larger  size.

## I'm submitting a...

* [X ] feature request

## What is the current behavior?

No control on root block device,  disk space is used out quickly

## What's the expected behavior?

```
The root_block_device mapping supports the following:

volume_type - (Optional) The type of volume. Can be "standard", "gp2", or "io1". (Default: "standard").
volume_size - (Optional) The size of the volume in gigabytes.
iops - (Optional) The amount of provisioned IOPS. This must be set with a volume_type of "io1".
delete_on_termination - (Optional) Whether the volume should be destroyed on instance termination (Default: true).
```
Currently, only enable `delete_on_termination`

## Are you able to fix this problem and submit a PR? Link here if you have already.

Yes, I will

## Environment details

* Affected module version: v1.1.0
* OS: Ubuntu
* Terraform version: 0.11.7

brandoconnor:
I like the idea. Waiting on your mark to PR this.

max-rocket-internet:
This has been addressed here: https://github.com/terraform-aws-modules/terraform-aws-eks/commit/b71e3e09282ae1e8661e965543463620639930ef

PR #66, opened 2018-07-13
Manage eks workers' root_block_device

ozbillwang:
# PR o'clock

## Description

for #57 to manage eks workers' root volume size, type and iops (if volume type is iops)

### Checklist


brandoconnor:
Let me know when this is ready to ship. I've added WIP to the title but that can be removed when this work is ready for review.

ozbillwang:
@brandoconnor

It can be reviewed already

variables.tf
```diff
@@ -85,6 +85,9 @@ variable "workers_group_defaults" {
     asg_max_size         = "3"           # Maximum worker capacity in the autoscaling group.
     asg_min_size         = "1"           # Minimum worker capacity in the autoscaling group.
     instance_type        = "m4.large"    # Size of the workers instances.
+    root_volume_size     = "20"          # root volume size of workers instances.
```
brandoconnor: Should have asked before I merged: Are these values the defaults that worker ASGs received before this change?

ozbillwang: Not sure. When I start EKS worker as `m5.xlarge`, its root disk size is 20GB, has no chance to test with other instance. 

But if AWS designs the EKS workers and set 20GB with gp2, it should be very reasonable and I also think it works for most case. 

But if you need double check, I can try to start other instance type or if anyone can directly give the data if they already run with other instance types.

ozbillwang:
By the way, do you need me add details in README (in main.tf) to explain how to use the map variable `worker_groups`, it becomes an important feature for this module now.

variables.tf
ozbillwang: @brandoconnor 

I can confirm the default root size now: 

https://github.com/awslabs/amazon-eks-ami/blob/master/eks-worker-al2.json#L22

above repo is managed by AWS EKS team and use packer to make worker AMI images.

brandoconnor:
Nice!

Issue #40, opened 2018-06-27
Cluster and worker security group specification doesn't work


jimbeck:
# I have issues
I am creating an eks cluster with providing a `cluster_security_group_id` and `worker_security_group_id`.

## I'm submitting a

* [X] bug report
* [ ] feature request
* [ ] support request

## What is the current behavior
When specifying a security group aka `cluster_security_group_id = "sg-123"` or `worker_security_group_id = "sg-123"`. I get 
```
Error: Error running plan: 1 error(s) occurred:
module.eks.local.cluster_security_group_id: local.cluster_security_group_id: Resource 'aws_security_group.cluster' not found for variable 'aws_security_group.cluster.id'`<br>
```

## If this is a bug, how to reproduce? Please include a code sample
Create an eks cluster with a cluster_security_group_id or worker_security_group_id specified.<br>
Terraform does not support short circut evaluation in it's ternary operator. The fix to the issue is specified [here](http://67bricks.com/blog/?p=85)

## What's the expected behavior
We should be able to create an EKS cluster while specifying cluster sg or worker sg as the documentation currently specifies.

## Environment

* Affected module version: 1.1.0
* OS: 
* Terraform version: 0.11.7

brandoconnor:
Ahh I wish I would have tested for this now. Thanks I'll merge now.

PR #69, opened 2018-07-15
Don't call out to fetch the current ip address if it isn't needed.

sdavids13:
# PR o'clock

## Description

If we specify a workstation_cidr block don't go out and fetch your current external ip address for the "default" value. This breaks systems if they don't have a route out to "https://ipv4.icanhazip.com"

### Checklist

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -1,6 +1,7 @@
 data "aws_region" "current" {}
 
 data "http" "workstation_external_ip" {
+  count = "${var.workstation_cidr == "" ? 1 : 0}"
   url = "https://ipv4.icanhazip.com"
 }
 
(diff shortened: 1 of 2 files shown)


brandoconnor:
It turns out as @bshelton229 mentioned in #40 this stanza does nothing. There's no way to lock down access to the EKS control plane through terraform or otherwise. Docs dance around this a bit but I think we should remove the security group stanza altogether as it doesn't appear to be useful.

If you do see some use for the security group remaining, I'm all ears to understand its value. As it is, this doesn't work how you would think it should, and hashicorp's initial stab at a guide didn't understand that either.

Want to change this PR to remove that SG and local/var references entirely?

brandoconnor:
@sdavids13 - see my comment above. I'm so very on board with deleting code where possible.

Issue #87, opened 2018-08-09
Bug - EKS can not create load balancers after module provisioned in new AWS account


mmcaya:
# I have issues
Provisioning EKS cluster in **new AWS account** will result in an error when attempting to provision a load balancer if no load balancers of any kind have been provisioned before.

## I'm submitting a...

* [X] bug report

## What is the current behavior?

No previous load balancers ( i.e. service-link role **AWSServiceRoleForElasticLoadBalancing** doesn't exist)

`AccessDenied: User: <MODULE-PROVISIONED-ROLE> is not authorized to perform: iam:CreateServiceLinkedRole on resource: arn:aws:iam::<ACCOUNT-ID>:role/aws-service-role/elasticloadbalancing.amazonaws.com/AWSServiceRoleForElasticLoadBalancing`

because EKS is attempting to create the ELB service-link role for you, and the roles created by the module lack `iam:CreateServiceLinkedRole` 

## If this is a bug, how to reproduce? Please include a code sample if relevant.

- Provision EKS cluster using module into new account (or ensure service-link role **AWSServiceRoleForElasticLoadBalancing** doesn't exist)
- Attempt to provision a Service of type LoadBalancer via kubernetes

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 1 
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.15.2
        ports:
        - containerPort: 80
---
kind: Service
apiVersion: v1
metadata:
  name: nginxservice
spec:
  type : LoadBalancer
  selector:
    app: nginx
  ports:
  - protocol: TCP
    port: 80
    targetPort: 80
```

## What's the expected behavior?
EKS should provision load balancer.

Module should optionally provision  (via flag) a `resource "aws_iam_service_linked_role"`, or include updated IAM policies (`iam:CreateServiceLinkedRole`) to allow the EKS cluster to provision the required service-link role.  Alternatively, if this is deemed not the responsibility of the module, the "Assumptions" section in README.md should note the issue.

## Are you able to fix this problem and submit a PR? Link here if you have already.
Possibly, depending on the choice of solution (implementation change, documentation update)

## Environment details

* Affected module version: All

## Any other relevant info
AWS Service Link FAQ:
[...]

PR #91, opened 2018-08-13
Adding aws_iam_service_linked_role to fix ELB creation error

max-rocket-internet:
A fix for [this error](https://stackoverflow.com/questions/51597410/aws-eks-is-not-authorized-to-perform-iamcreateservicelinkedrole/51757830) and [this issue](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/87).

Do you want me to update the change log for this?


max-rocket-internet:
@strygin exactly what part do you think should be split into a separate PR?

workers.tf
```diff
@@ -122,3 +122,8 @@ resource "null_resource" "tags_as_list_of_maps" {
     "propagate_at_launch", "true"
   )}"
 }
+
+resource "aws_iam_service_linked_role" "elasticloadbalancing" {
```
tonyxiao: Should this be inside `workers.tf`? This role is used by the cluster.

max-rocket-internet: Good point. I'll move it.

PR #92, opened 2018-08-13
Setting worker default volume size to 100GB

max-rocket-internet:
I think as a default, 20GB is way too low. For example, GKE uses 100GB.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -87,7 +87,7 @@ variable "workers_group_defaults" {
     asg_min_size         = "1"           # Minimum worker capacity in the autoscaling group.
     instance_type        = "m4.large"    # Size of the workers instances.
     spot_price           = ""            # Cost of spot instance.
-    root_volume_size     = "20"          # root volume size of workers instances.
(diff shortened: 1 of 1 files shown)


max-rocket-internet:
Oh someone just made a PR to change the official CFN templates:
https://github.com/awslabs/amazon-eks-ami/pull/20

They also state that 20 is too low but then set the default to 20?? 🤷‍♂️

brandoconnor:
LGTM - This will be released with the next major rev (within the week) since without action (explicitly defining disk size where the 20 GB default was previously used), it would blow away and recreate the worker groups.

max-rocket-internet:
> it would blow away and recreate the worker groups
I think it will just replace the LC on the ASG? Nothing more?

brandoconnor:
Good call and apologies on the alarmism. I'm more accustomed to the rolling release setup where entire ASGs are replaced on a change to the underlying LC but those facilities aren't in place here so all remains quiet.

PR #98, opened 2018-08-22
add t3 instance type settings

rochacon:
## Description

AWS announced [1] new t3 instance types, this PR adds the necessary configuration to use them. 

[1] https://aws.amazon.com/blogs/aws/new-t3-instances-burstable-cost-effective-performance/

### Checklist

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -64,6 +64,11 @@ locals {
     t2.large    = 35
     t2.xlarge   = 44
     t2.2xlarge  = 44
+    t3.small    = 8
+    t3.medium   = 17
+    t3.large    = 35
+    t3.xlarge   = 44
+    t3.2xlarge  = 44
     x1.16xlarge = 234
     x1.32xlarge = 234
   }
@@ -178,6 +183,11 @@ locals {
     "t2.nano"      = false
     "t2.small"     = false
     "t2.xlarge"    = false
+    "t3.small"     = false
+    "t3.medium"    = false
+    "t3.large"     = false
+    "t3.xlarge"    = false
+    "t3.2xlarge"   = false
     "x1.16xlarge"  = true
     "x1.32xlarge"  = true
     "x1e.16xlarge" = true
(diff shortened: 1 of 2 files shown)


local.tf
```diff
@@ -178,6 +183,13 @@ locals {
     "t2.nano"      = false
     "t2.small"     = false
     "t2.xlarge"    = false
+    "t3.2xlarge"   = false
+    "t3.large"     = false
+    "t3.medium"    = false
+    "t3.micro"     = false
+    "t3.nano"      = false
```
max-rocket-internet: Maybe don't include nano and micro here if they aren't in the `max_pod_per_node` map also?

rochacon: Agreed. I basically duplicated the t2 block and replaced t2 with t3.

Should I drop the t2.nano and t2.micro from this list also?

max-rocket-internet: Ah I see what you did. Yes, I think make your change consistent. Don't worry about the other values.

rochacon: Got it. It's done.

strygin: @max-rocket-internet 
If you request changes in your "reviews", please try to inspect the code more deeply. This is not a linux kernel repo and your analysis is too superficial. What you do here is actually slowing the progress down.

rochacon:
Thank you @max-rocket-internet!! Any idea on the release date to 1.5.0? I'm also interested in the monitoring flag and spot instances.

local.tf
brandoconnor: Good catch on these @rochacon and I agree with what you've suggested @max-rocket-internet . I've been experimenting with a way we can start relying on [this utility module](https://github.com/terraform-aws-modules/terraform-aws-ebs-optimized) for the `ebs_optimized` portion and will pitch sourcing it from there.

@strygin - Max's comments here are the furthest thing from pedantic or taste-based. Again, see my feedback elsewhere and prioritize being respectful, thoughtful, and on discussion in future comments.

brandoconnor:
Apologies I haven't released sooner @rochacon . I might have time this weekend but I've been super busy as of late. @max-rocket-internet if you've got the bandwidth, feel free to get a release PR up so we can discuss and ship it.

max-rocket-internet:
> Thank you @max-rocket-internet!! Any idea on the release date to 1.5.0?

I'll do it this week. Tomorrow or Wednesday 🙂

PR #121, opened 2018-09-07
Allow to use existing IAM resources instead of creating new ones

mandrean:
# PR o'clock

## Description

This PR adds the ability to conditionally use existing IAM roles and instance profiles instead of creating new ones.

The feature is crucial in organizations where you have very strict `iam:*` access and can't freely create/attach/pass roles.

### Checklist

- [X] Docs have been updated using `terraform-docs` per `README.md` instructions
- [X] Any breaking changes are highlighted above

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ Full contributing [guidelines are covered here](https://github.com/terraform-aws
 Testing and using this repo requires a minimum set of IAM permissions. Test permissions
 are listed in the [eks_test_fixture README](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/eks_test_fixture/README.md).
 
+For environments with strict `iam:*` permissions, existing IAM roles can be provided. See `cluster_service_role_name`, `worker_instance_role_name` and `worker_instance_profile_name`.
+
 ## Change log
 
 The [changelog](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/CHANGELOG.md) captures all important release notes.
@@ -100,6 +102,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 |------|-------------|:----:|:-----:|:-----:|
 | cluster_name | Name of the EKS cluster. Also used as a prefix in names of related resources. | string | - | yes |
 | cluster_security_group_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingres/egress to work with the workers and provide API access to your current IP/32. | string | `` | no |
+| cluster_service_role_name | If provided, the EKS cluster will use this service role. If not given, a service IAM role will be created with the necessary policies attached. | string | `` | no |
 | cluster_version | Kubernetes version to use for the EKS cluster. | string | `1.10` | no |
 | config_output_path | Determines where config files are placed if using configure_kubectl_session and you want config files to land outside the current working directory. Should end in a forward slash / . | string | `./` | no |
 | create_elb_service_linked_role | Whether to create the service linked role for the elasticloadbalancing service. Without this EKS cannot create ELBs. | string | `false` | no |
@@ -117,6 +120,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | worker_additional_security_group_ids | A list of additional security group ids to attach to worker instances | list | `<list>` | no |
 | worker_group_count | The number of maps contained within the worker_groups list. | string | `1` | no |
 | worker_groups | A list of maps defining worker group configurations. See workers_group_defaults for valid keys. | list | `<list>` | no |
+| worker_instance_profile_name | If provided, the EKS worker nodes will spawn using this EC2 instance IAM profile. If not given, an instance profile will be created with the necessary policies attached. | string | `` | no |
+| worker_instance_role_name | If provided, the EKS worker nodes will spawn using this EC2 instance IAM role. If not given, an instance role will be created with the necessary policies attached. | string | `` | no |
 | worker_security_group_id | If provided, all workers will be attached to this security group. If not given, a security group will be created with necessary ingres/egress to work with the EKS cluster. | string | `` | no |
 | worker_sg_ingress_from_port | Minimum port number from which pods will accept communication. Must be changed to a lower value if some pods in your cluster will expose a port lower than 1025 (e.g. 22, 80, or 443). | string | `1025` | no |
 | workers_group_defaults | Override default values for target groups. See workers_group_defaults_defaults in locals.tf for valid keys. | map | `<map>` | no |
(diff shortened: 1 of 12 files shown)


mandrean:
I'm not sure how you want to do with the test fixture? Maybe add a second one which assumes existing service roles, instance roles and instance profiles?

README.md
```diff
@@ -126,15 +131,21 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
...
-| worker_iam_role_name | IAM role name attached to EKS workers |
+| config_map_aws_auth | A kubernetes configuration to authenticate to the EKS cluster. |
+| kubeconfig | kubectl config file contents for the EKS cluster. |
+| worker_instance_profile_arn | EC2 instance IAM role ARN attached to the EKS workers. |
+| worker_instance_profile_name | EC2 instance IAM role ARN attached to the EKS workers. |
+| worker_instance_role_arn | Instance IAM role ARN attached to the EKS workers. |
+| worker_instance_role_name | Instance IAM role name attached to the EKS workers. |
 | worker_security_group_id | Security group ID attached to the EKS workers. |
-| workers_asg_arns | IDs of the autoscaling groups containing workers. |
+| workers_asg_arns | ARNs of the autoscaling groups containing workers. |
```
max-rocket-internet: 👏

README.md
```diff
@@ -117,6 +120,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | worker_additional_security_group_ids | A list of additional security group ids to attach to worker instances | list | `<list>` | no |
 | worker_group_count | The number of maps contained within the worker_groups list. | string | `1` | no |
 | worker_groups | A list of maps defining worker group configurations. See workers_group_defaults for valid keys. | list | `<list>` | no |
+| worker_instance_profile_name | If provided, the EKS worker nodes will spawn using this EC2 instance IAM profile. If not given, an instance profile will be created with the necessary policies attached. | string | `` | no |
+| worker_instance_role_name | If provided, the EKS worker nodes will spawn using this EC2 instance IAM role. If not given, an instance role will be created with the necessary policies attached. | string | `` | no |
```
max-rocket-internet: Why pass the role name when really only the instance profile is used by the launch configuration?

examples/eks_test_fixture/outputs.tf
```diff
@@ -1,8 +1,28 @@
...
+}
+
+output "cluster_id" {
+  description = "The name/id of the EKS cluster."
+  value       = "${module.eks.cluster_id}"
+}
+
+output "cluster_arn" {
+  description = "The Amazon Resource Name (ARN) of the cluster."
+  value       = "${module.eks.cluster_arn}"
```
max-rocket-internet: 👏

max-rocket-internet:
Hey @mandrean,
Thanks for the PR!

There's quite a lot of complexity here for something that probably 95% of users won't use. I'm not against merging it, just want to question a few things....

In regards to the workers, put simply, all you want to do is change the `iam_instance_profile` attribute of the worker LC for 1 or all worker groups. Am I wrong?

Wouldn't this work?

```hcl
resource "aws_launch_configuration" "workers" {
  iam_instance_profile = "${lookup(var.worker_groups[count.index], "iam_instance_profile", aws_iam_instance_profile.workers.id}"
```

I know this will still create `aws_iam_role` and `aws_iam_instance_profile` but they won't be used.

max-rocket-internet:
@sc250024 

> I would disagree with @max-rocket-internet that it's something "95% of users won't use"

Fair enough, maybe you are right, it might be more than I think 🙂

> It seems to me that #137 is excessive since with Kube2IAM / Kiam

I don't think we want to merge changes with the assumption of people using a specific tool unless it's REALLY common. I don't think Kube2IAM / Kiam fits that criteria right now. The cluster-autoscaler would fit that criteria I think.

I still don't quite understand your point though. If you want to use Kube2IAM and assign specific roles to pods then neither of these PRs make any difference to that ability. Unless I'm mistaken?

Anyway, none of this really matters. The author of this PR hasn't replied to my question from 2 weeks ago and @laverya has been very responsive. I also prefer https://github.com/terraform-aws-modules/terraform-aws-eks/pull/137 since it's quite a lot simpler.

If either @sc250024 or @mandrean want to make a new PR then all good 👍

README.md
mandrean: @max-rocket-internet The worker instance role name is used in [`aws_auth.tf` to pass into the kubeconfig](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/121/files#diff-2812d2bbb29e6bacd1bed6f9acf4a797R27) and [worker-role.tpl](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/templates/worker-role.tpl#L1)

mandrean:
> There's quite a lot of complexity here for something that probably 95% of users won't use. I'm not against merging it, just want to question a few things....

Maybe these stats are, maybe not.

I'm sure a majority of enterprise users like banks, insurance companies, large corporations etc. would need these features though. And with EKS being out in Europe, more and more enterprises (just like my current client) will look at moving their existing `kube-aws` or `kops` clusters to EKS instead for manageability and cost-saving reasons.

It would be a shame if they couldn't use this great module and would have to resort to CloudFormation instead, just because this module assumes you have full `iam:*` permissions which is rare in large companies.

sc250024:
@max-rocket-internet To respond to your points:

> I still don't quite understand your point though. If you want to use Kube2IAM and assign specific roles to pods then neither of these PRs make any difference to that ability. Unless I'm mistaken?

You assign specific roles to pods, but those pod roles must be assumed by the worker node role that's attached to the worker nodes. As I understand it, when using this Terraform module in its current form, a new worker node role is generated for every cluster. It would be nice to have one worker node role for all potential EKS clusters; that way, someone doesn't have to re-grant any new worker node role privileges to assume all the pod roles required. At my startup, we have around ~15 pod roles for 15 apps, each with their own set of permissions; all 15 of these need to trust the worker node role to allow that role to assume the pod role.

I think Kube2IAM / Kiam are used quite frequently actually. A quick Google search reveals this:

- [Securing AWS EKS with Calico and Kube2iam](http://www.clearscale.com/blog/securing-aws-eks-with-calico-and-kube2iam/)
- [Securing AWS IAM access in Kubernetes](https://medium.com/merapar/securing-iam-access-in-kubernetes-cfbcc6954de)
- [Moving from Kube2Iam to Kiam](https://medium.com/building-ibotta/moving-from-kube2iam-to-kiam-a000639b839e)
- [Installing kube2iam in AWS Kubernetes Kops Cluster](https://blog.csanchez.org/2018/06/12/installing-kube2iam-in-aws-kubernetes-kops-cluster/)
- [CloudPosse Kube2IAM Docs](https://docs.cloudposse.com/kubernetes-backing-services/iam/kube2iam/)

What I would say to people running **any** Kubernetes cluster in AWS is, if you're not using Kube2IAM / Kiam, you absolutely should; otherwise you're allowing any pod to have credentials to resources that it shouldn't. Google Cloud works in the same way, although they have a direct link between service accounts created in the GCloud IAM area, and the service accounts that exist in Kubernetes. AWS doesn't have this as such, but the closest thing is Kube2IAM or Kiam.

@mandrean I think first to resolve the merge conflicts.

max-rocket-internet:
@sc250024 

> As I understand it, when using this Terraform module in its current form, a new worker node role is generated for every cluster.

By default, yes.

> It would be nice to have one worker node role for all potential EKS clusters

But now that 137 is merged, you can just pass the same role ARN for each of your work groups if you want. So that's all sorted then?

>  if you're not using Kube2IAM / Kiam, you absolutely should; otherwise you're allowing any pod to have credentials and access to resources that it shouldn't.

Only if you are giving access to your applications via the worker node role, which you should not be doing. So currently by default any pod would have [this access and no more](https://github.com/awslabs/amazon-eks-ami/blob/master/amazon-eks-nodegroup.yaml#L193-L195).

> At the end of the day, whichever PR solves the base problem (re-using worker IAM roles) is fine with me.

This is already done as I understand it, you just need to set `iam_role_id` in your worker group config to whatever you want. Or you can even use the `workers_group_defaults` variable to add a default that applies for all your worker groups.

@mandrean 

I understand all your points. But I think PR137 has solved this, with less complexity in this module. Am I wrong?

> It would be a shame if they couldn't use this great module and would have to resort to CloudFormation instead (😱)

BTW, they current CFN templates from AWS don't support this feature either 😉

Always happy to accept more PRs guys ♥️

sc250024:
Thanks @max-rocket-internet for the merge on https://github.com/terraform-aws-modules/terraform-aws-eks/pull/137, great points too. Cheers!

max-rocket-internet:
I'm gonna close this PR as we merged the other one in the end.

Feel free to create a new PR if you feel that there is still functionality missing though 🙂

PR #129, opened 2018-09-11
add var to allow some commands to go before aws-iam-authenticator

nicgrayson:
# PR o'clock

## Description

I want my full auth command to be:
`aws-okta exec <account> exec -- aws-iam-authenticator token -i <cluster_name>`
So I want some args to go first while not breaking the current template.

### Checklist


max-rocket-internet:
Sounds a little niche but not sure. I'll defer to Brandon.

nicgrayson:
@brandoconnor It's about the ordering of the arguments. I'm open to a more elegant solution if there is one. When using this branch I can build this command:
```users:
- name: eks-rnd
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: aws-okta
      args:
        - exec
        - rnd
        - --
        - aws-iam-authenticator
        - token
        - -i
        - eks-rnd```
If i provided `kubeconfig_aws_authenticator_additional_args` it would put them at the end. This is to make https://github.com/segmentio/aws-okta or https://github.com/99designs/aws-vault to work

**Comment by max-rocket-internet** [comment]
@nicgrayson can't you just set `kubeconfig_aws_authenticator_command` to `aws-okta exec rnd -- aws-iam-authenticator` to achieve this?

**Comment by nicgrayson** [comment]
@max-rocket-internet I tried that first but it seems like that isn't a supported format for the config

```
users:
- name: eks_nic-test
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: aws-okta -- exec turbo-rnd -- aws-iam-authenticator
      args:
        - "token"
        - "-i"
        - "nic-test"
```
```
kubectl --kubeconfig kubeconfig_nic-test get nodes
Unable to connect to the server: getting credentials: exec: exec: "aws-okta -- exec nic-test -- aws-iam-authenticator": executable file not found in $PATH
```

**Comment by ngrayson-amfam** [comment]
We could also fix this by changing default args to "token -i cluster" and allowing a user to overwrite all of the args

**Comment by max-rocket-internet** [comment]
Sorry for the delay.

> We could also fix this by changing default args to "token -i cluster" and allowing a user to overwrite all of the args

I think that would be more elegant. Maybe something like this?

```
      command: ${aws_authenticator_command}
      args:
${aws_authenticator_command_args}
${aws_authenticator_additional_args}
${aws_authenticator_env_variables}
```

And then just make `aws_authenticator_command_args` formatted like `aws_authenticator_additional_args` with default values that produce the same output as we have now.

Does that make sense? Then let's get this merged.

**Comment by nicgrayson** [comment]
@max-rocket-internet How does it look now?

**Comment by max-rocket-internet** [comment]
It's not quite right, currently it produces this:

```
      args:

        - token
        - -i
        - cluster-name

      env:
```

Perhaps remove the leading `\n`?

nicgrayson:
@max-rocket-internet I amended my commit. Actually testing both setups this time.

max-rocket-internet:
Now the values are not quoted..

max-rocket-internet:
> Sorry I didn't think that was required

It's just YAML so probably not required but it would be nice to have this PR not change the default behaviour 🙂

> Also do you think this way of doing it is fine or

I think we merge this now as I think quite a few people use okta. This PR does make it a little hard to see what's going on but feel free to fix that in a new PR with a local if it looks tidier.

PR #131, opened 2018-09-12
Cosmetic fixes

max-rocket-internet:
# PR o'clock

## Description

Cosmetic syntax changes to improve readability. Credit goes to @strygin for the inspiration.

2 changes:

- Remove lookup function when no default is required to just refer to the map and key like this `map["key"]`. This reduces the `()` which improves readability.
- Where possible, use an actual map instead of using the map function. e.g. `{key = "value"}`

### Checklist


brandoconnor:
Looks like `fmt` is failing against `workers.tf`. Run that quick and this should be good to ship.

max-rocket-internet:
Ahh yeah. I tried a rebase and then manual edit in Gitkraken and it left some blank lines.

PR #137, opened 2018-09-17
allow specifying an IAM role for each worker group

### LINKED ISSUE #136: Allow giving different ASGs different IAM roles
# I have issues

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request

One `aws_iam_instance_profile` is created and shared between all ASGs. This greatly complicates giving different ASGs different permissions.

The desired behavior would be to have a flag that could be passed that would cause an IAM role to be created for each cluster, and have lists of role names+ARNs returned in addition to the current `worker_iam_role_name` and `worker_iam_role_arn` outputs.

Yes, #137 

## Environment details

* Affected module version: 1.6.0
* OS:
* Terraform version:

laverya:
# PR o'clock

## Description

-resolves #136 
-added `iam_role_id` to the worker group config

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,6 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | map_accounts | Additional AWS account numbers to add to the aws-auth configmap. See examples/eks_test_fixture/variables.tf for example format. | list | `<list>` | no |
 | map_roles | Additional IAM roles to add to the aws-auth configmap. See examples/eks_test_fixture/variables.tf for example format. | list | `<list>` | no |
 | map_users | Additional IAM users to add to the aws-auth configmap. See examples/eks_test_fixture/variables.tf for example format. | list | `<list>` | no |
-| multiple_worker_group_iam_roles | If provided, an IAM role will be created for each worker group instead of all worker groups sharing the same role. | string | `` | no |
 | subnets | A list of subnets to place the EKS cluster and workers within. | list | - | yes |
 | tags | A map of tags to add to all resources. | map | `<map>` | no |
 | vpc_id | VPC where the cluster and workers will be deployed. | string | - | yes |
@@ -134,10 +133,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | cluster_version | The Kubernetes server version for the EKS cluster. |
 | config_map_aws_auth | A kubernetes configuration to authenticate to this EKS cluster. |
 | kubeconfig | kubectl config file contents for this EKS cluster. |
-| worker_iam_role_arn | IAM role ID attached to the first EKS worker group |
-| worker_iam_role_arns | IAM role IDs attached to EKS workers |
-| worker_iam_role_name | IAM role name attached to the first EKS worker group |
-| worker_iam_role_names | IAM role names attached to EKS workers |
+| worker_iam_role_arn | default IAM role name for EKS worker groups |
+| worker_iam_role_name | default IAM role name for EKS worker groups |
 | worker_security_group_id | Security group ID attached to the EKS workers. |
 | workers_asg_arns | IDs of the autoscaling groups containing workers. |
 | workers_asg_names | Names of the autoscaling groups containing workers. |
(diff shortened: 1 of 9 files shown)


variables.tf
```diff
@@ -123,3 +123,8 @@ variable "kubeconfig_name" {
   description = "Override the default name used for items kubeconfig."
   default     = ""
 }
+
+variable "multiple_worker_group_iam_roles" {
+  description = "If provided, an IAM role will be created for each worker group instead of all worker groups sharing the same role."
+  default     = ""
```
max-rocket-internet: It looks like this is a list that would need to be the same length as the amount of work groups? Is that correct?

laverya: No, this was a flag/boolean - it's used here: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/137/files#diff-0973ac2dc1229ea1662319bdc6954f08R161 and here: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/137/files#diff-b66ca10954cbb6be58939e90083e932fR28
Basically, "if not empty make multiple IAM roles"

I'm not positive that this is the best way of accomplishing this, though - it might just be best to optionally include the IAM role as part of the workers group settings.

aws_auth.tf
```diff
@@ -16,11 +16,20 @@ resource "null_resource" "update_config_map_aws_auth" {
   count = "${var.manage_aws_auth ? 1 : 0}"
 }
 
+data "template_file" "worker_role_arns" {
+  count    = "${var.worker_group_count}"
+  template = "${file("${path.module}/templates/worker-role.tpl")}"
+
+  vars {
+    worker_role_arn = "${lookup(var.worker_groups[count.index], "iam_role_id",  lookup(local.workers_group_defaults, "iam_role_id", "") != "" ? lookup(local.workers_group_defaults, "iam_role_id", "") : aws_iam_instance_profile.workers.id)}"
```
laverya: This needs a fix to return ARN and not just ID

laverya:
@max-rocket-internet I think this is good to review - moved to the "allow specifying IAM roles in worker group config" model

README.md
```diff
@@ -133,8 +133,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | cluster_version | The Kubernetes server version for the EKS cluster. |
 | config_map_aws_auth | A kubernetes configuration to authenticate to this EKS cluster. |
 | kubeconfig | kubectl config file contents for this EKS cluster. |
-| worker_iam_role_arn | IAM role ID attached to EKS workers |
-| worker_iam_role_name | IAM role name attached to EKS workers |
+| worker_iam_role_arn | default IAM role name for EKS worker groups |
```
max-rocket-internet: This should say "default IAM role ARN..."

outputs.tf
```diff
@@ -55,11 +55,11 @@ output "worker_security_group_id" {
...
 
 output "worker_iam_role_name" {
-  description = "IAM role name attached to EKS workers"
+  description = "default IAM role name for EKS worker groups"
   value       = "${aws_iam_role.workers.name}"
 }
 
 output "worker_iam_role_arn" {
-  description = "IAM role ID attached to EKS workers"
+  description = "default IAM role name for EKS worker groups"
```
max-rocket-internet: Shouldn't this be "..role ARN..." too?

local.tf
```diff
@@ -30,6 +30,7 @@ locals {
     autoscaling_enabled           = false                           # Sets whether policy and matching tags will be added to allow autoscaling.
     additional_security_group_ids = ""                              # A comman delimited list of additional security group ids to include in worker launch config
     protect_from_scale_in         = false                           # Prevent AWS from scaling in, so that cluster-autoscaler is solely responsible.
+    iam_role_id                   = ""                              # Use the specified IAM role if set.
```
max-rocket-internet: Can we not just set `aws_iam_role.workers.id` here to make your code for the `role` attribute of `resource.aws_iam_instance_profile.workers` simpler?

outputs.tf
laverya: Will fix - this is what generates the description in the README

local.tf
laverya: Yes that would - will do

laverya:
@max-rocket-internet I think I've fixed everything - let me know if there's anything else to do!

syst0m:
Hi guys, 
found this PR while digging through repo issues related to my usecase.
See here:
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/200

Would be very grateful if you could provide an example on how to attach this policy required by alb-ingress-controller to a worker group.

Thanks i advance! :)

syst0m:
Also,
what does this do exactly:
`  default_iam_role_id      = "${element(concat(aws_iam_role.workers.*.id, list("")), 0)}"`
?

Taken from here:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/local.tf

Concatenates roles from all workers, separated by "" and then takes the element 0?
What's the logic behind this?

laverya:
Responded to the first comment in #200
The second is answered by the PR it was introduced in: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/147

PR #142, opened 2018-09-27
Add option to change worker placement_tenancy.

monsterxx03:
# PR o'clock

## Description

Support launching worker node with non default tenancy.

Some companies have compliance requirements,  ec2 instances must be dedicated instances.

Without specific `placement_tenacy`, the value will keep emtpy(same as specified as `default`), won't show any surprising diff output for users not using before

### Checklist


max-rocket-internet:
Thanks for the PR @monsterxx03 ! Can you please review the checklist and do some more of those items? e.g. at least put your change in the change log.

PR #148, opened 2018-10-02
Allowing 443 to nodes from EKS service

max-rocket-internet:
# PR o'clock

## Description

This allows the EKS cluster service to connect to the nodes on port 443. This is require to run the metrics-server, details [here](https://github.com/kubernetes-incubator/metrics-server/issues/45). Metrics-server is required to run horrizontal pod autoscalers.

AWS CFN is here: https://github.com/awslabs/amazon-eks-ami/blob/master/amazon-eks-nodegroup.yaml#L251-L260

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - A useful addition (slam dunk, @self 🔥)
 - Worker groups can be created with a specified IAM profile. (from @laverya)
 - exposed `aws_eks_cluster` create and destroy timeouts (by @RGPosadas)
+- Allow port 443 from EKS service to nodes to run `metrics-server`. (by @max-rocket-internet)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


max-rocket-internet:
> had been doing this after the fact with module outputs.

Nice. I think it's reasonable to have this access in place by default.

> Any reason not to tighten up the control plan egress rules to match that latest cloud formation as well

Sure but what do you mean exactly? In the AWS CFN port 1025-65535 is allowed from EKS service to nodes. And this module is the same?

mmcaya:
Current module uses egress to the internet from control plane by default:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/cluster.tf#L25
vs the more granular control plane egress rules described by the cloud formation template

max-rocket-internet:
@mmcaya OK good idea. I'll do that in a separate PR.

mmcaya:
great, thanks!

max-rocket-internet:
Paging @brandoconnor 🙂

brandoconnor:
Thanks as always @max-rocket-internet and @mmcaya

Issue #144, opened 2018-09-28
Cannot create a second cluster: Service role name AWSServiceRoleForElasticLoadBalancing has been taken in this account, please try a different suffix.


bkmeneguello:
# I have issues

## I'm submitting a...

* [x] bug report

## What is the current behavior?

I can't create a second cluster because it throws this error: "Service role name AWSServiceRoleForElasticLoadBalancing has been taken in this account, please try a different suffix"

## If this is a bug, how to reproduce? Please include a code sample if relevant.

try to create a second cluster with the same configs but a different name.

## What's the expected behavior?

It should be able to create as many clusters I wish

## Environment details

* Affected module version: v1.4.0
* OS: Linux Mint 19 Tara
* Terraform version: v0.11.8

## Any other relevant info

The error happens with this resource: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/3b546a9c1bc9fab24b3bd0ca5ec30b5defed5e50/cluster.tf#L62-L65

max-rocket-internet:
Another issue related to the `aws_iam_service_linked_role`. We also have this older issue: 
 https://github.com/terraform-aws-modules/terraform-aws-eks/issues/132

okgolove:
Same situation. Want to have multiple clusters in one AWS account.

max-rocket-internet:
So I think to resolve this error you need to delete any ELBs created by the first cluster, then delete the service linked role: `terraform destroy -target=module.xxx.aws_iam_service_linked_role.elasticloadbalancing`

Then allow the clusters to create new ELBs and they should just create their own service linked roles.

Can you guys test that?

okgolove:
I just have set create_elb_service_linked_role = false and it works fine

max-rocket-internet:
PR is now merged so I'll close this issue.

Issue #132, opened 2018-09-12
Cannot Destroy Role: `module.eks.aws_iam_service_linked_role.elasticloadbalancing`


Jeeppler:
# I have issues

## I'm submitting a...

* [x] bug report

## What is the current behavior?

While destroying the cluster I get the following error:

```
Error: Error applying plan:

2 error(s) occurred:

* module.eks.aws_iam_service_linked_role.elasticloadbalancing (destroy): 1 error(s) occurred:

* aws_iam_service_linked_role.elasticloadbalancing: Error waiting for role (arn:aws:iam::<my-account>:role/aws-service-role/elasticloadbalancing.amazonaws.com/AWSServiceRoleForElasticLoadBalancing) to be deleted: unexpected state 'FAILED', wanted target 'SUCCEEDED'. last error: %!s(<nil>)
* module.vpc.aws_vpc.this (destroy): 1 error(s) occurred:

* aws_vpc.this: DependencyViolation: The vpc 'vpc-0df03bd251f8445a6' has dependencies and cannot be deleted.
	status code: 400, request id: 91485bb5-a679-4617-beb6-f4bb3112cdf4
```

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Set `create_elb_service_linked_role` to true:

```
module "eks" {
  source = "terraform-aws-modules/eks/aws"
  version = "1.6.0"

  ...

  # Create an ELB service role to be able to deploy load balancer
  create_elb_service_linked_role = true

  ...
}
```

## What's the expected behavior?

`module.eks.aws_iam_service_linked_role.elasticloadbalancing` can be destroyed/deleted without 
any problems.

## Environment details

* Affected module version: 1.6.0
* OS: Ubuntu
* Terraform version: v0.11.8

max-rocket-internet:
Can you run `aws iam get-service-linked-role-deletion-status` and related commands to see why it didn't delete? I think you need to get the `deletion-task-id`

Jeeppler:
I am not seeing a way to get the `deletion-task-id` from the Terraform error. 

What I tried:

```
$ aws iam delete-service-linked-role --role-name AWSServiceRoleForElasticLoadBalancing
aws iam delete-service-linked-role --role-name AWSServiceRoleForElasticLoadBalancing
{
[...]

PR #160, opened 2018-10-11
Removing aws_iam_service_linked_role from module

max-rocket-internet:
# PR o'clock

## Description

Removing this resource. I think it's not required and seems to be the source of various errors anyway.

It was added here: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/91

And issues here:
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/132
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/144

If anyone gets an error about removing the `aws_iam_service_linked_role.elasticloadbalancing` resource, I think you can just remove it from the state: `terraform state rm aws_iam_service_linked_role.elasticloadbalancing`

### Checklist


tonyxiao:
@max-rocket-Internet this used to be required to fix the issue of EKS not being able to create ELB. Is that no longer an issue? How do you know we didn’t introduce a regression here?

max-rocket-internet:
> Is that no longer an issue? How do you know we didn’t introduce a regression here?

Because I tested it. Also, the default AWS IAM policy `arn:aws:iam::aws:policy/AmazonEKSClusterPolicy` includes this now:

```json
        {
            "Effect": "Allow",
            "Action": "iam:CreateServiceLinkedRole",
            "Resource": "*",
            "Condition": {
                "StringLike": {
                    "iam:AWSServiceName": "elasticloadbalancing.amazonaws.com"
                }
            }
        }
```

I would be curious if you still have that error when your cluster is using the default AWS IAM policy.

tonyxiao:
Ah ok. Got it thank you.

PR #163, opened 2018-10-12
Adjust the order and correct/update the info

chenrui333:
- adjust the ec2 instance type order
- update/correct the EBS opt EC2 instance type information.

Two references:
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html
- https://www.ec2instances.info


local.tf
```diff
@@ -41,125 +41,126 @@ locals {
...
+    "m5d.4xlarge"  = true
+    "m5d.12xlarge" = true
+    "m5d.24xlarge" = true
     "p2.xlarge"    = true
-    "p3.16xlarge"  = true
+    "p2.8xlarge"   = true
+    "p2.16xlarge"  = true
     "p3.2xlarge"   = true
     "p3.8xlarge"   = true
-    "r3.2xlarge"   = false
```
chenrui333: dup and wrong info line.

local.tf
```diff
@@ -41,125 +41,126 @@ locals {
...
-    "t2.large"     = false
-    "t2.medium"    = false
-    "t2.micro"     = false
     "t2.nano"      = false
+    "t2.micro"     = false
     "t2.small"     = false
+    "t2.medium"    = false
+    "t2.large"     = false
     "t2.xlarge"    = false
-    "t3.small"     = false
```
chenrui333: t3 types are all ebs_optimized now.

max-rocket-internet:
Thanks @chenrui333 ! Really nice. I like the ordering and the extra effort to look up all this must have taken some time.

chenrui333:
Sure, I will put my change into the changelog.

chenrui333:
@max-rocket-internet  Do you think we need to sort out the order of the instance type alphabetically?

Issue #109, opened 2018-08-28
Open PR to remove providers from module


gb-ckedzierski:
# I have issues

## I'm submitting a...

* [ ] bug report
* [X] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
Currently, because of providers defined at the module level, you cannot delete the module to remove your infrastructure. Conversely, you have to use commands to remove items from the state file before you can delete. See #108 

Per these 2 issues, it seems after terraform 0.11 these providers should no longer be in modules.
https://github.com/hashicorp/terraform/issues/16824
https://github.com/hashicorp/terraform/issues/17928

## If this is a bug, how to reproduce? Please include a code sample if relevvant.
User Terraform > 0.11 and add the module to create infra, after the apply remove the code, you will get error.

## What's the expected behavior?
All infra in this module is deleted.

## Are you able to fix this problem and submit a PR? Link here if you have already.
Yes, but don't have permissions to contribute. eg create branch for a PR.

## Environment details

* Affected module version: terraform-aws-eks
* OS: N/A
* Terraform version: v0.11.7

## Any other relevant info
Something like this should fix the issues, I tested locally and it works on my machine.
```
provider "null" {
  version = "~> 1.0"
  alias = "default"
}

provider "template" {
  version = "~> 1.0"
  alias = "default"
}

module "eks" {
  source                     = terraform-aws-modules/eks/aws"
  cluster_name          = "development"
  subnets                   = "${module.vpc_app.private_subnets}"
  tags                         = "${map("Environment", "${var.stage}")}"
  vpc_id                      = "${module.vpc_app.vpc_id}"
  manage_aws_auth   = false

  providers = {
    aws = "aws.default"
    null = "null.default"
    template = "template.default"
  }
}
```

Next __remove__ these lines from main.tf
```
...
- provider "null" {}
- provider "template" {}
```

max-rocket-internet:
Is this really standard practice now? We need to pass providers into every module?

dpiddockcmp:
> Is this really standard practice now? We need to pass providers into every module?

No.

The primary provider instance does not need an `alias` specified. You only need to use `alias` if you are dealing with multiple instances of the same provider (e.g. default aws provider is in `us-east-1` and you also want to add DR resources in `us-west-2`).

```hcl
provider "aws" {
  region = "us-east-1"
}

provider "aws" {
  region = "us-east-2"
  alias  = "dr"
[...]

PR #168, opened 2018-10-16
Removing 2 providers from the module

max-rocket-internet:
# PR o'clock

## Description

Removing providers from module to make deletion work correctly.

See https://github.com/terraform-aws-modules/terraform-aws-eks/issues/109

### Checklist


antonbabenko:
Good PR.

I think that logical providers (external, null, http, template) can be present in the module because they have no configuration options other than the definition of the required version (https://www.terraform.io/docs/configuration/providers.html#provider-versions).

In this module, it is not a problem, so it is a good idea to remove them.

PR #161, opened 2018-10-12
Update version

chenrui333:
- Simplify some travis build process
- Ensure download the latest terraform version
- Add some flexibility to the `required_version`

PR #181, opened 2018-10-24
Fix the terraform download version

chenrui333:


chenrui333:
relates to #161, just to check the terraform download flow.

Closing the PR now.

PR #187, opened 2018-11-07
Add wait_nodes_max_tries to wait for nodes to be available before applying the kubernetes configurations

rmakram-ims:
# PR o'clock

## Description

Add wait_nodes_max_tries to wait for nodes to be available before applying the kubernetes configurations

### Checklist


max-rocket-internet:
Hi @rmakram-ims,
Is this to solve a particular problem? I've destroyed and recreated EKS clusters with this module many many times and never had `kubectl` timeout after cluster creation. Have you had this problem? What was the exact error message?

rmakram-ims:
Hi @max-rocket-internet ,
Yeah, I was getting the following error:
```
Error applying plan:
1 error(s) occurred:
* module.test_eks_module.module.eks.null_resource.update_config_map_aws_auth: Error running command 'kubectl apply -f ./config-map-aws-auth_terratest_bfdlhtl1cfnion1pqqfg.yaml --kubeconfig ./kubeconfig_terratest_bfdlhtl1cfnion1pqqfg': exit status 1. Output: Unable to connect to the server: dial tcp 34.195.179.27:443: i/o timeout


Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```
I would see this happening intermittently.

max-rocket-internet:
I think from that error you have some network connectivity problem. Because it's resolved the endpoint correctly but you are not able to connect to it. And the API hasn't returned an error like "service not ready", you are simply not able to connect to the AWS IP.

Do you get what I mean?

rmakram-ims:
Thanks for the comment, I will check again on my side and will update this PR if needed.

pkoraca:
Had the same issue: It failed on `kubectl apply` with same `i/o timeout` message. Tried to run it again and it was able to complete successfully. 
It does not happen on every run, but I'd say approx in 30%.

max-rocket-internet:
OK but I think `i/o timeout` is a local problem, not really a problem with this module or the EKS service.

Also I've built and destroyed around 50 clusters with this module, even 4 new clusters this week, and never had `kubectl` timeout before. So I'm a little reluctant to add 3 resources, another script and variable to solve this problem unless it's super common.

Have you guys tested the `--request-timeout` option for `kubectl`?
Does `kubectl` timeout regularly, after cluster creation?

voigt:
Hey,
exactly the same issue here. @max-rocket-internet totally see your point. What about making `module.test_eks_module.module.eks.null_resource.update_config_map_aws_auth` optional? or allowing to sleep inbetween :)

Creating cluster from germany in `us-east1` and run regularly in this timeout issue.

max-rocket-internet:
OK it seems some other people also have this problem (e.g @Jeeppler in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/199) so let's get something merged 😎

Can we greatly simplify this though? I think 3 new resources, another script and variable is a little overkill.

Is there anything wrong with this?

```
  depends_on = ["aws_eks_cluster.this"]

  provisioner "local-exec" {
    command = "for i in {1..5}; do kubectl apply -f ${var.config_output_path}config-map-aws-auth_${var.cluster_name}.yaml --kubeconfig ${var.config_output_path}kubeconfig_${var.cluster_name} && break || sleep 10; done"
  }
...
```

max-rocket-internet:
@Dolape1983 do you have network problem where you are running terraform? Like a bad connection? Does increasing `{1..5}` or `sleep 10` help for you?

Dolape1983:
Hi @max-rocket-internet , I'm running mi pipeline in gitlab.com and I have the same issue. I think that problem is regarding about permissions, because I'm using aws_security enabled. Locally I don't have the issue. I'm investigating about the base setup for use aws security and eks. If I have news I'll share it with you. 

Warm Regards.

RothAndrew:
The script in the null resource breaks the module on Windows. It is now a shell-specific script, which won't work in `cmd`. This should be reverted or fixed immediately.

PR #209, opened 2018-12-12
Adding new "--with-aggregate-type-default" option to terraform-docs

max-rocket-internet:
# PR o'clock

## Description

Just a small change to add the new `--with-aggregate-type-defaults` option to terraform-docs instructions.

### Checklist

- [ ] `terraform fmt` and `terraform validate` both work from the root and `examples/eks_test_fixture` directories (look in CI for an example)
- [ ] Tests for the changes have been added and passing (for bug fixes/features)
- [ ] Test results are pasted in this PR (in lieu of CI)
- [ ] Docs have been updated using `terraform-docs` per `README.md` instructions
- [ ] I've added my change to CHANGELOG.md
- [ ] Any breaking changes are highlighted above

PR #211, opened 2018-12-12
Use sed instead of tail for terraform-docs command

RothAndrew:
# PR o'clock

## Description

Use `sed` instead of `tail` for the `terraform-docs` command since it is more universally compatible. Thanks to @dpiddockcmp for your suggestion at https://github.com/terraform-aws-modules/terraform-aws-eks/pull/209#issuecomment-446671578

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Documentation should be modified within `main.tf` and generated using [terraform
 Install it with `go get github.com/segmentio/terraform-docs` or `brew install terraform-docs` and then use it to generate the `README.md` file:
 
 ```
-terraform-docs --with-aggregate-type-defaults md ./ | cat -s | tail -r | tail -n +2 | tail -r > README.md
+terraform-docs --with-aggregate-type-defaults md ./ | sed -e '$ d' -e 'N;/^\n$/D;P;D' > README.md
 ```
 
 ## Contributing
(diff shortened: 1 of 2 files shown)


RothAndrew:
Not my call, I'm just a peon helping out, but I like that idea.

antonbabenko:
@max-rocket-internet What do you think?

max-rocket-internet:
@antonbabenko 

> Guys, I may sound strange, but why don't you want to use the same pre-commit-terraform which I am using in other terraform-aws-modules.

Sure thing, let's do it 👍

Will it work on multiple platforms like what this PR is solving?

> Let me know if you want me to help with it and I can open a PR.

Please do!

RothAndrew:
Oops wrong button...

RothAndrew:
> @antonbabenko
> 
> > Guys, I may sound strange, but why don't you want to use the same pre-commit-terraform which I am using in other terraform-aws-modules.
> 
> Sure thing, let's do it 
> 
> Will it work on multiple platforms like what this PR is solving?
> 
> > Let me know if you want me to help with it and I can open a PR.
> 
> Please do!

Both options should be available and equivalent. The pre-commit needs to be installed locally for it to work.

max-rocket-internet:
OK well let's merge this now as it solves a problem and await a nice PR from Anton.

PR #217, opened 2018-12-14
Added pre-commit-terraform to update docs

antonbabenko:
# PR o'clock

## Description

Added pre-commit-terraform to update docs

### Checklist


RothAndrew:
Due to the fact that documentation is already set up in main.tf, I'd prefer to see the new `terraform_docs_replace` hook used over `terraform_docs`. There are two advantages I see for doing so:

1. Users who can't/wont install and use `pre-commit` can still run `terraform-docs` in the same way as has been supported in the past
1. It's a less drastic change over how the documentation has been handled up until now for this module.

I would submit a PR but I don't want to step on @antonbabenko's toes, since it would directly conflict with this PR.

Disclaimer: I wrote the new `terraform_docs_replace` hook.

antonbabenko:
This PR is ready to merge. It is a simple one-time replacement of content of `main.tf`, that's it.

If a developer can't install `pre-commit` we can't help him/her really. I'm not interested in making and then maintaining several solutions to such an easy task which is already automated for more than a year in dozens of other repositories (not only those I maintain).

PR #218, opened 2018-12-14
ignore target-group-arn changes

jpke:
# PR o'clock

## Description

Create an EKS cluster through this project defaults to no ALB target group attachments. While these can be added through setting target_group_arns  in worker_groups, we have another terraform project managing ingress ALBs. This other module attaches ASGs to ALBs via aws_autoscaling_attachment resources.

The problem arises when the module managing ingress ALBs and ASG attachments runs, and then this module is rerun. This module will remove the ASG attachments created by the other module.

Adding 'target_group_arns' to 'ignore_changes' in the worker lifecycle config fixes this problem.

### Checklist


max-rocket-internet:
Could this change affect other people who are using `target_group_arns` for whatever reason?

@brandoconnor I think wrote the TF ALB module so I'll defer to him.

jpke:
incorporated feedback into another approach, sidestepping this issue. closing and cancelling pr

grahamar:
@max-rocket-internet `target_group_arns` is not configurable for the workers ASG, so since it's a value that can only be configured outside of this module it should be ignored.

Your comment here: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/219#issuecomment-447765926 is not possible since it's not a variable that's consumed here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers.tf#L1

My use case for this is that my kubernetes ingress controller manages the target group & ELB, therefore running an update will clobber this. Even if `target_group_arns` was configurable in the workers ASG I would assume for many who use kubernetes service resources of type `LoadBalancer` or like me is managed via a Kubernetes controller this will be a blocker.

Please let me know if I've made an incorrect assumption above 😄but as far as i can tell this should be safe to make for everyone and essential for those of us who's target groups are managed in K8s.

max-rocket-internet:
Hey @grahamar!
But is in the configuration, it's [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers.tf#L6)?

So if this PR was merged, then someone who wanted to change the `target_group_arns` for one or more of their worker_groups, then it wouldn't work.

Am I misunderstanding something here?

> My use case for this is that my kubernetes ingress controller manages the target group & ELB, therefore running an update will clobber this

Understood but are you sure that's the case?

> I would assume for many who use kubernetes service resources of type `LoadBalancer`

This will just create a new ELB, associated it with subnets that have the correct tags and add all cluster nodes to the ELB. Nothing to do with the ASG AFAIK.

grahamar:
🤦‍♂️no idea how I missed it in the configuration! Thanks, that at least is a workaround.

Again my apologies it's just the official AWS ALB controller not the AWS kubernetes default provider so you're right about services of type `LoadBalancer`

I'll manually hard code my target groups for now until https://github.com/hashicorp/terraform/issues/3116 is fixed. Thanks @max-rocket-internet !

max-rocket-internet:
No worries at all!

But still, I'm curious about this

1. If you are using the ALB controller, the controller will create the ALB and do the association with the ASG itself, right?
2. In that scenario, does TF undo that association next time it is run?

grahamar:
I'm using https://github.com/zalando-incubator/kube-ingress-aws-controller which also does the ASG association, but have used the AWS ALB controller previously.

Correct. Running terraform removed the association as my config has nothing specified for `target_group_arns` since it's managed by K8s. This only started since I updated the module from 1.7.0.

max-rocket-internet:
I see.

Hopefully in the next release of Terraform we can have interpolations in lifecycle rules.

Or perhaps there is more people in your situation and it would warrant merging this PR making other users update `target_group_arns` manually.

Issue #194, opened 2018-11-20
Support EC2 Fleets (manage On-Demand, RI, Spot Instances with a single ASG)


geota:
# I have issues

## I'm submitting a...

* [ ] bug report
* [X] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
This module only provisions vanilla ASGs.

## What's the expected behavior?
Support EC2 Fleets with Spot + ASG support.
Amazon just shipped support for managing RI, Spot, and On-Demand in a single ASG. I would like to leverage these ASGs inside of my EKS worker groups.

## Are you able to fix this problem and submit a PR? Link here if you have already.
Happy to submit a PR, but want to get general alignment that this would be merged/accepted.
## Environment details

* Affected module version:
* OS:
* Terraform version:

## Any other relevant info
https://github.com/kubernetes/autoscaler/issues/838
https://github.com/terraform-providers/terraform-provider-aws/issues/6478
https://aws.amazon.com/about-aws/whats-new/2018/11/scale-instances-across-purchase-options-in-a-single-ASG/

max-rocket-internet:
I would love this also 🎉

skang0601:
So I took a crack at converting it but there seems to be some bugs in the aws provider that is currently blocking it, mainly this: https://github.com/terraform-providers/terraform-provider-aws/issues/4570

max-rocket-internet:
Launch templates have been added so this is now possible 🎉

PR #222, opened 2018-12-19
Adding support for Launch Templates

skang0601:
…licy.

# PR o'clock

## Description

Converting from using aws_launc_configuration to aws_launch_template to support mixed_instances_policy in the worker node's autoscaling group.
Addresses https://github.com/terraform-aws-modules/terraform-aws-eks/issues/194

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,10 +136,13 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | worker\_additional\_security\_group\_ids | A list of additional security group ids to attach to worker instances | list | `[]` | no |
 | worker\_create\_security\_group | Whether to create a security group for the workers or attach the workers to `worker_security_group_id`. | string | `true` | no |
 | worker\_group\_count | The number of maps contained within the worker_groups list. | string | `1` | no |
-| worker\_groups | A list of maps defining worker group configurations. See workers_group_defaults for valid keys. | list | `[ { "name": "default" } ]` | no |
+| worker\_group\_launch\_template\_count | The number of maps contained within the worker_groups_launch_template list. | string | `0` | no |
+| worker\_groups | A list of maps defining worker group configurations to be defined using AWS Launch Configurations. See workers_group_defaults for valid keys. | list | `[ { "name": "default" } ]` | no |
+| worker\_groups\_launch\_template | A list of maps defining worker group configurations to be defined using AWS Launch Templates. See workers_group_defaults for valid keys. | list | `[ { "name": "default" } ]` | no |
 | worker\_security\_group\_id | If provided, all workers will be attached to this security group. If not given, a security group will be created with necessary ingres/egress to work with the EKS cluster. | string | `` | no |
 | worker\_sg\_ingress\_from\_port | Minimum port number from which pods will accept communication. Must be changed to a lower value if some pods in your cluster will expose a port lower than 1025 (e.g. 22, 80, or 443). | string | `1025` | no |
 | workers\_group\_defaults | Override default values for target groups. See workers_group_defaults_defaults in locals.tf for valid keys. | map | `{}` | no |
+| workers\_group\_launch\_template\_defaults | Override default values for target groups. See workers_group_defaults_defaults in locals.tf for valid keys. | map | `{}` | no |
 | write\_kubeconfig | Whether to write a Kubectl config file containing the cluster configuration. Saved to `config_output_path`. | string | `true` | no |
 
 ## Outputs
@@ -158,4 +161,5 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | worker\_security\_group\_id | Security group ID attached to the EKS workers. |
 | workers\_asg\_arns | IDs of the autoscaling groups containing workers. |
 | workers\_asg\_names | Names of the autoscaling groups containing workers. |
+
 <!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -21,6 +21,15 @@ resource "null_resource" "update_config_map_aws_auth" {
 
 data "aws_caller_identity" "current" {}
 
+data "template_file" "launch_template_worker_role_arns" {
+  count    = "${var.worker_group_launch_template_count}"
+  template = "${file("${path.module}/templates/worker-role.tpl")}"
+
+  vars {
+    worker_role_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${element(aws_iam_instance_profile.workers_launch_template.*.role, count.index)}"
+  }
+}
+
 data "template_file" "worker_role_arns" {
   count    = "${var.worker_group_count}"
   template = "${file("${path.module}/templates/worker-role.tpl")}"
@@ -34,7 +43,7 @@ data "template_file" "config_map_aws_auth" {
   template = "${file("${path.module}/templates/config-map-aws-auth.yaml.tpl")}"
 
   vars {
-    worker_role_arn = "${join("", distinct(data.template_file.worker_role_arns.*.rendered))}"
+    worker_role_arn = "${join("", distinct(concat(data.template_file.launch_template_worker_role_arns.*.rendered, data.template_file.worker_role_arns.*.rendered)))}"
     map_users       = "${join("", data.template_file.map_users.*.rendered)}"
     map_roles       = "${join("", data.template_file.map_roles.*.rendered)}"
     map_accounts    = "${join("", data.template_file.map_accounts.*.rendered)}"
--- a/examples/eks_test_fixture/main.tf
+++ b/examples/eks_test_fixture/main.tf
@@ -3,7 +3,7 @@ terraform {
 }
 
 provider "aws" {
-  version = ">= 1.24.0"
+  version = ">= 1.47.0"
   region  = "${var.region}"
 }
 
@@ -41,15 +41,24 @@ locals {
 
   worker_groups = [
     {
-      instance_type       = "t2.small"
-      additional_userdata = "echo foo bar"
-      subnets             = "${join(",", module.vpc.private_subnets)}"
+      # This will launch an autoscaling group with only On-Demand instances
+      instance_type        = "t2.small"
+      additional_userdata  = "echo foo bar"
+      subnets              = "${join(",", module.vpc.private_subnets)}"
+      asg_desired_capacity = "2"
     },
+  ]
+  worker_groups_launch_template = [
     {
-      instance_type                 = "t2.small"
-      additional_userdata           = "echo foo bar"
-      subnets                       = "${join(",", module.vpc.private_subnets)}"
-      additional_security_group_ids = "${aws_security_group.worker_group_mgmt_one.id},${aws_security_group.worker_group_mgmt_two.id}"
+      # This will launch an autoscaling group with only Spot Fleet instances
+      instance_type                            = "t2.small"
+      additional_userdata                      = "echo foo bar"
+      subnets                                  = "${join(",", module.vpc.private_subnets)}"
+      additional_security_group_ids            = "${aws_security_group.worker_group_mgmt_one.id},${aws_security_group.worker_group_mgmt_two.id}"
+      override_instance_type                   = "t3.small"
+      asg_desired_capacity                     = "2"
+      spot_instance_pools                      = 10
+      on_demand_percentage_above_base_capacity = "0"
     },
   ]
   tags = {
@@ -133,7 +142,9 @@ module "eks" {
   tags                                 = "${local.tags}"
   vpc_id                               = "${module.vpc.vpc_id}"
   worker_groups                        = "${local.worker_groups}"
-  worker_group_count                   = "2"
+  worker_groups_launch_template        = "${local.worker_groups_launch_template}"
+  worker_group_count                   = "1"
+  worker_group_launch_template_count   = "1"
   worker_additional_security_group_ids = ["${aws_security_group.all_worker_mgmt.id}"]
   map_roles                            = "${var.map_roles}"
   map_roles_count                      = "${var.map_roles_count}"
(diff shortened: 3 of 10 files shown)


workers.tf
```diff
@@ -1,10 +1,80 @@
...
+      spot_max_price                           = "${var.spot_max_price}"
+    }
+
+    launch_template {
+      launch_template_specification {
+        launch_template_id = "${element(aws_launch_template.workers.*.id, count.index)}"
+        version            = "$Latest"
+      }
+
+      override {
```
skang0601: It seems like the AWS provider devs are also waiting on Terraform 0.14 to fix everything as well...

It's currently not possible to make this overrides dynamically generated and you're required to have 2 or more of these overrides.

workers.tf
```diff
@@ -1,10 +1,80 @@
...
+  }
+
+  count = "${var.worker_group_count}"
+
+  lifecycle {
+    create_before_destroy = true
+  }
+
+  block_device_mappings {
+    device_name = "${data.aws_ami.eks_worker.root_device_name}"
```
skang0601: I'm a little worried about this. 

I could generate a list of the problem root device name by using a template file trick but I'm not sure what you all might prefer.

max-rocket-internet:
Hey @skang0601,

Super excited to have Launch Template support!

Do you think we could implement in a way that does not replace the existing  arrangement with `aws_launch_configuration`? Like could we choose on a per worker group basis whether a group uses a `aws_launch_configuration` or `aws_launch_template`?

Reason being this is quite a large change with a lot of extra options and I think it might be nice to have it optional for a while so people can test it out.

What do you think? And others?!?

skang0601:
@max-rocket-internet I think that's a valid ask. I took the naive approach and just created a separate input for worker groups to be created using the launch template.
Either in the future, this can probably be cleaned up further by switching to launch templates entirely or when some of the newer HCL features finally land on a Terraform release.

examples/eks_test_fixture/main.tf
```diff
@@ -39,18 +39,27 @@ locals {
...
+
+  launch_template_worker_groups = [
     {
       instance_type                 = "t2.small"
       additional_userdata           = "echo foo bar"
       subnets                       = "${join(",", module.vpc.private_subnets)}"
       additional_security_group_ids = "${aws_security_group.worker_group_mgmt_one.id},${aws_security_group.worker_group_mgmt_two.id}"
+      override_instance_type        = "t3.small"
+      asg_desired_capacity          = "2"
+      override_instance_type        = "t3.small"
```
antonbabenko: Same as line 58.

skang0601: Fixed

skang0601:
Made some additional updates.
I moved the mixed_instances_policy param from a global config to per worker group configuration and updated the changelog

runningman84:
is there any eta for release of this feature?

variables.tf
```diff
@@ -60,21 +60,36 @@ variable "vpc_id" {
   description = "VPC where the cluster and workers will be deployed."
 }
 
-variable "worker_groups" {
```
max-rocket-internet: Like I said before, can we make this PR backwards compatible? i.e leave this variable and all related stuff in place so that a user can use launch templates if they want but without destroying existing resources.

workers.tf
```diff
@@ -1,20 +1,47 @@
-resource "aws_autoscaling_group" "workers" {
```
max-rocket-internet: This will cause the existing ASG to be destroyed.

workers.tf
```diff
@@ -24,33 +51,117 @@ resource "aws_autoscaling_group" "workers" {
...
-    volume_type           = "${lookup(var.worker_groups[count.index], "root_volume_type", local.workers_group_defaults["root_volume_type"])}"
-    iops                  = "${lookup(var.worker_groups[count.index], "root_iops", local.workers_group_defaults["root_iops"])}"
+    volume_size           = "${lookup(var.launch_configuration_worker_groups[count.index], "root_volume_size", local.workers_group_defaults["root_volume_size"])}"
+    volume_type           = "${lookup(var.launch_configuration_worker_groups[count.index], "root_volume_type", local.workers_group_defaults["root_volume_type"])}"
+    iops                  = "${lookup(var.launch_configuration_worker_groups[count.index], "root_iops", local.workers_group_defaults["root_iops"])}"
     delete_on_termination = true
   }
 }
 
+resource "aws_iam_instance_profile" "launch_configuration_workers" {
```
max-rocket-internet: Do we need a second instance profile? Can't both LC and LT type ASGs use the same one?

README.md
```diff
@@ -119,8 +124,6 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | tags | A map of tags to add to all resources. | map | `<map>` | no |
 | vpc_id | VPC where the cluster and workers will be deployed. | string | - | yes |
 | worker_additional_security_group_ids | A list of additional security group ids to attach to worker instances | list | `<list>` | no |
-| worker_group_count | The number of maps contained within the worker_groups list. | string | `1` | no |
-| worker_groups | A list of maps defining worker group configurations. See workers_group_defaults for valid keys. | list | `<list>` | no |
```
max-rocket-internet: To make the ever growing list of variables easier to read, perhaps let's keep them grouped like this?

```
worker_groups
worker_group_count
worker_groups_launch_template
worker_group_launch_template_count
```

workers.tf
max-rocket-internet: I don't think we need this?

workers.tf
max-rocket-internet: Ahh, yes, I think it needs to stay.

max-rocket-internet:
Maybe we just create `workers_launch_template.tf` to keep everything clear and separate?

workers.tf
skang0601: Yep, especially since this change will support launching with both types.

variables.tf
skang0601: Sorry, misinterpreted your previous ask there.
Will update PR

workers.tf
skang0601: Fixed

workers.tf
skang0601: The API will complain if you do not specify this. This field is required on launch templates.

README.md
skang0601: Fixed

aws_auth.tf
```diff
@@ -20,7 +20,16 @@ resource "null_resource" "update_config_map_aws_auth" {
 
 data "aws_caller_identity" "current" {}
 
-data "template_file" "worker_role_arns" {
```
max-rocket-internet: Can we leave this resource named the same? And call the second one `worker_launch_template_role_arns` to keep it consistent?

examples/eks_test_fixture/main.tf
```diff
@@ -39,17 +39,26 @@ locals {
   #   },
   # ]
 
-  worker_groups = [
+  launch_configuration_worker_groups = [
```
max-rocket-internet: Can we keep this naming consistent also? i.e. keep this as `worker_groups` and name the new one `worker_groups_launch_template`.

examples/eks_test_fixture/main.tf
```diff
@@ -127,15 +136,17 @@ module "vpc" {
...
-  map_users                            = "${var.map_users}"
-  map_accounts                         = "${var.map_accounts}"
+  source                                  = "../.."
+  cluster_name                            = "${local.cluster_name}"
+  subnets                                 = ["${module.vpc.private_subnets}"]
+  tags                                    = "${local.tags}"
+  vpc_id                                  = "${module.vpc.vpc_id}"
+  worker_groups                          = "${local.launch_configuration_worker_groups}"
+  worker_groups_launch_template           = "${local.launch_template_worker_groups}"
+  worker_group_count = "1"
```
max-rocket-internet: `terraform fmt`

local.tf
```diff
@@ -10,31 +10,39 @@ locals {
   kubeconfig_name          = "${var.kubeconfig_name == "" ? "eks_${var.cluster_name}" : var.kubeconfig_name}"
 
   workers_group_defaults_defaults = {
```
max-rocket-internet: This is a little confusing as now we have the defaults for both worker group types mixed together.

How would someone reading this now what parameter is valid for each type of work group?

How about we leave `workers_group_defaults_defaults` as it is and create `workers_group_launch_template_defaults_defaults`? Does that make sense and work?

workers_launch_template.tf
```diff
@@ -0,0 +1,103 @@
...
+      volume_size           = "${lookup(var.worker_groups_launch_template[count.index], "root_volume_size", local.workers_group_defaults["root_volume_size"])}"
+      volume_type           = "${lookup(var.worker_groups_launch_template[count.index], "root_volume_type", local.workers_group_defaults["root_volume_type"])}"
+      iops                  = "${lookup(var.worker_groups_launch_template[count.index], "root_iops", local.workers_group_defaults["root_iops"])}"
+      delete_on_termination = true
+    }
+  }
+}
+
+resource "aws_iam_instance_profile" "workers_launch_template" {
+  name_prefix = "${aws_eks_cluster.this.name}"
```
max-rocket-internet: Will this name always be unique if someone is using both types of worker groups?

max-rocket-internet:
Also please rebase 🙂

local.tf
skang0601: I thought about that but it would result in some duplication if we wanna make it very separate if you're fine with that.
The other approach is to just merge the two defaults for the launch_template but I think that's as confusing as mixing the two.

max-rocket-internet: Hmmm I see your point about duplication.

At this point I really think the separation is important since launch template is a such a new feature and I don't want it to interfere with the current ASG/LC arrangement until everyone is confident in the ASG/LT approach.

So how about we have `workers_group_defaults_defaults` and `workers_group_launch_template_defaults_defaults` but the latter only contains the special values for LTs? e.g. `spot_max_price`, `spot_instance_pools` etc.

I think this will be clear and also won't change the existing code at all?

workers_launch_template.tf
skang0601: I'm not completely sure what you mean by this.

local.tf
skang0601: I had a chance to think about this for a bit. I ended up just duplicating the defaults along.

workers_launch_template.tf
max-rocket-internet: No worries, I think what you have now is good 🙂

README.md
```diff
@@ -106,56 +106,61 @@ Many thanks to [the contributors listed here](https://github.com/terraform-aws-m
...
+| kubeconfig_name | Override the default name used for items kubeconfig. | string | `` | no |
+| local_exec_interpreter | Command to run for local-exec resources. Must be a shell-style interpreter. If you are on Windows Git Bash is a good choice. | list | `<list>` | no |
+| manage_aws_auth | Whether to write and apply the aws-auth configmap file. | string | `true` | no |
+| map_accounts | Additional AWS account numbers to add to the aws-auth configmap. See examples/eks_test_fixture/variables.tf for example format. | list | `<list>` | no |
+| map_accounts_count | The count of accounts in the map_accounts list. | string | `0` | no |
+| map_roles | Additional IAM roles to add to the aws-auth configmap. See examples/eks_test_fixture/variables.tf for example format. | list | `<list>` | no |
+| map_roles_count | The count of roles in the map_roles list. | string | `0` | no |
+| map_users | Additional IAM users to add to the aws-auth configmap. See examples/eks_test_fixture/variables.tf for example format. | list | `<list>` | no |
+| map_users_count | The count of roles in the map_users list. | string | `0` | no |
```
max-rocket-internet: I don't think any of this needs to be changed so I think you are running on older version of terraform-docs that is removing the `\_`? I'm using version `0.5.0`. Can you install that version and then fix these by running the commit hook again or fixing it manually?

max-rocket-internet:
Thanks @skang0601! I did some testing in my envs and it looks nice. After you fix the readme changes I think we can merge 🎉

README.md
skang0601: Oh wow sorry about that, I didn't realize that it produced this...
Let me update.

skang0601:
> Thanks @skang0601! I did some testing in my envs and it looks nice. After you fix the readme changes I think we can merge 🎉

I believe I've resolved the above issues. @max-rocket-internet

examples/eks_test_fixture/main.tf
```diff
@@ -41,15 +41,24 @@ locals {
...
-      subnets                       = "${join(",", module.vpc.private_subnets)}"
-      additional_security_group_ids = "${aws_security_group.worker_group_mgmt_one.id},${aws_security_group.worker_group_mgmt_two.id}"
+      # This will launch an autoscaling group with only Spot Fleet instances
+      instance_type                            = "t2.small"
+      additional_userdata                      = "echo foo bar"
+      subnets                                  = "${join(",", module.vpc.private_subnets)}"
+      additional_security_group_ids            = "${aws_security_group.worker_group_mgmt_one.id},${aws_security_group.worker_group_mgmt_two.id}"
+      override_instance_type                   = "t3.small"
+      asg_desired_capacity                     = "2"
+      spot_instance_pools                      = 10
```
zytek: Does not really make sense to request 2 instances across 10 spot pools. Such configuration is confusing, especially when put in examples. I would go with something along `asg_desired_capacity = 4` and `spot_instance_pools = 2` - 2x2 instance types = four instances across two spot pools (types).

max-rocket-internet: I'm gonna merge this now as it's been open and on-going for a while now. The LT/ASG arrangement is also completely optional for now and not used by default. By merging this people can at least have a play with it for now.

@zytek perhaps you could make a PR with the changes suggested here?

zytek: Awesome. I'll play with it and can make a PR.
Could you push new version to terraform repository ?

max-rocket-internet: > Could you push new version to terraform repository ?

Sure but in a couple days. This repo has sucked up enough of my time for today. You can always point your module source directly at github until a new release is created 🙂

RothAndrew: FYI:

```
module "eks" {
  source = "git::https://github.com/terraform-aws-modules/terraform-aws-eks.git?ref=whatever_ref_you_want_goes_here"
}

Issue #170, opened 2018-10-17
Feature request: option to not create the aws_auth files


philwinder:
Can I suggest adding the ability to not create the aws_auth files, like you do for the kubeconfig file?

The reason why I would like this is because I use the same module on multiple different clusters. I don't want to store all those aws-auth files because the logic is contained within the code. Since I don't want to store them, I don't have them in git. So each time I re-apply, terraform thinks it needs to recreate then again, even though the actual aws_auth configmap hasn't changed.

(The same applies to the kubeconfig file too, I don't want to create that either)

In the past I have achieved this by piping the manifest into kubectl like this: `printf ${data.manifest.rendered} | kubectl apply...`

But note that this still relies on a kubeconfig file to exist (see #169).

max-rocket-internet:
Sure! Sounds reasonable. PRs welcome 🙂

max-rocket-internet:
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/228

Issue #169, opened 2018-10-17
Bug: using write_kubeconfig="false" and manage_aws_auth="true" causes "no such file"


philwinder:
If you set  write_kubeconfig="false" and manage_aws_auth="true", then the aws_auth file will try to `kubectl apply` the auth configmap. See [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/aws_auth.tf#L11).

But if the kubeconfig file isn't written, then it can't find the kubeconfig.

The easiest fix is to always write the file, but I would like to not create the kubeconfig file for other reasons.

max-rocket-internet:
Understood. Happy to accept a PR for this also.

max-rocket-internet:
@philwinder can you test this PR? https://github.com/terraform-aws-modules/terraform-aws-eks/pull/228

max-rocket-internet:
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/228

PR #228, opened 2018-12-27
Added write_aws_auth_config option

yutachaos:
# PR o'clock

## Description

Added an option not to create aws_auth configmap.
I made it possible to apply kubectl apply without generating a file to correspond.

Refer issues
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/169
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/170

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ##### Changed
 
-- Removed invalid action from worker_autoscaling iam policy
+- Change worker group ASG to use create_before_destroy (by @stefansedich)
 - Fixed a bug where worker group defaults were being used for launch template user data (by @leonsodhi-lf)
 - Managed_aws_auth option is true, the aws-auth configmap file is no longer created, and write_aws_auth_config must be set to true to generate config_map. (by @yutachaos)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,13 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Ability to configure force_delete for the worker group ASG (by @stefansedich)
 - Added EBS optimized mapping for the g3s.xlarge instance type (by @stefansedich)
 - `enabled_metrics` input (by @zanitete)
+- write_aws_auth_config to input (by @yutachaos)
 
 ##### Changed
 
-- Change worker group ASG to use create_before_destroy (by @stefansedich)
+- Removed invalid action from worker_autoscaling iam policy
 - Fixed a bug where worker group defaults were being used for launch template user data (by @leonsodhi-lf)
+- Managed_aws_auth option is true, the aws-auth configmap file is no longer created, and write_aws_auth_config must be set to true to generate config_map. (by @yutachaos)
 
 # History
 
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -1,19 +1,32 @@
 resource "local_file" "config_map_aws_auth" {
   content  = "${data.template_file.config_map_aws_auth.rendered}"
   filename = "${var.config_output_path}config-map-aws-auth_${var.cluster_name}.yaml"
-  count    = "${var.manage_aws_auth ? 1 : 0}"
+  count    = "${var.write_aws_auth_config ? 1 : 0}"
 }
 
 resource "null_resource" "update_config_map_aws_auth" {
   depends_on = ["aws_eks_cluster.this"]
 
   provisioner "local-exec" {
-    command     = "for i in `seq 1 10`; do kubectl apply -f ${var.config_output_path}config-map-aws-auth_${var.cluster_name}.yaml --kubeconfig ${var.config_output_path}kubeconfig_${var.cluster_name} && exit 0 || sleep 10; done; exit 1"
+    working_dir = "${path.module}"
+
+    command = <<EOS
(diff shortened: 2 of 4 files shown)


aws_auth.tf
```diff
@@ -1,19 +1,32 @@
...
 
 resource "null_resource" "update_config_map_aws_auth" {
   depends_on = ["aws_eks_cluster.this"]
 
   provisioner "local-exec" {
-    command     = "for i in {1..5}; do kubectl apply -f ${var.config_output_path}config-map-aws-auth_${var.cluster_name}.yaml --kubeconfig ${var.config_output_path}kubeconfig_${var.cluster_name} && break || sleep 10; done"
+    working_dir = "${path.module}"
+
+    command = <<EOS
+mkfifo aws_auth_configmap kube_config & \
```
max-rocket-internet: `mkfifo` won't work on Windows.

max-rocket-internet:
How about simply if a user sets `write_kubeconfig=false` and `manage_aws_auth=true` then when TF runs `kubectl` to apply the `aws-auth` configmap, it just removes the `--kubeconfig` option and then it is up to the user solve the authentication themselves?

aws_auth.tf
yutachaos: Thanks for your comment.
For compatibility with Windows, it changed to a method not using mkfifo command.

max-rocket-internet:
I'm still not convinced this is a good approach. Think about these scenarios:

- What if the user already has a file called `aws_auth_configmap.yaml` or `kube_config.yaml` in CWD? It will be overwritten and then deleted.
- What if the user has 2 clusters? TF runs many things in parallel and this could cause either the same config being applied twice or to the wrong cluster.
- What the user runs CTRL-C during run? It could leave the files there.

I think you can solve these by writing the file to `"${file("${path.module}/..}` instead of CWD. Can you test that?

yutachaos:
I am using the local-exec working_dir option to run under module.

FYI
https://www.terraform.io/docs/provisioners/local-exec.html#working_dir

max-rocket-internet:
> I am using the local-exec working_dir option to run under module.

You are right! My mistake, sorry!

> I am assuming to use the aws eks update-kubeconfig command.

You mean the AWS CLI? But in this PR, you are not assuming that. If you were assuming that, then we wouldn't need to write `kube_config` in `update_config_map_aws_auth` because this would be the responsibility of the user. Right? Or am I misunderstanding?

max-rocket-internet:
Since @philwinder opened the original issue, I've asked them to test this PR 🙂

yutachaos:
> You mean the AWS CLI? But in this PR, you are not assuming that. If you were assuming that, then we wouldn't need to write `kube_config` in `update_config_map_aws_auth` because this would be the responsibility of the user. Right? Or am I misunderstanding?

 
Updating using kubeconfig in update_config_map_aws_auth is only used to join eks node to eks cluster.

see
https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

yutachaos:
@max-rocket-internet 
How about this PR?
If further explanation is necessary, I will add it.
Thank you.

CHANGELOG.md
```diff
@@ -12,10 +12,13 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ##### Added
 
 - Ability to configure force_delete for the worker group ASG (by @stefansedich)
+- write_aws_auth_config to input (by @yutachaos)
 
 ##### Changed
 
 - Change worker group ASG to use create_before_destroy (by @stefansedich)
+- Removed invalid action from worker_autoscaling iam policy
```
max-rocket-internet: I think this is a mistake? This change is not in this PR.

max-rocket-internet:
Hi @yutachaos 
I had a look a couple days ago but there were conflicts. And now this PR is undoing the changes in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/245/files for ZSH compatibility 🙂

Fix that and we can merge.

aws_auth.tf
```diff
@@ -1,20 +1,32 @@
...
 
 resource "null_resource" "update_config_map_aws_auth" {
   depends_on = ["aws_eks_cluster.this"]
 
   provisioner "local-exec" {
-    command     = "for i in `seq 1 10`; do kubectl apply -f ${var.config_output_path}config-map-aws-auth_${var.cluster_name}.yaml --kubeconfig ${var.config_output_path}kubeconfig_${var.cluster_name} && exit 0 || sleep 10; done; exit 1"
+    working_dir = "${path.module}"
+
+    command = <<EOS
+for i in {1..5}; do \
```
max-rocket-internet: This was changed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/245 for ZSH

yutachaos: Thank you fixed it.

CHANGELOG.md
yutachaos: Thank you fixed it.

max-rocket-internet:
fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/169
fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/170

a7i:
@max-rocket-internet Hi there, thank you for adding this. Are there any plans to create a Release Tag?

max-rocket-internet:
Just for you @a7i: https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v2.2.0

Issue #224, opened 2018-12-21
Add ssm agent support


runningman84:
## I'm submitting a...

* [x] feature request

## What is the current behavior?

You cannot access any worker node via ssh.

## What's the expected behavior?

It would be cool to access the worker nodes via the SSM session manager. This would help to debug any issues with the worker nodes.

## Are you able to fix this problem and submit a PR? Link here if you have already.

It is quite easy to implement:

role:
```
resource "aws_iam_role_policy_attachment" "node-AmazonEC2RoleforSSM" {
  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM"
  role       = "${aws_iam_role.node.name}"
}
```

userdata:
```
yum install -y amazon-ssm-agent
systemctl start amazon-ssm-agent
systemctl enable amazon-ssm-agent
```

brandoconnor:
PR discussed and closed. Hopefully you're managing to make this work through the flexibility we provide in the userdata variables.

Issue #99, opened 2018-08-22
Should we manage k8s resources with this module?


max-rocket-internet:
This is a general question about the direction of this module.

We get requests that would require this module to manage or create Kubernetes resources. Some examples:

- Modifying CNI configuration before worker ASG creation: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/96
- Deploying cluster autoscaler: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/71
- Manage add-ons: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/19

I think we should have a clear position on this types of issues.

brandoconnor:
^ we're on the same page here. 100% we should avoid adding bits of shell script, here, there, and everywhere to accommodate individual needs because those would stretch a mile long. Providing adequate facilities for others to flexibly add what they'd want on top of what we have (e.g. `pre-userdata` and `post-userdata`) is a better approach and I'm all ears for those sorts of additions. I think what we have will more than likely be flexible enough for the long haul but perhaps you see room for improvement.

I'm a bit more torn if the kubernetes provider gets to the point of useful in conjunction with this module. When/if that happens, I'd at least be interested to see how people would want to use that within the module. My inclination is that most if not all of this would be better suited for a separate state and as a consumer of the module rather than within it, but I'm reserving the chance that some is worth integrating.

Curious to hear any other thoughts you might have. If we're seeing eye to eye on most or all of the above, let's close this out.

max-rocket-internet:
OK cool, I think we are definitely on the same page about this! That's great.

> Providing adequate facilities for others to flexibly add what they'd want on top of what we have (e.g. pre-userdata and post-userdata) is a better approach and I'm all ears for those sorts of additions.

:100:

> I'm a bit more torn if the kubernetes provider gets to the point of useful in conjunction with this module. 

Perhaps later but for now I think it's a little tricky due to providers not supporting any sort of [depends_on](https://github.com/hashicorp/terraform/issues/2430) functionality. We had this when trying to create DBs inside an RDS instance using the PostgreSQL terraform provider.

> My inclination is that most if not all of this would be better suited for a separate state

Agreed.

Anyway, for now I'll close this and we can refer to this for future issues. And if it changes in the future then we can re-evaluate the situation.

rodlogic:
[...]

PR #229, opened 2018-12-27
Feature/ssm agent

runningman84:
# PR o'clock

## Description

This pull requests implements ssm agent support as described here:
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/224

### Checklist

### DIFF
--- a/workers.tf
+++ b/workers.tf
@@ -131,7 +131,7 @@ resource "aws_iam_role_policy_attachment" "workers_AmazonEC2ContainerRegistryRea
   role       = "${aws_iam_role.workers.name}"
 }
 
-resource "aws_iam_role_policy_attachment" "node-AmazonEC2RoleforSSM" {
+resource "aws_iam_role_policy_attachment" "workers_AmazonEC2RoleforSSM" {
   policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM"
   role       = "${aws_iam_role.workers.name}"
 }
--- a/workers.tf
+++ b/workers.tf
@@ -131,6 +131,11 @@ resource "aws_iam_role_policy_attachment" "workers_AmazonEC2ContainerRegistryRea
   role       = "${aws_iam_role.workers.name}"
 }
 
+resource "aws_iam_role_policy_attachment" "node-AmazonEC2RoleforSSM" {
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM"
+  role       = "${aws_iam_role.workers.name}"
+}
+
 resource "null_resource" "tags_as_list_of_maps" {
   count = "${length(keys(var.tags))}"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ## [[v2.0.1](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v2.0.0...HEAD)] - 2019-01-??]
 
 ### Added
- - Write your awesome addition here (by @you)
-
-### Changed
- - Write your awesome change here (by @you)
+ - Installed and enabled ssm agent by default (by @runningman84)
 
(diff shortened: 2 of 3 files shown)


max-rocket-internet:
I'm convinced all users of this module would be happy to have this enabled by default on every cluster. What do other people think?

workers.tf
```diff
@@ -131,6 +131,11 @@ resource "aws_iam_role_policy_attachment" "workers_AmazonEC2ContainerRegistryRea
   role       = "${aws_iam_role.workers.name}"
 }
 
+resource "aws_iam_role_policy_attachment" "node-AmazonEC2RoleforSSM" {
```
antonbabenko: Please rename `node-AmazonEC2RoleforSSM` => `workers_AmazonEC2RoleforSSM`

runningman84: done

runningman84:
> I'm not convinced all users of this module would be happy to have this enabled by default on every cluster. What do other people think?

I think this is the preferred solution from AWS instead off ssh keys which should be disabled by default.

max-rocket-internet:
OK but if [AWS themselves are reluctant to put this in by default](https://github.com/awslabs/amazon-eks-ami/issues/127) and it's very easy to add this to your worker nodes currently (that's what `pre_userdata` and `iam_role_id` are for) then it seems a little heavy handed to install and run an extra service for all users of this module.

Maybe other people can chime in.

hobbsh:
I agree with @max-rocket-internet - it should be optional and `pre_userdata` is a fine place to specify - if my vote means anything. It is not really related to this module nor necessary to run an EKS cluster.

runningman84:
Ok I guess this feature is not needed by all users and will therefore close this pull request

max-rocket-internet:
Hey @runningman84 
Here's a much more "k8s" approach to installing ssm-agent: https://github.com/mumoshu/kube-ssm-agent
🙂

Jeeppler:
Actually, I would like to have AWS SSM support. It would be nice to have AWS SSM support in the EKS cluster module. AWS SSM works on node level and not on container level.

The `kube-ssm-agent` raises some problems in my understanding. The biggest issue I see is that kube-ssm-agent mounts several system files into the container to be able to provide the SSM agent functionality and requires privileged access:

```
        securityContext:
          runAsUser: 0
          privileged: true
        volumeMounts:
        - name: dbus
          mountPath: /var/run/dbus
        - name: run-systemd
          mountPath: /run/systemd
        - name: var-run
          mountPath: /var/run
        - name: etc-sudoers
          mountPath: /etc/sudoers.d
``` 

I personally, would much rather like to see the AWS SSM agent installed directly on the node. 

The changes @runningman84 introduced are pretty small. However, I do agree with @max-rocket-internet regarding installing the AWS SSM agent by default. I would like it to be an optional feature. 

Maybe with a module variable `install_aws_ssm_agent = <false|true>`. `false` should be the default.

max-rocket-internet:
I'm not keen in including it in the module, even if it defaults to `false`.

First, I don't think we should be in the business of installing software. Either on the nodes or in the cluster. @brandoconnor and I already discussed it in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/99

I know here we are talking about software on the nodes here, and not in k8s, but IMO it's the same (or even a little worse). If we start this, I'm sure we will get more PRs for other random system management software.

Second, a daemonset is the *proper* way of doing this in k8s.

> The biggest issue I see is that kube-ssm-agent mounts several system files into the container to be able to provide the SSM agent functionality and requires privileged access

This is normal for a daemonset like this. The extra access and privilege is no different to running it on the host. The [Sysdig helm chart](https://github.com/helm/charts/blob/master/stable/sysdig/templates/daemonset.yaml) would be another example exactly like this. Or look at fluentd charts. Also, using a daemonset means that the resources that are used by the container are accounted for by k8s, just like any other container. If the process is run on the host, outside of k8s, then it is neither accounted for or monitored.

If ssm-agent provides such good functionality then I recommend you create a PR to the helm charts repo 🔥

Third, AWS didn't want to include it in their AMI: https://github.com/awslabs/amazon-eks-ami/issues/127

And my forth and final point: It's super easy to add it already, that's what pre_userdata and iam_role_id are for 😃

PR #230, opened 2018-12-28
Update userdata.sh.tpl with node_labels based on ondemand/spot detection

runningman84:
# PR o'clock

## Description

In order to detect wheter an instance is running ondemand or spot I integrated a small detection logic from @pahud.

This is a change based on this template:
https://github.com/pahud/eks-templates/blob/master/cloudformation/nodegroup.yaml

This will allow us to use his eks lambda drainer in the future:
https://github.com/pahud/eks-lambda-drainer

This change is also relevant for the mixed instances pull request because once you have a mixed fleet you want to know which instances are ondemand or spot:
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/222

### Checklist

### DIFF
--- a/templates/userdata.sh.tpl
+++ b/templates/userdata.sh.tpl
@@ -3,8 +3,17 @@
 # Allow user supplied pre userdata code
 ${pre_userdata}
 
+# Detect instance life cycle
+iid=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+export AWS_DEFAULT_REGION=${AWS::Region}
+ilc=`aws ec2 describe-instances --instance-ids  $iid  --query 'Reservations[0].Instances[0].InstanceLifecycle' --output text`
+
 # Bootstrap and join the cluster
-/etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' --kubelet-extra-args '${kubelet_extra_args}' '${cluster_name}'
+if [ "$ilc" == "spot" ]; then
+  /etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' --kubelet-extra-args --node-labels=ondemand=yes '${kubelet_extra_args}' '${cluster_name}'
+else
+  /etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' --kubelet-extra-args '--node-labels=spotfleet=yes --register-with-taints=spotInstance=true:PreferNoSchedule' '${kubelet_extra_args}' '${cluster_name}'
+end
 
 # Allow user supplied userdata code
 ${additional_userdata}


max-rocket-internet:
Hi @runningman84 
Thanks for the PR!

There's nothing preventing these being added in `kubelet_extra_args` on a per worker group basis right now. Do you think every user will want these labels and taints applied to their clusters by default? 

I'm open for opinions!

runningman84:
> Hi @runningman84
> Thanks for the PR!
> 
> There's nothing preventing these being added in `kubelet_extra_args` on a per worker group basis right now. Do you think every user will want these labels and taints applied to their clusters by default?
> 
> I'm open for opinions!

I think once https://github.com/terraform-aws-modules/terraform-aws-eks/pull/222 is merged you will see a lot of setups where you have a mixed fleet of instances. Therefore it is important to know which instance is ondemand and which instance is spot. An extra arg per worker group would not work because it would use the same arg for spot and on demand.

The question is what is the best name for the label?
https://github.com/pusher/k8s-spot-rescheduler uses:
--on-demand-node-label (default: node-role.kubernetes.io/worker) Name of label on nodes to be considered for draining.
--spot-node-label (default: node-role.kubernetes.io/spot-worker) Name of label on nodes to be considered as targets for pods.

Maybe we should make the label configurable...

max-rocket-internet:
> Maybe we should make the label configurable...

Definitely.

> it is important to know which instance is ondemand and which instance is spot. An extra arg per worker group would not work because it would use the same arg for spot and on demand.

Understood.

I'm hoping either AWS themselves or one of the k8s projects will provide some direction here. Either by adding labels/taints by default, e.g. like `beta.kubernetes.io/instance-type` and `failure-domain.beta.kubernetes.io/zone` etc, or something from AWS in their bootstrap or at least in their own CFN templates.

max-rocket-internet:
I will comment here, feel free to add your input too! https://github.com/awslabs/amazon-eks-ami/issues/85

skang0601:
Hmm I'm not really sure how to interpret the response there. It seems like that they have other plans that would make this solution not viable but it is a feature that I would like now.
To those waiting for something like this, I've been thinking about just setting up separate worker groups that's either all on-demand or spot-fleet to make labeling/scheduling possible. I dunno if at this point I can recommend having this detection logic as part of the userdata script.

max-rocket-internet:
> Hmm I'm not really sure how to interpret the response there

Me neither. I've asked for more clarification.

Either way, do we really want to deviate away from the AWS approach here when what's in this PR can be achieved in `additional_userdata` anyway?

skang0601:
@max-rocket-internet I would recommend not to merge this PR but perhaps making this known through the documentation since this can be achieved regardless of whether this is merged or not.
If this gets merged though it'll have the unintended side effect that this is the recommended method supported by this module.

max-rocket-internet:
> I would recommend not to merge this PR

Agreed. I don't think we want to force some taints on everyone using spot instances.

For future reference, this can be achieved with something like this for `additional_userdata`:

```
iid=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
export AWS_DEFAULT_REGION=${AWS::Region}
ilc=`aws ec2 describe-instances --instance-ids  $iid  --query 'Reservations[0].Instances[0].InstanceLifecycle' --output text`
...
echo "--node-labels=spotfleet=xxx --register-with-taints=spotInstance=true:PreferNoSchedule" > /etc/systemd/system/kubelet.service.d/30-kubelet-extra-args.conf
systemctl start kubelet
```

Issue #202, opened 2018-12-04
kubeconfig file not being written


aimanparvaiz:
# I have issues

## I'm submitting a...

* [* ] bug report
* [ ] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
I am seeing the kubeconfig file content as output but it's not being written to disk. I have AWS_PROFILE set and I am assuming a role mentioned in the .envrc file. Everything works as expected except the kubeconfig file being written to disk. By default write_kubeconfig is true and I am not changing that or the path, also if I copy the kubeconfig content from terraform output and put it in a file I am able to use kubectl.

## If this is a bug, how to reproduce? Please include a code sample if relevant.
module "eks" {
  source = "repo link"
  cluster_name = "core-dev-eks"
  vpc_id = "vpc_id"
  subnets = "subnets"
  worker_group_count = 1
  worker_groups = "${local.worker_groups}"
  tags = {
    Terraform = "true"
    Environment = "dev"
  }

}
  worker_groups = [
    {
      instance_type = "t2.xlarge"
      asg_max = 4
      autoscaling_enabled = true
      asg_desired_capacity = 3
      autoscaling_enabled = true
      subnets = "private_subnets"
    }
  ]
}
Post successfully apply I dont see kubeconfig file in the `./` path.
## What's the expected behavior?
See kubeconfig file in the current dir

## Environment details

* Affected module version: 1.8.0
* OS: macOS 10.14.1
* Terraform version: v0.11.10

stijndehaes:
If you retry terraform apply a second time is the file written the second time?

max-rocket-internet:
OK cool so mystery solved, right?

Issue #199, opened 2018-11-29
Null Resource Update Config Map AWS Auth


Jeeppler:
# I have issues

## I'm submitting a...

* [x] bug report

## What is the current behavior?

Every time, I create a cluster I get this error message:

```
* module.eks.null_resource.update_config_map_aws_auth: Error running command
  'kubectl apply -f /k8s/config-map-aws-auth_cluster.yaml --kubeconfig /k8s/kubeconfig_cluster': exit status 1.
  Output: Unable to connect to the server: dial tcp xxx.20.110.151:443: i/o timeout
```

if I run `terraform apply` again, it will work without any error.

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Create a cluster using the module.

## What's the expected behavior?

No, error message while creating the cluster.

## Are you able to fix this problem and submit a PR? Link here if you have already.

I do not know what causes the issue.

## Environment details

* Affected module version: 1.7.0
* OS: Ubuntu
* Terraform version:

```
Terraform v0.11.10
+ provider.aws v1.48.0
+ provider.local v1.1.0
+ provider.null v1.0.0
+ provider.template v1.0.0
```

max-rocket-internet:
OK, it looks like a problem that https://github.com/terraform-aws-modules/terraform-aws-eks/pull/187 will solve.

Jeeppler:
I upgraded to `1.8.0`, but then it just silently fails: 

```
module.eks.null_resource.update_config_map_aws_auth: Provisioning with 'local-exec'...
module.eks.null_resource.update_config_map_aws_auth (local-exec): Executing: ["/bin/sh" "-c" "for i in {1..5}; do kubectl apply -f /k8s/config-map-aws-auth_cluster.yaml --kubeconfig /k8s/kubeconfig_cluster && break || sleep 10; done"]
module.eks.aws_iam_role_policy_attachment.workers_autoscaling: Creation complete after 1s (ID: cluster20181207234911216400000003-2018120723491235770000000a)
```

in version `1.7.0` it just throws an error: 

```
Releasing state lock. This may take a few moments...

Error: Error applying plan:

1 error(s) occurred:

[...]

PR #231, opened 2019-01-04
Aws auth not being applied to the cluster during the first terraform apply

stijndehaes:
# PR o'clock

## Description

I changed the name of the workers.tf to be 0_workers.tf file. Because of some dependency issues the aws auth configmap was not deployed the first time you executed terraform apply.
The following issue could point to the same behaviour: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/202

### Checklist


max-rocket-internet:
I don't think this makes any sense. AFAIK Terraform doesn't care what the file names are?

stijndehaes:
Well I have had some problems with the first time that the configmap wasn't applied. I will look into it further. But terraform loads the files alphabetically so if dependencies haven't correctly been calculated this can fix issues.

max-rocket-internet:
OK, let us know how it goes!

Also similar to this issue: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/199

marcelloromani:
FWIW I don't think renaming a tf file is the correct way to ensure a resource is created before others.
Proper dependencies via `depends_on` should be established.

@stijndehaes you say
> Well I have had some problems with the first time that the configmap wasn't applied
this is exactly the same issue I've been having for a few days.
I believe I found the culprit as per #238 

As a quick test I think you could try setting `interpreter` to `/bin/zsh`

stijndehaes:
Well had the same issue today. The configmap wasn't applied, but this time rerunning it didn't apply the configmap. Meaning that the file was generated correctly but not applied, maybe changing the interpreter could fix that. It's kind of hard to debug though as it doesn't happen always. 
Does anybody know why this resource is using a `local-exec` instead of the `terraform-provider-kubernetes` ?

marcelloromani:
@stijndehaes That is exactly the behaviour I noticed while trying to debug this issue.
The reason (as far as I could understand) is that TF doesn't check whether the config map has actually been applied to the cluster (i.e. there's no check of the k8s resource itself). Therefore if the `kubectl apply -f <aws_auth_config_map>` is not executed (as reported in quite a few issues), there's no way TF will ever re-run that `null_resource` on its own.

max-rocket-internet:
I just merged a PR to fix the shell interpreter issue. Also another PR to add another trigger to the configmap update.

Can you retest? As I don't think renaming a file is a normal solution 😅

stijndehaes:
Just had the same error happen with the latest 2.1.0. However it's very hard to debug. But indeed this renaming is not the solution. Will close this PR and try to find a consistent way to reproduce it first.

renperez:
any update on this? just ran into the same issue with the latest eks module. i had to apply the configmap manually in order for my workers to join the cluster

max-rocket-internet:
Please follow in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/420#issuecomment-507470797

Issue #234, opened 2019-01-07
ConfigMap may not get applied to recreated clusters


elatt:
# I have issues
We recently had to recreate our EKS cluster and were surprised to realize that the ConfigMap wasn't being applied.

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
Currently, a new cluster will be created but if a custom ConfigMap is defined, it will not get applied to the new cluster.

## If this is a bug, how to reproduce? Please include a code sample if relevant.
Create a cluster with a custom ConfigMap settings. Delete the cluster (either from the web UI or CLI) and re-run the terraform command from the same git repo (where the existing config-map is present in the workspace).

## What's the expected behavior?
I would expect a custom ConfigMap to always be applied to newly created clusters or when the existing ConfigMap changes.

## Environment details

* Affected module version: v2.0.0
* OS: MacOS
* Terraform version: v0.11.11

elatt:
@marcelloromani #238 looks like a separate issue. For my issues, the terraform command never failed, instead it failed to run the `kubectl` command at all because the `null_resource` was not fired.

elatt:
This is fixed in v2.1.0

PR #235, opened 2019-01-07
Fix issue where ConfigMap isn't applied to new cluster

elatt:
If you are trying to recover a cluster that was deleted, the current
code will not re-apply the ConfigMap because it is already rendered so
kubectl command won't get triggered.

This change adds the cluster endpoint (which should be different when
spinning up a new cluster even with the same name) so we will force a
re-render and cause the kubectl command to run.

# PR o'clock

## Description

This should be a simple way to work around the issue that we saw in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/234.

### Checklist

- [ ] ~~Any breaking changes are highlighted above~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
  - Removed invalid action from worker_autoscaling iam policy
+ - Updated the `update_config_map_aws_auth` resource to trigger when the EKS cluster endpoint changes. This likely
+   means that a new cluster was spun up so our ConfigMap won't exist (fixes #234) (by @elatt)
 
 ## [[v2.0.0](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v1.8.0...v2.0.0)] - 2018-12-14]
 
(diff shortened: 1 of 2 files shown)


elatt:
I'm not exactly sure how to write tests for this. The real goal would be to write a test that can detect that when the endpoint of an EKS cluster changes, the config-map will still get applied because this likely means that a new cluster has been created.

max-rocket-internet:
Hi @elatt !
But a change in endpoint means the whole EKS cluster would have to be destroyed and recreated, right? When else does an endpoint change? Am I missing something?

To be clear, you are solving these 2 issues?
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/231
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/199

elatt:
Ah, I also wasn't familiar with the `null_resource` and it seems the more terraform-y way would be to just add another trigger on the endpoint rather than having it be part of the ConfigMap template, I think?

max-rocket-internet:
> it seems the more terraform-y way would be to just add another trigger on the endpoint 

Correct!

Quite an edge case indeed because I think you would have had to delete your cluster outside of TF and then rerun TF to hit this?

Either way, all good! Fix your conflict and we can merge 🙂

elatt:
Yes, the cluster was deleted outside of TF and it was nice to have all our infrastructure checked into git but this bug took as a bit to understand why the new cluster wasn’t coming up correctly. The merge conflicts should be resolved.

PR #236, opened 2019-01-08
GH-233: initial commit adding support for ebs volumes in worker groups

### LINKED ISSUE #233: Feature request: ability to attach EBS volumes to worker nodes
# Feature request: ability to attach EBS volumes to worker nodes

I would like to be able to attach EBS volumes to worker nodes. Currently it is not possible to do so.

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request

It is not possible to attach EBS volumes to worker nodes.

## What's the desired behavior?

I would like to be able to include either an array of EBS volumes details for this module to attach or create-and-attach to the worker nodes.

I plan on submitting a PR.

maxenglander:
Resolves #233 

This PR adds support for between 0-2 EBS volumes to the EKS worker nodes by adding the following configuration options block:

```
worker_groups = [
  {
      ebs_device_block_0_device_name = "/dev/xvdb"
      ebs_block_device_0_volume_size           = "300"
      ebs_block_device_0_volume_type           = "gp2"
      ebs_block_device_0_delete_on_termination = true
      ebs_block_device_0_encrypted             = false

      ebs_device_block_1_device_name = "/dev/xvdb"
      ebs_block_device_1_volume_size           = "300"
      ebs_block_device_1_volume_type           = "gp2"
      ebs_block_device_1_delete_on_termination = true
      ebs_block_device_1_encrypted             = false
  }
]

worker_group_ebs_block_device_count = 2
```

The `ebs_block_device_n_*` arguments won't take effect unless the `worker_group_ebs_block_device_count` is greater than zero. I tried using a less awkward configuration syntax (e.g. an `ebs_block_device` array), but it appears Terraform won't  support indexing into nested arrays/maps [very well until 0.12](https://github.com/hashicorp/terraform/issues/2114).


max-rocket-internet:
hmmmmm, this is a considerable amount of added complexity/duplication.

Do you, or others, think it might just make more sense to just allow the LC to be defined outside this module and then set per worker group?

maxenglander:
@max-rocket-internet I agree that this is a lot of duplication. I like your idea as well, and don't feel strongly about what the right approach is. However, here's my argument for doing things this way:

Re: complicated syntax/duplication. I think that the "lot of duplication" may be an issue that resolves itself over time as more people move to using TF 0.12, and we can use [more elegant syntax](https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each) to attach EBS devices.

Re: supplying LC as a parameter. I think it's nice that this module currently is able to manage creating the LC and the security group that allows communication between worker nodes and the master, as well as setting the user data. Creating that security group and supplying appropriate user data are two steps that are likely to create pitfalls for new users of EKS. In my opinion, the fact that this module handles all that for you in a reliable way is one of the reasons for using this module in the first place. Asking users to supply their own LC means also asking them to attach appropriate security groups and user data, which increases the likelihood that they'll have a rocky start using EKS.

max-rocket-internet:
Sorry, I mean *optionally* supplying an LC per worker group.

i.e. current arrangement stays as is BUT someone who wants something more complicated, e.g. 2x EBS volumes, then this person can create their own LC outside of the module.

Know what I mean?

To do this, you could change [this line](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers.tf#L7) to:

```hcl
launch_configuration  = "${lookup(var.worker_groups[count.index], "launch_configuration", element(aws_launch_configuration.workers.*.id, count.index))}"
```

maxenglander:
@max-rocket-internet Sure thing I'm happy to change this PR to drop the current changes and use that approach instead

max-rocket-internet:
> I'm happy to change this PR to drop the current changes and use that approach instead

That would be awesome!

szpuni:
Hi guys, I see this PR is closed but I do not see any merges and proposed solution is not on mainstream:
`launch_configuration  = "${lookup(var.worker_groups[count.index], "launch_configuration", element(aws_launch_configuration.workers.*.id, count.index))}"`

Current master code [line](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/4369f0271b12cd3f98f705df036662f0660efa4b/workers.tf#L45) do not have ability to use custom launch configuration from outside module

max-rocket-internet:
> Current master code line do not have ability to use custom launch configuration from outside module

Correct. Feel free to make a PR.

PR #242, opened 2019-01-10
Update cluster.tf

sagivle:
adding enable condition

# PR o'clock

## Description

### Checklist

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -1,4 +1,5 @@
 resource "aws_eks_cluster" "this" {
+  count    = "${var.module_enabled ? 1 : 0}"
   name     = "${var.cluster_name}"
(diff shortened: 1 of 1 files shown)


cluster.tf
```diff
@@ -1,4 +1,5 @@
 resource "aws_eks_cluster" "this" {
+  count    = "${var.module_enabled ? 1 : 0}"
```
max-rocket-internet: This variable, `module_enabled`, is not defined though?

max-rocket-internet:
Also, even if we did add this variable, I think there would still be a load of TF errors if it was set to `false` as there would be many other resources created that depend on this resource? Do you know what I mean?

max-rocket-internet:
Closing due to no response.

PR #247, opened 2019-01-14
README: fix small typo in variable description

jonboulle:


skang0601:
You should update the descript in the variables.tf file and regenerate the docs rather than updating the generated docs themselves.

max-rocket-internet:
Thanks!

Interesting we are seeing a few changes in the way `terraform-docs` is presenting the values/defaults etc.

PR #249, opened 2019-01-18
Add use_max_pods

knqyf263:
# PR o'clock

## Description

Add use_max_pods to `templates/userdata.sh.tpl`
https://github.com/awslabs/amazon-eks-ami/blob/208c114737303cc350a5778fe56d489c9476f945/files/bootstrap.sh#L19

Since the default value of `--use-max-pods` is true, an limit is set based on the following values.
https://github.com/awslabs/amazon-eks-ami/blob/208c114737303cc350a5778fe56d489c9476f945/files/eni-max-pods.txt

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ##### Added
 
+- Add `use_max_pods` (by @knqyf263)
 - Write your awesome addition here (by @you)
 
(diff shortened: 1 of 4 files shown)


max-rocket-internet:
Hi @knqyf263!

Thanks for the PR.

What's the use case for this exactly? You want to reduce the max pods value for some reason?

knqyf263:
I close this because it is no longer necessary...

PR #252, opened 2019-01-25
Worker group tags

### LINKED ISSUE #89: Be able to define per ASGs tags
# I have issues
Tags currently are too prescriptive. I have a use case where I need to tag different ASGs with different tags. Im using the ability to push these tags down to node labels and taints to drive different workloads on my kubernetes cluster. Atm, it seems that I can only define tags once on the top level EKS module and these tags are used globally throughout. I would like to be able to define tags per ASGs. A sensible place to provide these seems to be the list of worker_groups maps.

## I'm submitting a...

* [ ] bug report
* [X] feature request
* [ ] support request

Tags are defined once in var.tags and used throughout both to tag the cluster resources itself, as well as, all ASGs that are created.

Should be able to provide tags in the list of worker_groups maps and these should be used to tagging the corresponding ASGs that are created for each respective worker group. If the tags are not set, they can default to the existing top-level global tag variable that is used.  

I can submit a PR if this is reasonable.

## Environment details

* Affected module version:
* OS:
* Terraform version:

stefansedich:
# PR o'clock

## Description

Adding ability to configure worker group ASG tags which resolves #89 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ##### Added
 
 - Ability to configure force_delete for the worker group ASG (by @stefansedich)
+- Ability to configure worker group ASG tags (by @stefansedich)
 
 ##### Changed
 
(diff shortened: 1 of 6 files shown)


stefansedich:
@max-rocket-internet looks like this may have been a fail, closing off until I can work something out, unless there are some TF hacks you know of many of the things I tried don't seem to work in 0.11 and may in 0.12, unless we had the users always defining the tags with empty arrays to match the worker group count which would not be ideal.

stefansedich:
@max-rocket-internet ok that involved ALOT of hair pulling but the best I could think of was that by default there is always one item in the list, then we use the matching item if we can find one otherwise we fallback to the first item in the list, this is not ideal but I tried a million other things and this was the only way I could get it working, any thoughts if this is an acceptable hack?

stefansedich:
@max-rocket-internet are you just running the examples? I can successfully run a `terraform plan` for the examples which are using defaults.

max-rocket-internet:
OK great! I think this looks pretty tidy 🙂

Can you fix conflict and update the README using `terraform-docs`?

stefansedich:
Thanks @max-rocket-internet should be good to go!

PR #260, opened 2019-01-31
Manage aws auth configmap without needing temp files

mbarrien:
# PR o'clock

## Description

This PR builds on top of #228 and removes the need for temp files altogether by using shell heredocs and file descriptors. This should be handled by most shells, although I've only tested in Bash.

Instead of writing temporary files, we use 2 heredocs to write to file descriptors 3 and 4, with each containing the contents of the config map and the kubectl respectively. We continue with the write_aws_auth_config variable from #228.

Unclear to me is whether this works in zsh, Windows, or pure sh. If it doesn't feel free to reject, but I feel this is MUCH cleaner of an implementation than all this local file management.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,13 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Ability to configure force_delete for the worker group ASG (by @stefansedich)
 - Added EBS optimized mapping for the g3s.xlarge instance type (by @stefansedich)
 - `enabled_metrics` input (by @zanitete)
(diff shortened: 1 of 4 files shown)


mbarrien:
Hmm... this may not work for kubectl versions above 1.10.x; 1.11+ reads the kubeconfig multiple times within a single invocation and since a heredoc is not readable multiple times, this fails.

max-rocket-internet:
> I feel this is MUCH cleaner of an implementation than all this local file management.

Agreed.

Sad it didn't work out though!

PR #272, opened 2019-02-07
Added output for generated kubeconfig filename

syst0m:
# PR o'clock

## Description

The terraform helm provider uses a kubectl config filename as a config parameter.
The eks module generates a kubectl config with a random filename.
In order to enable integration between the two and facilitate automated provisioning of an eks cluster, 
by passing the kubectl config filename as an input 
to the terraform helm provider, I added a new output to the eks module.

terraform helm provider: https://github.com/terraform-providers/terraform-provider-helm

### Checklist

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -39,6 +39,11 @@ output "kubeconfig" {
   value       = "${data.template_file.kubeconfig.rendered}"
 }
 
+output "kubeconfig_filename" {
+  description = "The filename of the generated kubectl config."
+  value       = "${local_file.kubeconfig.*.filename}"
+}
+
 output "workers_asg_arns" {
   description = "IDs of the autoscaling groups containing workers."
   value       = "${concat(aws_autoscaling_group.workers.*.arn, aws_autoscaling_group.workers_launch_template.*.arn)}"
--- a/outputs.tf
+++ b/outputs.tf
@@ -41,7 +41,7 @@ output "kubeconfig" {
 
 output "kubeconfig_filename" {
   description = "The filename of the generated kubectl config."
-  value       = "${local_file.kubeconfig.*.filename}"
+  value       = "${element(concat(local_file.kubeconfig.*.filename, list("")), 0)}"
 }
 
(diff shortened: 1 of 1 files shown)


okgolove:
@syst0m hello!
Does Helm provider wait for EKS cluster if you use this variable?
I mean are there errors like:
`Post http://localhost/apis/extensions/v1beta1/namespaces/kube-system/deployments: dial tcp [::1]:80: connect: connection refused`
?

syst0m:
@okgolove I cannot reproduce your error.
Can you elaborate a bit on your config?

okgolove:
I meant, I don't have kubectl config path output from EKS module (but you've added it within this PR) and my Helm provider isn't able to wait EKS module creates kubeconfig file. By default (without kubeconfig) Helm provider tries to access tiller via localhost.

loesak:
@okgolove i believe, unless things have changed in the last year, that what you are referring to is a general problem with Terraform itself in that the output of resources from one provider cannot be used as input to another provider. 

Where this change helps, at least for me, when you want to reference the output of Terraform state as the input to a different set of Terraform configuration.

outputs.tf
```diff
@@ -39,6 +39,11 @@ output "kubeconfig" {
   value       = "${data.template_file.kubeconfig.rendered}"
 }
 
+output "kubeconfig_filename" {
+  description = "The filename of the generated kubectl config."
+  value       = "${local_file.kubeconfig.*.filename}"
```
antonbabenko: If `kubeconfig` is disabled (`var.write_kubeconfig = 0`) this value won't work. Change this to:
`value = ${element(concat(local_file.kubeconfig.*.filename, list("")), 0)}`

syst0m: Thanks for the suggestion, @antonbabenko .
I updated the PR.

antonbabenko:
[v2.2.1](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v2.2.1) has been released.

PR #284, opened 2019-02-25
Add optional iam_path

tekn0ir:
# PR o'clock

## Description

In some corporate environments you may only be allowed to create iam roles and policies within a certain path. Added an optional iam_path variable to allow support this type of configuration.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | workers\_group\_launch\_template\_defaults | Override default values for target groups. See workers_group_defaults_defaults in locals.tf for valid keys. | map | `{}` | no |
 | write\_aws\_auth\_config | Whether to write the aws-auth configmap file. | string | `"true"` | no |
 | write\_kubeconfig | Whether to write a Kubectl config file containing the cluster configuration. Saved to `config_output_path`. | string | `"true"` | no |
+| iam\_path | If provided, all IAM roles will be created with path. | string | `"/"` | no |
 
 ## Outputs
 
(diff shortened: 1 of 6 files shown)


variables.tf
```diff
@@ -221,3 +221,8 @@ variable "permissions_boundary" {
   description = "If provided, all IAM roles will be created with this permissions boundary attached."
   default     = ""
 }
+
+variable "iam_path" {
+  description = "If provided, all IAM roles will be created on this path."
+  default     = ""
```
max-rocket-internet: Does this definitely work as `""`? It shouldn't be `/`?

README.md
```diff
@@ -148,6 +148,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | workers\_group\_launch\_template\_defaults | Override default values for target groups. See workers_group_defaults_defaults in locals.tf for valid keys. | map | `{}` | no |
 | write\_aws\_auth\_config | Whether to write the aws-auth configmap file. | string | `"true"` | no |
 | write\_kubeconfig | Whether to write a Kubectl config file containing the cluster configuration. Saved to `config_output_path`. | string | `"true"` | no |
+| iam_path | If provided, all IAM roles will be created with path. | string | `""` | no |
```
max-rocket-internet: I think this is missing `\` for the underscore. See https://github.com/terraform-aws-modules/terraform-aws-eks#doc-generation

antonbabenko: This is because the version of `terraform-docs` is not new enough. Update `terraform-docs` and rerun `pre-commit run -a` to get this fixed.

tekn0ir: I updated the README so it follows the convention of escaped underscores.

variables.tf
brandoconnor: Looks like this was fixed in a later commit. Resolving.

PR #292, opened 2019-03-02
Adding workers_launch_template ebs encryption

russki:
# PR o'clock

## Description

Adding EBS encryption option for EKS workers in workers_launch_template

### Checklist

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -62,6 +62,8 @@ locals {
     root_volume_size                         = "100"                                         # root volume size of workers instances.
     root_volume_type                         = "gp2"                                         # root volume type of workers instances, can be 'standard', 'gp2', or 'io1'
     root_iops                                = "0"                                           # The amount of provisioned IOPS. This must be set with a volume_type of "io1".
+    root_encrypted                           = false                                         # root volume encryption for workers.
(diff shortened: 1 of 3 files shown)


local.tf
```diff
@@ -62,6 +62,8 @@ locals {
     root_volume_size                         = "100"                                         # root volume size of workers instances.
     root_volume_type                         = "gp2"                                         # root volume type of workers instances, can be 'standard', 'gp2', or 'io1'
     root_iops                                = "0"                                           # The amount of provisioned IOPS. This must be set with a volume_type of "io1".
+    root_encrypted                           = false                                         # root volume encryption for workers.
+    kms_key_id                               = ""                                            # AWS KMS customer master key to use when creating encrypted volume. ASG must have access to CMK.
```
max-rocket-internet: Can you fix the comment here? "customer" is not relevant and what is CMK?

I think just `KMS key ID used for encrypted block device` would be good 🙂

russki: @max-rocket-internet updated the comment! Let me know if this looks better. I wanted to make a distinction between the default KMS `aws/ebs` key and the `Customer Master Keys (CMK)` that one could want to use

PR #316, opened 2019-03-22
Cni custom networking

frimik:
# PR o'clock

## Description

Favour some simplicity in the kubectl manifest applications. Terraform properly fails its run now unless kubectl explicitly succeeds.
Favour an `asset_dir` parameter over the `config_output_path` which has been explicitly removed. The asset dir will include manifests and the kubeconfig file.

Previously, these were only temporarily written to disk - in the module path. Now it will be properly output in the run, it will be the operators job to decide what to do with the files.

The `write_aws_auth_config` and `write_kubeconfig` parameters have been removed.

### Checklist

- [ ] Test results are pasted in this PR (in lieu of CI)  - Test kitchen fails due to it attempts terraform destroy on non-existing infra. https://github.com/newcontext-oss/kitchen-terraform/issues/271

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,13 +110,14 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
-| asset\_dir | Absolute path to base directory of bootstrap manifests and secrets such as kubeconfig. [Example: `${path.root}/assets`] | string | n/a | yes |
+| asset\_dir | Absolute path to base directory of bootstrap manifests and secrets such as kubeconfig. [Example: `${path.cwd}/assets`] | string | `""` | no |
 | cluster\_create\_security\_group | Whether to create a security group for the cluster or attach the cluster to `cluster_security_group_id`. | string | `"true"` | no |
 | cluster\_create\_timeout | Timeout value when creating the EKS cluster. | string | `"15m"` | no |
 | cluster\_delete\_timeout | Timeout value when deleting the EKS cluster. | string | `"15m"` | no |
 | cluster\_name | Name of the EKS cluster. Also used as a prefix in names of related resources. | string | n/a | yes |
 | cluster\_security\_group\_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingres/egress to work with the workers and provide API access to your current IP/32. | string | `""` | no |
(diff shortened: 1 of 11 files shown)


max-rocket-internet:
I see your idea @frimik, and it looks cool, but I don't think this is something we will merge. It's very complex and managing CNI related resources, or really any resources inside k8s, is not something we want to do in this module: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/99

It looks like you can implement this outside the module quite easily though, right? There's no outputs missing that would stop you doing that?

frimik:
@max-rocket-internet I understand your position exactly. Because I'm of the same idea. Spawn cluster, then have next stage which applies "components".

In this case though, it's so much tied to the cluster inception itself (CIDR, sec groups, subnets). As well as not being NEW manifests, but rather patching the existing one (`aws-node`).

Plus there's already `kubectl apply` in places. Though, I didn't look at them to compare them conceptually to what I was doing here.

I did create two implementations of this, one which is a "wrapper" module, and then this because wrapper modules in Terraform is simply [awful to manage](https://github.com/hashicorp/terraform/issues/15818) today.

Module "passthrough" could have been made bareable if tf supported [undef](https://github.com/hashicorp/terraform/issues/5471).

Enough ranting about that though :smile:, like any of the other parts in this pull req?

frimik:
Obviously we're now also impatiently waiting for the official release of the [vpc cni](https://github.com/aws/amazon-vpc-cni-k8s) to actually include the new label code according to the *Custom CNI Networking* documentation ...

max-rocket-internet:
> I did create two implementations of this, one which is a "wrapper" module

I think this is the best solution. We do this for many TF modules. For example, we wrap public RDS and Aurora modules in order to get cloudwatch alarms and route53 records all done in one module.

morganchristiansson:
This PR would be very useful to automatically create ENIConfig resources in kubernetes for each subnet used.

Not sure how it could be done as a wrapper. Can be done as a secondary module perhaps...

This is to support CNI Custom networking per Amazon EKS doco https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html

morganchristiansson:
Actually I have discovered that ENIConfig resources are not needed if setting AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG=false

Pods are simply created with same subnet and sg as worker eni. Just start workers in different subnets and use node-label and nodeSelector to schedule pods on them.

Issue #310, opened 2019-03-21
Worker nodes are not joining the cluster using version 2.3.0


armandorvila:
# I have issues
While creating a new cluster using the version 2.3.0 of the module and the AMI `ami-01e08d22b9439c15a` for the worker nodes, not a single worker node joins the cluster. 
## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behaviour?

* The EKS cluster is created, the autoscaling group, the lunch configuration and nodes are created. However, when running kubectl get nodes there are no nodes at all. 

* Running cat /var/lib/kubelet/kubeconfig within any of the worker nodes, we can see a miss-configuration issue: 

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority: /etc/kubernetes/pki/ca.crt
    server: https://<Skipping this part>.sk1.eu-west-1.eks.amazonaws.com
  name: kubernetes
contexts:
- context:
    cluster: kubernetes
    user: kubelet
  name: kubelet
current-context: kubelet
kind: Config
preferences: {}
users:
- name: kubelet
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - token
      - -i
      - --enable-docker-bridge
      command: /usr/bin/aws-iam-authenticator
      env: null
```

Instead of the cluster name, after the `-i` option we are getting `--enable-docker-bridge`

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Create a new cluster using the 2.3.0 version of the module and the AMI `ami-01e08d22b9439c15a` and run:
```
kubectl get nodes
```
SSH into one of the worker nodes and run:
```
cat /var/lib/kubelet/kubeconfig
```
## What's the expected behaviour?

Worker nodes joining the cluster properly, just like they do with the version 2.2.1 of the module.

## Environment details

* Affected module version: 2.3.0
* OS: Mac OS and Linux Alpine
* Terraform version: 0.11.13.

max-rocket-internet:
That's strange. I think it's related to https://github.com/terraform-aws-modules/terraform-aws-eks/pull/302

What is your AMI release? Perhaps in your AMI version the script `/etc/eks/bootstrap.sh` doesn't have support for `--enable-docker-bridge`?

armandorvila:
[...]

Issue #317, opened 2019-03-22
Replace the `enable-docker-bridge` parameter with a generic one


RothAndrew:
# I have issues

per @max-rocket-internet's comment [here](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/310#issuecomment-475629728) in #310 :

> Maybe we just replace the enable-docker-bridge parameter with just a general parameter like bootstrap_extra_args that defaults to "". This is safer for AMI changes, will solve this issue and means people can change the docker bridge setting also.

Note: This would be a breaking change which would mean a major version bump.

## I'm submitting a...

* [ ] bug report
* [X] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

Older AMI's don't work since they don't have the new `--enable-docker-bridge` flag, but this module adds it with a `true` or `false`. Changing it to a generic parameter like `bootstrap_extra_args` will eliminate that issue and add future proofing for other similar stuff that may come up in the future.

## Are you able to fix this problem and submit a PR? Link here if you have already.

I am capable of doing a PR. I don't have the time right now, but if this sits untouched for long enough I will eventually come along to do it.

Fixes #310

max-rocket-internet:
Maybe @michaelmccord can make the PR 😃

michaelmccord:
@max-rocket-internet -- I'll try and make some time to get around to this.

max-rocket-internet:
I made a PR: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/320

@RothAndrew could you test it perhaps??

RothAndrew:
I will give it a shot :)

RothAndrew:
Testing this now, about 20 days late, sorry :)

max-rocket-internet:
This was resolved with the PR above.

PR #320, opened 2019-03-26
Replacing enable_docker_bridge with a generic option called bootstrap_extra_args

max-rocket-internet:
## Description

Replacing enable_docker_bridge with a generic option called bootstrap_extra_args

Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/310 https://github.com/terraform-aws-modules/terraform-aws-eks/issues/317

### Checklist

### DIFF
--- a/templates/userdata.sh.tpl
+++ b/templates/userdata.sh.tpl
@@ -4,7 +4,7 @@
 ${pre_userdata}
 
 # Bootstrap and join the cluster
-/etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' --enable-docker-bridge '${enable_docker_bridge}' --kubelet-extra-args '${kubelet_extra_args}' '${cluster_name}'
+/etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' ${bootstrap_extra_args} --kubelet-extra-args '${kubelet_extra_args}' '${cluster_name}'
 
 # Allow user supplied userdata code
 ${additional_userdata}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- Replaced `enable_docker_bridge` with a generic option called `bootstrap_extra_args` to resolve [310](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/310) (by @max-rocket-internet)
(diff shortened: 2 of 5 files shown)


templates/userdata.sh.tpl
```diff
@@ -4,7 +4,7 @@
 ${pre_userdata}
 
 # Bootstrap and join the cluster
-/etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' --enable-docker-bridge '${enable_docker_bridge}' --kubelet-extra-args '${kubelet_extra_args}' '${cluster_name}'
+/etc/eks/bootstrap.sh --b64-cluster-ca '${cluster_auth_base64}' --apiserver-endpoint '${endpoint}' ${bootstrap_extra_args} --kubelet-extra-args '${kubelet_extra_args}' '${cluster_name}'
```
max-rocket-internet: Not sure about the no quoting here.

max-rocket-internet:
@RothAndrew OK good idea. I've highlighted it in the changelog in the same way it's been done previously

RothAndrew:
I was referring more to a section that says something like: "Use the `bootstrap_extra_args` variable to add other arguments such as the new `--enable-docker-bridge` flag, which is necessary to for functionality like docker-in-docker"

max-rocket-internet:
OK but where do we put that though? Not in the main readme as that is documentation for the module itself, not for random gotchas and issues.

We could create another file in `/docs`? Or perhaps we create a file `/docs/known-issues.md` where we could detail this and also all the subnet tagging issues people seem to have?

max-rocket-internet:
@RothAndrew I will merge this as I've tested it both with and without the docker bridge option and the nodes join the cluster.

If you feel there is some documentation about that issue that is missing then you can make a PR for that 🙂

RothAndrew:
@max-rocket-internet sorry, this got pushed off my radar.

There could definitely be a troubleshooting or FAQ doc, that's a good idea.

On a more troubling topic, this should have resulted in a major version bump, since it has a breaking change. Recommend immediately releasing a v2.3.2 tag with this change reverted and then releasing v3.0.0.

Without doing that, someone who had the version they are using as `~> 2.3.0`, and using `enable_docker_bridge` will get hosed when their `terraform init` installs v2.3.1.

max-rocket-internet:
> this should have resulted in a major version bump

I think you are right but it's a little late now. I don't want to delete a tag as that is just messy.

> Without doing that, someone who had the version they are using as ~> 2.3.0, and using enable_docker_bridge will get hosed

Not exactly. The docker bridge option will stop working ONLY if they update their version of this module without checking what's changed AND if they don't look at what terraform is actually changing.

PR #329, opened 2019-03-29
adding IAM instance profiles to outputs, addresses #323

### LINKED ISSUE #323: Include worker_iam_instance_profile_arn as an output
Currently there are outputs for `worker_iam_role_name` and `worker_iam_role_arn` but I need at least `worker_iam_instance_profile_arn` as well. While you can [look up an instance profile for a role](https://docs.aws.amazon.com/cli/latest/reference/iam/list-instance-profiles-for-role.html) using `aws` CLI commands, Terraform doesn't provide a path to do this directly using the existing [data sources](https://www.terraform.io/docs/providers/aws/d/iam_instance_profile.html). Thus we really need this as an output variable in this module.

soapergem:
# PR o'clock

## Description

Adds two outputs for IAM instance profile (role was already present, but not instance profile). See #323 

### Checklist

- [N/A] Tests for the changes have been added and passing (for bug fixes/features)
- [N/A] Test results are pasted in this PR (in lieu of CI)
- [N/A] Any breaking changes are highlighted above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Write your awesome addition here (by @you)
+- Added outputs for worker IAM instance profile(s) (by @soapergem)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


soapergem:
@max-rocket-internet good catch. I've updated the README.

soapergem:
@max-rocket-internet just checking in... can we get this merged please?

PR #339, opened 2019-04-10
Option to install kubectl and aws-iam-authenticator in module path

vchinnakotla:
# PR o'clock

Added a option to install kubectl and aws-iam-authenticator, this will help users running their infrastructure from CI/CD or Terraform enterprise. Helps us to not bother about invoking installation from outside the module.
Variable **install_kubectl** defaults to false. If true installs both of them in module path and uses them to apply configMap

### Checklist

### DIFF
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -17,7 +17,7 @@ resource "null_resource" "update_config_map_aws_auth" {
 for i in `seq 1 10`; do \
 echo "${null_resource.update_config_map_aws_auth.triggers.kube_config_map_rendered}" > kube_config.yaml & \
 echo "${null_resource.update_config_map_aws_auth.triggers.config_map_rendered}" > aws_auth_configmap.yaml & \
-kubectl apply -f aws_auth_configmap.yaml --kubeconfig kube_config.yaml && break || \
+${local.kubectl_command} apply -f aws_auth_configmap.yaml --kubeconfig kube_config.yaml && break || \
 sleep 10; \
 done; \
 rm aws_auth_configmap.yaml kube_config.yaml;
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -5,7 +5,10 @@ resource "local_file" "config_map_aws_auth" {
 }
 
 resource "null_resource" "update_config_map_aws_auth" {
-  depends_on = ["aws_eks_cluster.this"]
+  depends_on = [
+    "aws_eks_cluster.this",
+    "null_resource.install_kubectl",
+  ]
 
   provisioner "local-exec" {
     working_dir = "${path.module}"
--- a/data.tf
+++ b/data.tf
@@ -50,7 +50,7 @@ data "template_file" "kubeconfig" {
     endpoint                          = "${aws_eks_cluster.this.endpoint}"
     region                            = "${data.aws_region.current.name}"
     cluster_auth_base64               = "${aws_eks_cluster.this.certificate_authority.0.data}"
-    aws_authenticator_command         = "${var.install_kubectl ? "${path.module}/${var.kubeconfig_aws_authenticator_command}" : "${var.kubeconfig_aws_authenticator_command}"}"
+    aws_authenticator_command         = "${local.kubeconfig_aws_authenticator_command}"
     aws_authenticator_command_args    = "${length(var.kubeconfig_aws_authenticator_command_args) > 0 ? "        - ${join("\n        - ", var.kubeconfig_aws_authenticator_command_args)}" : "        - ${join("\n        - ", formatlist("\"%s\"", list("token", "-i", aws_eks_cluster.this.name)))}"}"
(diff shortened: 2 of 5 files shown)


max-rocket-internet:
Hey @vchinnakotla 
Thanks for the PR but we are not interested in managing software on people's computers. Also this will fail if `curl` is not installed or if it is run on Windows.

joseph-wortmann:
@max-rocket-internet I agree with both of your comments. Please consider the following though:

1. This module is almost impossible to run correctly from Terraform Enterprise because TFE does not install kubectl (or aws-cli, etc.) on the image running TF. Their best practice is to null_resource install the necessary software, but there is no good hook from outside the module to trigger the install.
2. This only installs when you tell it to, and then it only installs it in the module-specific path. It's not really managing the software on the computer, only in the module.
3. If you don't have curl, or you are running on windows, and you flip this flag then you are on your own. I don't see how that is any different than the current module requirements for understanding your underlying system.
4. If you run the current module on a Windows machine the null_resource "update_config_map_aws_auth" will probably fail anyhow. It's pretty sh/bash centric...

max-rocket-internet:
Hey @joseph-wortmann 
OK, I get you! I've never used Terraform Enterprise so don't really understand the use case.

Is getting binaries from the internet really the common way to do this in Terraform Enterprise modules?

Is it not possible to just wrap this module in another module with the extra bits?

Why isn't it possible to just do this outside the module? Is that a Terraform Enterprise limitation?

joseph-wortmann:
Hi @max-rocket-internet!

To answer your questions (hopefully):

1. Getting binaries from the internet (using a null_resource) is really the only way to get them into a TFE run. TFE spins up a fresh container per TF run.
2. You could wrap this in another module to install the extra bits. But - the wrapper would have to install it per module on every run, which is kind of brute force. Currently doing it in module only does the installation when it is flagged AND the triggers kick. So really the install would not happen on most runs.
3. Doing it completely outside the module is possible, but once again this is brute force, because you have to install it on every run as there are no reliable triggers to know exactly when the module will try to use the required extra bits.
4. The brute force method is a little (a lot?) annoying, in that every run planned will show changes to be applied, even if there really are no changes as the only "change" is the installation of the extra bits. This forces the inspection of every plan run, and I think may lead to people missing actual changes that are in the plan (alert overload, so to speak).

The net-net - TFE is fantastic and awesome, especially for larger teams or many TF top-level runs, but it does have an unfortunate limitation when it comes to system executables that the TF depends on. The pattern that we've come up with for our modules is to encapsulate that dependency within the module by doing a module-local install if flagged. This seems to work very well and abstracts the complexity away from our module users, while allowing advanced users or non-TFE users to operate as they want and do presently.

I hope all of that made sense.

RothAndrew:
I'm a 👎 on this one, I don't have any experience with Terraform Enterprise either, but my initial reaction when I see the word "container" is, If it is running in a container why don't you install the tools you need when building the container? Does Terraform Enterprise not let you use a container you specify?

joseph-wortmann:
@RothAndrew 

No, TFE does not let you specify the container. Obviously, if it did, that would be an easy fix.

max-rocket-internet:
> Doing it completely outside the module is possible, but once again this is brute force

I don't really understand what difference it makes where the curl requests are? Whether they are in this module or in a module-in-a-module.

I'm not super keen on this PR for all the obvious reasons. Also this is the first we've heard of solving these TFE shortcomings in this module. But let's leave this open for while, perhaps there's other TFE users out there and they can add their opinion.

joseph-wortmann:
@max-rocket-internet 

It has to do with the triggers for the null_resource that calls curl and does the install. Outside of the module there are really no good triggers to perform the install, leaving only the option of running the install on every TF run.

This will produce a planned change on every run, even when there is no actual change to the underlying infrastructure.

Thanks for leaving it open

max-rocket-internet:
> Outside of the module there are really no good triggers to perform the install

But the [3 triggers](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/aws_auth.tf#L27-L29) for the `null_resource` are all outputs of this module? Can't you just use them in the same way that the `null_resource` in the module uses uses them?

joseph-wortmann:
@max-rocket-internet 

You can't use the outputs of a module to trigger a `null_resource` to install something that must be in place before the module runs.

stijndehaes:
This PR https://github.com/terraform-aws-modules/terraform-aws-eks/pull/355 can fix your problem since kubectl or aws-iam-authenticator are not needed after this is merged

joseph-wortmann:
@stijndehaes That is a _very_ elegant solution. Once that is merged you're right, this PR is no longer needed.

max-rocket-internet:
@joseph-wortmann 

> You can't use the outputs of a module to trigger a null_resource to install something that must be in place before the module runs.

Sorry, maybe I'm not being clear. You don't need `kubectl` or `aws-iam-authenticator` in the module as you can set `manage_aws_auth` to `false`. Use the outputs of `kubeconfig` and `config_map_aws_auth` to manage the whole thing outside the module.

Does that make sense?

max-rocket-internet:
I'm gonna close this as the module has now moved to TF 0.12 Syntax and it's been a while. Feel free to reopen 🙂

PR #343, opened 2019-04-10
Fix launch templates

vchinnakotla:
# PR o'clock

## Description
Removed block-device mappings in the launch templates.
 You cannot use block-device mappings for root

From TF docs - “Configure additional volumes of the instance besides specified by the AMI.”

Note the “besides”.

Trying to configure the AMI’s root device in this way with a launch template results in some pretty weird behavior. Mostly bad weird.

### Checklist


kubectl.tf
```diff
@@ -1,3 +1,24 @@
+resource "null_resource" "install_kubectl" {
+  provisioner "local-exec" {
+    working_dir = "${path.module}"
+
+    command = <<EOH
+curl -LO https://storage.googleapis.com/kubernetes-release/release/v${local.kubectl_versions[var.cluster_version]}/bin/linux/amd64/kubectl && \
```
max-rocket-internet: I don't know what this is but it's nothing related to the title of this PR or the description. Please fix this to keep the description and the changes concise 🙂

vchinnakotla: the commits unrelated to this PR are removed. Please review.

vchinnakotla:
@max-rocket-internet resolved the conflict, added notes to changelog

local.tf
```diff
@@ -44,46 +44,39 @@ locals {
...
-    protect_from_scale_in                    = false                                         # Prevent AWS from scaling in, so that cluster-autoscaler is solely responsible.
-    iam_role_id                              = "${local.default_iam_role_id}"                # Use the specified IAM role if set.
-    suspended_processes                      = ""                                            # A comma delimited string of processes to to suspend. i.e. AZRebalance,HealthCheck,ReplaceUnhealthy
-    target_group_arns                        = ""                                            # A comma delimited list of ALB target group ARNs to be associated to the ASG
-    enabled_metrics                          = ""                                            # A comma delimited list of metrics to be collected i.e. GroupMinSize,GroupMaxSize,GroupDesiredCapacity
-    placement_group                          = ""                                            # The name of the placement group into which to launch the instances, if any.
```
max-rocket-internet: `placement_group` is in use here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers_launch_template.tf#L87

local.tf
```diff
@@ -44,46 +44,39 @@ locals {
...
-    pre_userdata                             = ""                                            # userdata to pre-append to the default userdata.
-    bootstrap_extra_args                     = ""                                            # Extra arguments passed to the bootstrap.sh script from the EKS AMI.
-    additional_userdata                      = ""                                            # userdata to append to the default userdata.
-    ebs_optimized                            = true                                          # sets whether to use ebs optimization on supported types.
-    enable_monitoring                        = true                                          # Enables/disables detailed monitoring.
-    public_ip                                = false                                         # Associate a public ip address with a worker
-    eni_delete                               = true                                          # Delete the ENI on termination (if set to false you will have to manually delete before destroying)
```
max-rocket-internet: `eni_delete` is in use here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers_launch_template.tf#L67

local.tf
```diff
@@ -44,46 +44,39 @@ locals {
...
+    spot_instance_pools                      = 10                              # "Number of Spot pools per availability zone to allocate capacity. EC2 Auto Scaling selects the cheapest Spot pools and evenly allocates Spot capacity across the number of Spot pools that you specify."
+    spot_max_price                           = ""                              # Maximum price per unit hour that the user is willing to pay for the Spot instances. Default is the on-demand price
+    spot_price                               = ""                              # Cost of spot instance.
+    placement_tenancy                        = "default"                       # The tenancy of the instance. Valid values are "default" or "dedicated".
+    root_volume_size                         = "100"                           # root volume size of workers instances. Only used for cluster-autoscaler tags on ASGs. You cannot change the root volume size with a launch template.
```
max-rocket-internet: > You cannot change the root volume size with a launch template.

How can this be? You can change this for any AMI using an LC. Are you saying this option is simply impossible with LTs?

max-rocket-internet:
@vchinnakotla sorry for the delay, I was on holiday.

> Trying to configure the AMI’s root device in this way with a launch template results in some pretty weird behavior. Mostly bad weird.

What do you mean? How was changing root volume size possible with LC but not with LT?

russki:
Per the https://www.terraform.io/docs/providers/aws/r/launch_template.html example, they're configuring `/dev/sda1` in `block_device_mappings`, which is a root partition

What weird behavior are you seeing? Ability to modify ebs root size and automatically encrypt root partition of EKS workers is a very nice feature to have

joseph-wortmann:
@max-rocket-internet @russki 

According to the AWS and TF docs it appears that you cannot alter the root volume with an LT.

From the AWS LT docs:
> For Storage (Volumes), specify volumes to attach to the instance **besides** the volumes specified by the AMI.

From the TF LT docs:
> Configure additional volumes of the instance **besides** specified by the AMI.

Emphasis added by me.

My testing on my clusters confirmed this to a fair degree of certainty.

When I was running with this altering the root volume, the behavior from AWS was inconsistent, but predominantly I would get an error saying that it couldn't modify the AMI's volume. Digging into this error is when I realized you can't muck with the root volume from a LT. Sorry that I have lost the underlying errors and research.

`/dev/sda1` is not a root device on many HVM AMIs (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html#available-ec2-device-names). It is for sure not on the `amazon-eks-node-*` AMIs, which uses `/dev/xvda`. In any case, I don't know that I would ascribe too much accuracy to TF examples for applicability to all use cases.

AFAIK, the only way to alter the root volume on instance launch (if you need that) is to use an LC.

max-rocket-internet:
Interesting points, @vchinnakotla!

TF docs have just copied the AWS docs, so I think we can ignore them and focus on the AWS docs.

I think that perhaps one of the following is the case:

1. The AWS docs are wrong
2. The AWS docs are misleading due to poor wording
3. The AWS docs are correct and you cannot resize the root volume

I think it's very unlikely number 3 is true. The EKS AMI would have a root volume of 20 GB in this case. That's unusable. Also I would also say the LTs are unusable if this is the case. Also, I did a test with this worker group configuration:

```hcl
worker_group_launch_template_count = 3

  {
    name                = "test-1-"
    instance_type       = "c4.large"
    root_volume_size    = 120
  },
  {
    name                = "test-2-"
    instance_type       = "c4.large"
    root_volume_size    = 140
  },
  {
    name                = "test-3-"
    instance_type       = "m5.large"
    root_volume_size    = 160
  },
]
```

And it works fine:

```
$ df -h /
/dev/xvda1      140G  3.1G  137G   3% /

$ df -h /
/dev/nvme0n1p1  160G  3.1G  157G   2% /

$ df -h /
/dev/nvme0n1p1  160G  3.1G  157G   2% /
```

Either way, in my testing it works for the EKS AMI, which is what we are using by default in this module.

Thoughts?

russki:
In addition to the above points, LC doesn’t allow you to use ebs/kms encryption

joseph-wortmann:
@max-rocket-internet 

Yeah, I had it working for pretty trivial cases too. Until I started crossing regions and encrypting the root drives. Then behavior became _interesting_, it did not perform the same way in us-east-1 and us-east-2, throwing errors in the ASG on attempted instance creation. I was unable to create a configuration that worked in both regions, btw, and I tried every permutation. Nothing like hours of staring at instances coming up (or not, as the case may be).

In any case, perhaps we can simply compromise and not make the root drive configuration required in the LT? Frankly, this is probably the best solution, since if my AMI is already correct why would I want to mess with it when launching?

max-rocket-internet:
I still can't understand what issues you're talking about 😃

> Until I started crossing regions

What does this mean, exactly?

> Then behavior became interesting

What does this mean, exactly?

> it did not perform the same way in us-east-1 and us-east-2

What does this mean, exactly?

> errors in the ASG on attempted instance creation

Exactly what error?

> perhaps we can simply compromise and not make the root drive configuration required in the LT

But most people are using the EKS AMI provided by AWS. This would make everyone's root volume 20GB. So not an option.

As shown above, setting the root volume size works. I tested it in `eu-west-1` region. If you are implying this won't work in other regions, then can you test my config above with the normal EKS AMI and see if you can reproduce your issues?

max-rocket-internet:
I'm gonna close this as the module has now moved to TF 0.12 Syntax and it's been a while. Feel free to reopen 🙂

PR #355, opened 2019-04-22
Use kubernetes provider to manage aws auth

stijndehaes:
# PR o'clock

## Description

This changes the aws auth configmap to be managed by the kubernetes provider. This is a breaking change as already created aws auth configmaps will have to be moved into the terraform state.
### Checklist


max-rocket-internet:
Interesting!

Previously this was avoided because there was no `depends_on` [for providers](https://github.com/hashicorp/terraform/issues/2430).

How does this work now? How does it work with 5 separate clusters defined?

stijndehaes:
I actually never tested that. We always only have one cluster per terraform state. I will try to test this somewhere this week. I guess you should be able to do this using provider aliases: https://www.terraform.io/docs/configuration/providers.html#alias-multiple-provider-instances

But then I should add the possibility to pass one through a variable.

stijndehaes:
I have done some testing and it is very easy to do it with more then one cluster. For an example see this comment:
https://github.com/hashicorp/terraform/issues/18682#issuecomment-438342610

I also always thought it was not possible to have depends_on for provider. The depends_on field is still not supported. But you can pass variables in the provider depending on other resources. I got the example from here: https://www.terraform.io/docs/providers/aws/d/eks_cluster_auth.html

I added an example in the eks test fixtures. Is this enough or should we document this somewhere?

simonvanderveldt:
> Previously this was avoided because there was no `depends_on` [for providers](https://github.com/hashicorp/terraform/issues/2430).
> 
> How does this work now? How does it work with 5 separate clusters defined?

@max-rocket-internet That's correct. We've worked around this by implementing the aws-auth part in a separate module, the module then ensures the dependency (in this case the Kubernetes/EKS cluster needs to be created/available) is met by taking for example the cluster ID as a variable.

max-rocket-internet:
> That's correct. We've worked around this by implementing the aws-auth part in a separate module

So we would have to move the aws-auth part out of this module for it to work correctly then?

stijndehaes:
> > That's correct. We've worked around this by implementing the aws-auth part in a separate module
> 
> So we would have to move the aws-auth part out of this module for it to work correctly then?


You don't have to move the aws-auth part outside of this module. Terraform is smart enough to notice the provider depends on a data object, that depends on a resource and waits nicely until the resource is created. I have tested this branch with a setup I am using as a client.

simonvanderveldt:
> So we would have to move the aws-auth part out of this module for it to work correctly then?

No, you can use a nested module for that :)

> You don't have to move the aws-auth part outside of this module. Terraform is smart enough to notice the provider depends on a data object, that depends on a resource and waits nicely until the resource is created. I have tested this branch with a setup I am using as a client.

AFAIK the issue with a provider depending on resource creation that's done by another provider hasn't been fixed. See the previously mentioned https://github.com/hashicorp/terraform/issues/2430
Does it work if there are no resources and you just run a `terraform plan`?

stijndehaes:
@simonvanderveldt A terraform plan works. You can easily try it by checking out my branch and doing a terraform plan of eks_test_fixture in the examples folder. Just to be sure I double checked and ran terraform plan locally and it works.

stijndehaes:
@max-rocket-internet No worries I tested a terraform apply from scratch as well. I just hadn't tested only the terraform plan so I double checked it. But as I said earlier I have tested the branch with a brand new terraform apply. Not of the example but of another setup where I used this module in.

If you want we could have a google hangouts chat about this PR to make communication easier?

stijndehaes:
To be 100% sure I also just tested the `examples/eks_test_fixture`. This one also works correctly the configmap is nicely created after the cluster has been created

outputs.tf
```diff
@@ -28,11 +28,6 @@ output "cluster_security_group_id" {
   value       = "${local.cluster_security_group_id}"
 }
 
-output "config_map_aws_auth" {
```
max-rocket-internet: Can we just keep this as an output? I'm sure someone will still want it. e.g. so they can have the file in their version control or something.

max-rocket-internet:
Sorry for the delay! 

OK let's get this merged but I'm still a little confused here. Do we have any issues at all with this PR? Here's my summary, correct me if I'm wrong:

1. `depends_on` for providers issue doesn't matter because it just works
2. Multiple clusters in the same code base works also with this PR?
3. Changes to the configmap are applied to the cluster correctly

Apart from rebase, what do we need to do to get this merged? What else needs to be tested?

stijndehaes:
Just tested applying when the aws-auth configmap already exists.
This does not work:

```
* kubernetes_config_map.aws_auth: configmaps "aws-auth" already exists
```

So we have to provide a migration path.

dpiddockcmp:
Fun story: providers initialized with data sources don't work with `terraform import`. If you merge this please add, or link to, documentation of how to use `terraform import` with this module. 

May make the migration path interesting too.

https://www.terraform.io/docs/commands/import.html#provider-configuration

tristanpemble:
hey, I started to do something similar on my own branch before noticing this PR. with Terraform 0.12 you can use for loops to avoid template generation, not sure if that would be of interest. I prefer avoiding templates where possible.

here's what I had put together:

```hcl
locals {
  aws_auth_role_arns = [
    for role in compact(concat(
      coalescelist(
        aws_iam_instance_profile.workers_launch_template_mixed.*.role,
        data.aws_iam_instance_profile.custom_worker_group_launch_template_mixed_iam_instance_profile.*.role_name,
        [""]
      ),
      coalescelist(
        aws_iam_instance_profile.workers_launch_template.*.role,
        data.aws_iam_instance_profile.custom_worker_group_launch_template_iam_instance_profile.*.role_name,
        [""]
      ),
      coalescelist(
        aws_iam_instance_profile.workers.*.role,
        data.aws_iam_instance_profile.custom_worker_group_iam_instance_profile.*.role_name,
        [""]
      )
    )) : "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${role}"
  ]
}

resource "kubernetes_config_map" "aws_auth" {
  count = var.manage_aws_auth ? 1 : 0

  metadata {
    name      = "aws-auth"
    namespace = "kube-system"
  }

  data = {
    mapRoles = jsonencode(concat([
      for role_arn in local.aws_auth_role_arns : {
        rolearn  = role_arn
        username = "system:node:{{EC2PrivateDNSName}}"
        groups   = ["system:bootstrappers", "system:nodes"]
      }
    ], [
      for v in var.map_roles : {
        rolearn  = v.role_arn
        username = v.username
        groups   = [v.group]
      }
    ]))

    mapUsers = jsonencode([
      for v in var.map_users : {
        userarn  = v.user_arn
        username = v.username
        groups   = [v.group]
      }
    ])

    mapAccounts = jsonencode(var.map_accounts)
  }

  depends_on = [
    aws_eks_cluster.this
  ]
}
```

Vlaaaaaaad:
@tristanpemble I really like that approach and I'd really like to see something like that merged. maybe creating a new PR with that would help? 

Right now we are using Atlantis to manage cross-account EKS clusters and we're forced to set `manage_aws_auth = false`.

I can help with testing next week.

pierresteiner:
I this one still a thing? I would have loved to see that coming in the v6 release

sppwf:
Hi Guys,

I used this thing on post kubernetes stage - i have deployed helm tiller to prepare eks for helm app deployment.
it worked fine for me, soe we can try using kubernetes provider to manage the config map. Thats will be huge improvement, as we do not need to have kubectl and aws-iam-authenticator on box that creates the EKS cluster.

Do we have to do some additional testings or what is the status?

Thanks
Sergiu

barryib:
@sppwf The problem here is described by https://github.com/terraform-aws-modules/terraform-aws-eks/pull/355#issuecomment-494524353 and https://github.com/terraform-aws-modules/terraform-aws-eks/pull/355#issuecomment-494881422

So we need to provided a migration path for users who are already managing the `aws-auth` configmap with this module (so with `kubectl`)

erumble:
Recently I ran into an issue where having the AWS provider assume a role to create an EKS cluster would cause the local exec kubectl to fail. The local exec only had access to the env var credentials of the entity that assumed the role, not the assumed role's credentials. Because of this, I figured I'd take a stab at updating the vendored copy of this module (v6.0.1) to use the `kubernetes_config_map` resource. Turns out it was much easier than I though.

Here are the snippets:
```
# aws_auth.tf
resource "kubernetes_config_map" "aws_auth" {
  count = var.manage_aws_auth ? 1 : 0

  metadata {
    name      = "aws-auth"
    namespace = "kube-system"
  }

  data = {
    mapAccounts = yamlencode(var.map_accounts)
    mapRoles    = yamlencode(concat(local.workers_group_role_maps, var.map_roles))
    mapUsers    = yamlencode(var.map_users)
  }
}
```
```
# locals.tf
locals {
  # ... other local defs

  workers_group_role_maps = [
    for name in distinct(concat(
      aws_iam_instance_profile.workers_launch_template.*.role,
      data.aws_iam_instance_profile.custom_worker_group_launch_template_iam_instance_profile.*.role_name,
      aws_iam_instance_profile.workers.*.role,
      data.aws_iam_instance_profile.custom_worker_group_iam_instance_profile.*.role_name,
      )) : {
      rolearn  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${name}"
      username = "system:node:{{EC2PrivateDNSName}}"
      groups = [
        "system:bootstrappers",
        "system:nodes",
      ]
    }
  ]
}
```

Still doesn't address the migration issue though.

barryib:
@stijndehaes @dpiddockcmp @max-rocket-internet I successfully imported `aws-auth` configmap into terraform state.

Here is my config:

```hcl
# Add your backend provider with its configuration
# Example: aws provider if you use S3 remote state

# Kubernetes provider
provider "kubernetes" {
  # Tell the kubernetes provider to load your kubeconfig file
  # because you can't use data source during Terraform import
  load_config_file = true
}

resource "kubernetes_config_map" "aws_auth" {
  metadata {
    name      = "aws-auth"
    namespace = "kube-system"
  }
}
```

1. First setup your kube config and context to point to your EKS cluster `AWS_PROFILE=<your-aws-profile> aws eks --region <your-region> update-kubeconfig --name <your-cluster-name>`. You kubeconfig and creds will be used by the kubernetes provider to read the `
kube-system/aws-auth` configmap.
2. cd into your terraform configuration and import `terraform import kubernetes_config_map.aws_auth kube-system/aws-auth`

dpiddockcmp:
We should also document how to configure the kubernetes provider for use with the module in README.md as well as this upgrade path. This PR already had it in the basic example.

I think getting users to define the provider outside the module is a better pattern than the one used by the GKE module [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/v5.0.0/auth.tf#L29..L34).

stijndehaes:
Sorry for not working on this PR for so long. I will update it this weekend and add documentation and examples.

pierresteiner:
Could something like that work for the migration?:
```
  count      = 0
  provisioner "local-exec" {
    command = kubectl apply ...
  }
  provisioner "local-exec" {
    when = "destroy"
    command = kubectl delete ...
  }
}
resource "kubernetes_config_map" "aws_auth" {
  depends_on = [null_resource.update_config_map_aws_auth]
  ...
}
```
The idea is to let the provisionner remove the config map and have it recreated right away by the kubernetes provider.

stijndehaes:
@pierresteiner That is an option. Problem with that is that you have to run this with the user or role that created the cluster. So we would have to put a big warning in the release notes.

pierresteiner:
> @pierresteiner That is an option. Problem with that is that you have to run this with the user or role that created the cluster. So we would have to put a big warning in the release notes.
Indeed ! Hopefully in most cases it the the case, since it is executed by the ci/cd pipeline

stijndehaes:
Ok I updated the code. 
What we need to decide however is how we will do the migration:

- Explain people to use terraform import to migrate, which also gives people the option to manually delete the configmap before running the new terraform apply
- Kubectl delete the configmap automatically like @pierresteiner suggested.

@max-rocket-internet Do you have any preference?

outputs.tf
stijndehaes: I added the configmap as output, but it is a terraform resource now. Which is not the same as the rendered yaml. But I can't think of an easy way to achieve that

Vlaaaaaaad:
The second option might not be available to some people. We run all out Terraform automation through Atlantis which does not have `kubectl`. We're also doing some nested role assumption that I have no idea how well it work with `aws-iam-authenticator`.

I am firmly for the first option.

aws_auth.tf
```diff
@@ -66,21 +32,20 @@ data "template_file" "worker_role_arns" {
...
-    map_roles    = yamlencode(var.map_roles),
-    map_accounts = yamlencode(var.map_accounts)
+  metadata {
+    name      = "aws-auth"
+    namespace = "kube-system"
+  }
+
+  data = {
+    mapRoles    = <<EOF
+${join("", distinct(concat(data.template_file.launch_template_worker_role_arns.*.rendered, data.template_file.worker_role_arns.*.rendered)))}
```
barryib: Can we just remove these template files and use `yamlencode` juste like https://github.com/terraform-aws-modules/terraform-aws-eks/pull/355#issuecomment-507007419 ? This will prevent us to have configuration in multiple location and error in yaml template (wrong indentation by example).

max-rocket-internet:
> @max-rocket-internet Do you have any preference?

I think the first option is fine. It's quite normal in my experience when modules have updated or changed resources to import or rename resources 🙂

CarpathianUA:
I would probably send a bottle of gin to the author if this PR will be merged :)
We're also running applies via TF Cloud, so this option is a killer-feature!

examples/launch_templates/main.tf
```diff
@@ -23,6 +23,23 @@ provider "template" {
   version = "~> 2.1"
 }
 
+data "aws_eks_cluster" "cluster" {
+  name = module.eks.cluster_id
+}
+
+data "aws_eks_cluster_auth" "cluster" {
```
gdurandvadas: It seems there is inconsistency in the usage of the `aws_eks_cluster_auth`.
The first time runs perfectly, but the subsequents run it fails with `unathorized`.
My guess here is that once Terraform stored the token data in `data`, is already expired for the next runs. I couldn't find a way to force the token update on each run.
I've opened this issue: https://github.com/terraform-providers/terraform-provider-aws/issues/10362

pierresteiner: Are you sure that it is not working in that context? We faced the same problem and it was caused by the depends_on in the "aws_eks_cluster_auth" - which is present in your sample, but not this code.

If the dependency is induced by referencing the output of the module it seems to work.

gdurandvadas: The problem wasn't the `depends_on` it was that the cluster name must be got from the resource  *aws_eks_cluster*, and not from a var or local.

barryib:
I would love to see this merged. How can we move forward on this PR ? How can I help ?

mzupan:
This would be a big help for those of us in terraform cloud that want to manage `aws-auth` configmap

wperron:
Maybe I'm missing something but it doesn't seem to work for me. I was trying to do something similar with my configuration (in my case, setting a k8s service account) and found this PR. I tried setting up the provider with the `aws_eks_cluster_auth` data block just like the proposed change and it failed. I then tried simply applying the configuration using the fork as the source for the module and it still crashed. This is how I set my config:

```hcl
data "aws_eks_cluster" "this" {
    name = module.tools_cluster.cluster_id
}

data "aws_eks_cluster_auth" "this" {
    name = module.tools_cluster.cluster_id
}

provider "kubernetes" {
    host                   = data.aws_eks_cluster.this.endpoint
    cluster_ca_certificate = base64decode(data.aws_eks_cluster.this.certificate_authority.0.data)
    token                  = data.aws_eks_cluster_auth.this.token
    alias                  = "my_alias"
    version                = "~> 1.9"
}

module "eks" {
    # source        = "terraform-aws-modules/eks/aws"
    source        = "git@github.com:stijndehaes/terraform-aws-eks.git?ref=feature/aws-auth-k8s"
    # version       = "7.0.0"
    cluster_name  = "foobar"
    vpc_id        = "${module.vpc.vpc_id}"
    subnets       = "${module.vpc.private_subnets}"
    worker_groups = [
        {
            instance_type        = "t3.medium"
            asg_max_size         = 6
            asg_min_size         = 3
            asg_desired_capacity = 3
            autoscaling_enabled  = true
        }
    ]

    tags = "${local.aws_tags}"

    providers = {
        "kubernetes" = "kubernetes.my_alias"
    }
}
```

When the apply gets to the Kubernetes resources I get this error:

```
Error: Post https://<cluster-id>.gr7.us-east-1.eks.amazonaws.com/api/v1/namespaces/kube-system/configmaps: net/http: TLS handshake timeout

  on .terraform/modules/my-cluster/aws_auth.tf line 35, in resource "kubernetes_config_map" "aws_auth":
  35: resource "kubernetes_config_map" "aws_auth" {
```

CarpathianUA:
@wperron,
By reading your code snippet, it looks like you gather a certificate from a provisioned EKS cluster via `aws_eks_cluster` data source and then try to use is for `kubernetes` provider?

aws_auth.tf
stijndehaes: @barryib I think it is possible I just never could get it to work properly. Just tried it again. But by now the windows instances complicates stuff even more.

stijndehaes:
@max-rocket-internet I added some documentation on the breaking change and how to upgrade. It is very short but should contain all information. Also rebased and retested the PR everything is working. I used the basic example to test everything.

max-rocket-internet:
Really sorry for delays. Time is money 💸

Anyway, I've tested this and it works great! Love it. I imported the configmap like this:

```
terraform import module.cluster1.kubernetes_config_map.aws_auth[0] kube-system/aws-auth
```

And it worked perfectly. I also tested adding and removing a user. All good 🎉

README.md
```diff
@@ -18,9 +18,29 @@ Read the [AWS docs on EKS to get connected to the k8s dashboard](https://docs.aw
 
 ## Usage example
 
-A full example leveraging other community modules is contained in the [examples/basic directory](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/basic). Here's the gist of using it via the Terraform registry:
+A full example leveraging other community modules is contained in the [examples/basic directory](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/basic).
+Please do not forget to set the provider to the EKS cluster. This is needed to provision the aws_auth configmap in
```
max-rocket-internet: But I think we remove all of this and put it in changelog. If we start a precedence of putting change related notes in the README it will get messy quite fast.

docs/upgrading-to-aws-auth-kubernetes-provider.md
```diff
@@ -0,0 +1,14 @@
+# Upgrading from version <= 7.x to 8.0.0
```
max-rocket-internet: I think we move these notes to the changelog as this will be the first place people look for upgrade notes.

Historically we haven't written much more than `breaking` for things like this but I think we should start writing more detail in the changelog and this change is the perfect example 😃

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Updated instance_profile_names and instance_profile_arns outputs to also consider launch template as well as asg (by @ankitwal)
+- **Breaking:** Configure the aws-auth configmap using the terraform kubernetes providers. Read the [docs](docs/upgrading-to-aws-auth-kubernetes-provider.md) for more info (by @sdehaes)
```
max-rocket-internet: Maybe add a new subheading under here like `### Notes` and detail the import and other notes.

max-rocket-internet:
I requested a few cosmetic changes just around where to place the upgrade notes. Let's resolve the conflicts and then we can merge ASAP 🙂

Thanks again @stijndehaes for your work on this, it's super appreciated 🙏

max-rocket-internet:
Actually, in the interest of moving things forward quicker, I'll just merge this and fix the docs myself 🚀

ivanich:
With this change I'm getting
`Error: Post https://FGFFFFFFFJDJKDNDHEJFBF.yl4.us-east-1.eks.amazonaws.com/api/v1/namespaces/kube-system/configmaps: dial tcp xx.xx.xx.xx:443: i/o timeout on .terraform/modules/eks/aws_auth.tf line 45, in resource "kubernetes_config_map" "aws_auth":
  45: resource "kubernetes_config_map" "aws_auth" {
`
Are you guys sure it was a good idea to merge this PR?

Vlaaaaaaad:
As an aside, I am using the Terraform Kubernetes provider to manage some CMs for like... very easy service discovery and I can confirm that it’s often timing out( cluster not yet ready, DNS not yet propagated, and so on)  or failing with unauthorized( no idea why). A second plan and apply usually works.

max-rocket-internet:
Interesting.

Whether you are using `kubectl` like before or the new provider config that is now merged, it wouldn't make any difference if you can't connect to the API endpoint or if DNS is not propagated.

BUT before there was `for i in `seq 1 10`; do xxx` so it would retry 10 times. Perhaps that's the issue here?

ivanich:
> BUT before there was for i in seq 1 10; do xxx so it would retry 10 times. Perhaps that's the issue here?
I'm sure it is, and with this PR we can't do any retries anymore, in my opinion previous implementation was ugly but it worked as it should.

max-rocket-internet:
https://github.com/terraform-providers/terraform-provider-kubernetes/issues/96

rchernobelskiy:
Yeah we often observe eks master timeouts for a minute or two after creation

PR #369, opened 2019-05-02
Disabling ASG process AZRebalance by default

max-rocket-internet:
# PR o'clock

## Description

Disabling ASG process `AZRebalance` by default.

I can't think of a single reason why this wouldn't be a good idea to have disabled by default? If the ASG makes any changes to your instances, they need to be drained first and not terminated at will by the ASG.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+ - Set default suspended processes for ASG to `AZRebalance` (by @max-rocket-internet)
  - Add .prettierignore file (by @rothandrew)
  - Switch to https for the pre-commit repos (by @rothandrew)
  - Add instructions on how to enable the docker bridge network (by @rothandrew)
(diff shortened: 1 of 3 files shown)


max-rocket-internet:
Some extra info:
https://github.com/kubernetes/autoscaler/pull/1802
https://github.com/kubernetes/kops/issues/3319

rchernobelskiy:
One reason I can think of to leave it enabled would be to prevent pods being unschedulable due to their ebs volumes being in a zone without a node.
Or is there a better way to prevent that?
Seems like the recommended solution of asg per zone would be pretty convoluted terraform.

max-rocket-internet:
> to prevent pods being unschedulable due to their ebs volumes being in a zone without a node.

This doesn't change the problem though: the ASG doesn't drain nodes gracefully, it just terminates them. 

EBS volume AZ is a problem, regardless of this setting. I think yes you need an worker group per AZ then. And also ensure your cluster-autoscaler is configured correctly 🙂

PR #371, opened 2019-05-03
Adding new mixed type of worker group with instance overrides and mixed instances policy

max-rocket-internet:
# PR o'clock

## Description

This PR adds a third worker group type, *Launch Template Mixed*, that includes 4x instance type overrides and a `mixed_instances_policy` in the ASG.

Also some cosmetic formatting changes for count and lifecycle options.

#### FAQ

> Why the added complexity?

Advanced users will be able to use this to do things like running mixed instance and lifecycle types in the same ASG. This has very large cost benefit for users who want to use either spot or reserved instances without giving up scalability.

> Why choose 4 instance type overrides?

In TF it's currently not possible to let the user choose the amount as each item is a separate configuration block in the ASG resource. I think 4 is a good choice as there's a natural clustering for instances of similar types, e.g. `c/m/r/t + .large` or `c/m/r/t + .2xlarge`.

> Why can't this be done in the existing work group?

Because each override and `mixed_instances_policy` is a separate configuration block in the ASG resource.

### Checklist

### DIFF
--- a/workers_launch_template_mixed.tf
+++ b/workers_launch_template_mixed.tf
@@ -1,0 +1,123 @@
+# Worker Groups using Launch Templates with mixed instances policy
+
+resource "aws_autoscaling_group" "workers_launch_template_mixed" {
+  name_prefix             = "${aws_eks_cluster.this.name}-${lookup(var.worker_group_launch_template_mixed[count.index], "name", count.index)}"
+  desired_capacity        = "${lookup(var.worker_group_launch_template_mixed[count.index], "asg_desired_capacity", local.workers_group_defaults["asg_desired_capacity"])}"
+  max_size                = "${lookup(var.worker_group_launch_template_mixed[count.index], "asg_max_size", local.workers_group_defaults["asg_max_size"])}"
+  min_size                = "${lookup(var.worker_group_launch_template_mixed[count.index], "asg_min_size", local.workers_group_defaults["asg_min_size"])}"
+  force_delete            = "${lookup(var.worker_group_launch_template_mixed[count.index], "asg_force_delete", local.workers_group_defaults["asg_force_delete"])}"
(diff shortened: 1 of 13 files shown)


workers_launch_template_mixed.tf
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  placement {
+    tenancy    = "${lookup(var.worker_group_launch_template_mixed[count.index], "launch_template_placement_tenancy", local.workers_group_defaults["launch_template_placement_tenancy"])}"
+    group_name = "${lookup(var.worker_group_launch_template_mixed[count.index], "launch_template_placement_group", local.workers_group_defaults["launch_template_placement_group"])}"
+  }
+
+  count = "${var.worker_group_launch_template_mixed_count}"
+
+  lifecycle {
```
antonbabenko: I'd move `lifecycle` (as meta-argument) to the end if `resource` block.

workers_launch_template_mixed.tf
```diff
@@ -0,0 +1,123 @@
...
+  monitoring {
+    enabled = "${lookup(var.worker_group_launch_template_mixed[count.index], "enable_monitoring", local.workers_group_defaults["enable_monitoring"])}"
+  }
+
+  placement {
+    tenancy    = "${lookup(var.worker_group_launch_template_mixed[count.index], "launch_template_placement_tenancy", local.workers_group_defaults["launch_template_placement_tenancy"])}"
+    group_name = "${lookup(var.worker_group_launch_template_mixed[count.index], "launch_template_placement_group", local.workers_group_defaults["launch_template_placement_group"])}"
+  }
+
+  count = "${var.worker_group_launch_template_mixed_count}"
```
antonbabenko: Put `count` at the top of `resource.

workers_launch_template_mixed.tf
max-rocket-internet: Good idea. I'll do it for the whole module as it looks nicer.

workers_launch_template_mixed.tf
max-rocket-internet: Good idea. I'll do it for the whole module as it looks nicer.

PR #387, opened 2019-06-04
Enable log retention for cloudwatch log groups

yuriipolishchuk:
# PR o'clock

## Description

Enable log retention for cloudwatch log groups. 
https://www.terraform.io/docs/providers/aws/r/eks_cluster.html#enabling-control-plane-logging

Breaking change:
If logging is already enabled, the cloudwatch log group resource cannot be created.
As a workaround disable logging, delete old log group and apply plan again.

### Checklist

- [x] CI tests are passing


version
```diff
@@ -1 +1 @@
-v4.0.2
+v4.0.3
```
max-rocket-internet: Please leave this file as is, it's only changed when a new release is created.

yuriipolishchuk-finsight: @max-rocket-internet done

yuriipolishchuk: @max-rocket-internet done

variables.tf
```diff
@@ -4,6 +4,11 @@ variable "cluster_enabled_log_types" {
   type        = "list"
 }
 
+variable "cluster_log_retention_in_days" {
+  default     = "0"
```
max-rocket-internet: I think maybe a default of `30` or `90` might be more sensible. What do you think?

max-rocket-internet: Cloudwatch logs can get expensive and on a busy cluster there is a lot of messages. So maybe let's go with `90`.

yuriipolishchuk: @max-rocket-internet makes sense

yuriipolishchuk: @max-rocket-internet set default retention to 90 days

Issue #150, opened 2018-10-02
Improve iam and security group naming convention 


simonmacklin:
## I'm submitting a...

* [ ] bug report
- [x] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

At the moment security groups and IAM roles are not immediately clear  on what infrastructure they belong too.  As a example if I name the cluster 'test' the cluster iam role is named something like 

test201810021044448743111111112

I can understand the prefix is to prevent collisions but in my opinion it would be better to prefix it with something like eks-${var.cluster_name}-role- which would up up with something like 

eks-test-role-201810021044448743111111112

## Are you able to fix this problem and submit a PR? Link here if you have already.

I have not submitted a PR yet until I can get an idea of everyones opinion but will be more then happy too.

Kind Regards

Simon

max-rocket-internet:
I agree.

I also created a similar issue here: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/88

I don't think the "prevent collisions" idea holds up, as what are the chances of you having a role of `eks-cluster-name` elsewhere?

simonmacklin:
Hi Max 

I agree totally.  I am more than happy to submit a pull request without any prefixing.

Many thanks

max-rocket-internet:
> I am more than happy to submit a pull request without any prefixing

Awesome 👍

dpiddockcmp:
Please note that changing the IAM role names will cause all existing clusters to be recreated. I agree that the naming sucks but it's now technical debt.

A change of this magnitude should wait for a major version number change and have some form of migration path for existing clusters. Something similar to #137 for the EKS cluster resource.

> I don't think the "prevent collisions" idea holds up, as what are the chances of you having a role of `eks-cluster-name` elsewhere?

Justification for the use of `name_prefix` was given here: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/88#issuecomment-412424297

Currently you can use the same terraform config in multiple regions as IAM is global.

max-rocket-internet:
> Please note that changing the IAM role names will cause all existing clusters to be recreated.

Yeah this is really annoying. You can't change the EKS cluster role without recreating it.

> Currently you can use the same terraform config in multiple regions as IAM is global.

This is true. The IAM role names will be unique but the cluster names could be identical.

Any more opinions on this? Personally, I could go either way.

simonmacklin:
[...]

Issue #88, opened 2018-08-09
Use of name_prefix


max-rocket-internet:
Currently we have:

```hcl
resource "aws_iam_role" "workers" {
  name_prefix        = "${aws_eks_cluster.this.name}"
  assume_role_policy = "${data.aws_iam_policy_document.workers_assume_role_policy.json}"
}

resource "aws_iam_instance_profile" "workers" {
  name_prefix = "${aws_eks_cluster.this.name}"
  role        = "${aws_iam_role.workers.name}"
}
```

Is there a reason to use `name_prefix` instead of just `name`? I ask because the resultant names are things like `my-cluster-20180808095045107900000005`.

We have to create cross account IAM policies for things like ECR and it would be nice to have a predictable and consistent name for the roles 🙂

brandoconnor:
Seems to be my personal preference to use `name_prefix` but both here now and in my ALB module, people seem to want the opposite. To lay it out there:

Brandon's opinion:

* `name` leaves room for collisions but `name_prefix` does not. Plays better with testing and means a consumer never needs to worry about creating the same resource in the same region/environment/account/project and hitting a name conflict. idk maybe that's not such a big deal.
* I care much more about a resource's ARN or its tags than its `name`. ARNs are hairy enough that I assume them to be consumed through remote states but I can see how cross-account needs might not render that possible...
* I (perhaps wrongly) assumed max resource lengths are handled for you with name prefixes... though now I'm second guessing whether this is the case for all resources. Also, unsure of how a `name_prefix` longer than the allowed name behaves...

Obviously, the drawback of `name_prefix` being the losing of control of names that become long and ugly.

In any case, I've seen this requested and never the opposite... not really enough data to say what consensus looks like but I'd be willing to make that switch over on the next major release. PR away.

brandoconnor:
I think I've evolved in my thought here as I've built some more things in the wild and use-cases have stacked up for this feature. I think there's a good case to be made for 100% user specified names of the resources in question rather than using `name_prefix`. This may be a breaking change but worth the trouble.

If a user wants to give randomization to any/all resource names, the module still allows that without enforcing it (i.e. append your `cluster_name` with `random_string.suffix.result`). I've come around that configuration details like naming should generally be passed, rather than assumed or enforced. Thanks for linking the issue and consider the field open for PR here. 👍

max-rocket-internet:
> Thanks for linking the issue and consider the field open for PR here.

Nice!

[...]

PR #390, opened 2019-06-06
adds tags to resource aws_iam_role.cluster

camilosantana:
# PR o'clock

## Description

We had a need in my company to tag the iam role created by this module so it can be found in other projects via terraform `data` resource. This adds that capability.

### Checklist

- [x] CI tests are passing


version
```diff
@@ -1 +1 @@
-v4.0.2
+v4.0.3
```
max-rocket-internet: Please remove this as it's only changed on release.

cluster.tf
```diff
@@ -54,11 +54,12 @@ resource "aws_security_group_rule" "cluster_https_worker_ingress" {
 
 resource "aws_iam_role" "cluster" {
   count                 = "${var.manage_cluster_iam_resources ? 1 : 0}"
-  name_prefix           = "${var.cluster_name}"
+  name                  = "${local.iam_role_name}"
```
max-rocket-internet: This will cause the role to be recreate, correct? What happens to the cluster when this happens? Is it also recreated?

camilosantana: > This will cause the role to be recreate, correct
yes, this is a resource recreate.

> What happens to the cluster when this happens? Is it also recreated?
yeah, we get a `forces new resource` on the cluster too, I think. Would you want me to confirm? i can do so after lunch today.

version
camilosantana: consider it done.

camilosantana: 72b18fc2d3a2185de6af4d4342845b3553e66781

cluster.tf
camilosantana: @max-rocket-internet here are the resources that want to be recreated when `aws_iam_role` is changed.

```hcl
-/+ module.eks.aws_eks_cluster.this (new resource required)
-/+ module.eks.aws_iam_role.cluster (new resource required)
-/+ module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSClusterPolicy (new resource required)
-/+ module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSServicePolicy (new resource required)
-/+ module.eks.aws_launch_configuration.workers (new resource required)
-/+ module.eks.null_resource.update_config_map_aws_auth (new resource required)
```

camilosantana: after reading [some issue comments](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/150#issuecomment-427057836) it seems this is a bigger deal than i expected or consider.

@max-rocket-internet do we want to go ahead and close this out or evolve this more in line with general consensus (based on existing issue read throughs)?

my case is that i use cloud posse's [null-label module](https://github.com/cloudposse/terraform-null-label) and pass `"${module.null-label_this.tags}"` to any resource that supports tagging.

once EKS is up, it has an iam role. it's useful to be able to find said role via `data` resource filtering on tags instead of doing a `remote_state` call (undocumented but i do remember seeing a post where it was silently added during one of the later updates to tf 0.11).

I'll follow the intent [here](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/88#issuecomment-482757743) and add region to the role name suffix instead of random string.

max-rocket-internet: > it seems this is a bigger deal than i expected or consider.

Exactly, that's what I'm getting at. We can't merge it if it will recreate the cluster 😅

max-rocket-internet: I still don't get why you need to change the name of the role just add tags?

max-rocket-internet: Or why we need a random string in the name?

camilosantana: > I still don't get why you need to change the name of the role just add tags

we don't need to change the name. that was an artifact for a previous effort here at work. cleaning up now.

camilosantana: > why we need a random string in the name?

the intent was to as closely match the result of `name_prefix` - again for something internal here at work. removed.

camilosantana: > We can't merge it if it will recreate the cluster 😅

totally. thanks for your patience.

max-rocket-internet:
> totally. thanks for your patience.

No worries and thanks again for your contribution @camilosantana!

PR #393, opened 2019-06-10
Termination Policy Option to worker ASG

undeadops:
# PR o'clock

## Description

Add option to change worker asg termination policy.

### Checklist

- [ X ] `terraform fmt` and `terraform validate` both work from the root and `examples/*` directories
- [ X ] CI tests are passing
- [ X ] I've added my change to CHANGELOG.md and highlighted any breaking changes
- [ X ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -46,7 +46,6 @@ locals {
     service_linked_role_arn       = ""                              # Arn of custom service linked role that Auto Scaling group will use. Useful when you have encrypted EBS
     termination_policies          = "Default"                       # A comma delimited list of policies to decide how the instances in the auto scale group should be terminated. 
 
-
     # Settings for launch templates
     root_block_device_name            = "${data.aws_ami.eks_worker.root_device_name}" # Root device name for workers. If non is provided, will assume default AMI was used.
     root_kms_key_id                   = ""                                            # The KMS key to use when encrypting the root storage device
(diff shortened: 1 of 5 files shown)


local.tf
```diff
@@ -44,6 +44,7 @@ locals {
     enabled_metrics               = ""                              # A comma delimited list of metrics to be collected i.e. GroupMinSize,GroupMaxSize,GroupDesiredCapacity
     placement_group               = ""                              # The name of the placement group into which to launch the instances, if any.
     service_linked_role_arn       = ""                              # Arn of custom service linked role that Auto Scaling group will use. Useful when you have encrypted EBS
+    termination_policies          = "Default"                       # A comma delimited list of policies to decide how the instances in the auto scale group should be terminated.
```
max-rocket-internet: Can we just have this default to `""` like all the other options? Otherwise we get unwanted changes:

```
  ~ module.xxxx.aws_autoscaling_group.workers_launch_template
      termination_policies.#: "0" => "1"
      termination_policies.0: "" => "Default"
```

undeadops: Sure thing, sorry didn't think about that.  AWS's default is "Default" so I was setting it explicitly.  I've made the change.

PR #409, opened 2019-06-24
added market options to request spot instances without pools

onur-sam-gtn-ai:
# PR o'clock

## Description
Correct me if I'm wrong: Currently, the only way to request spot instances in a launch template is by specifying pools. The change here may be not ideal, but it seems to have worked for me. 
Going by this:
https://www.terraform.io/docs/providers/aws/r/launch_template.html#market-options

It should be possible to implement this, maybe not ideally with this PR but with a similar approach. This is more a feature request than a full PR.

Reference: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotMarketOptions.html

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -80,6 +80,7 @@ locals {
     spot_allocation_strategy                 = "lowest-price"                                   # The only valid value is lowest-price, which is also the default value. The Auto Scaling group selects the cheapest Spot pools and evenly allocates your Spot capacity across the number of Spot pools that you specify.
     spot_instance_pools                      = 10                                               # "Number of Spot pools per availability zone to allocate capacity. EC2 Auto Scaling selects the cheapest Spot pools and evenly allocates Spot capacity across the number of Spot pools that you specify."
     spot_max_price                           = ""                                               # Maximum price per unit hour that the user is willing to pay for the Spot instances. Default is the on-demand price
+    use_spot_instances                       = false                                            # set to true to use spot instances with a LAUNCH TEMPLATE
   }
 
   workers_group_defaults = merge(
--- a/local.tf
+++ b/local.tf
@@ -80,7 +80,7 @@ locals {
     spot_allocation_strategy                 = "lowest-price"                                       # The only valid value is lowest-price, which is also the default value. The Auto Scaling group selects the cheapest Spot pools and evenly allocates your Spot capacity across the number of Spot pools that you specify.
     spot_instance_pools                      = 10                                                   # "Number of Spot pools per availability zone to allocate capacity. EC2 Auto Scaling selects the cheapest Spot pools and evenly allocates Spot capacity across the number of Spot pools that you specify."
     spot_max_price                           = ""                                                   # Maximum price per unit hour that the user is willing to pay for the Spot instances. Default is the on-demand price
-    market_type                              = null                                             # set to "spot" to use spot instances without defining pools
+    market_type                              = null                                                 # set to "spot" to use spot instances without defining pools
   }
 
   workers_group_defaults = merge(
--- a/local.tf
+++ b/local.tf
@@ -80,7 +80,7 @@ locals {
     spot_allocation_strategy                 = "lowest-price"                                   # The only valid value is lowest-price, which is also the default value. The Auto Scaling group selects the cheapest Spot pools and evenly allocates your Spot capacity across the number of Spot pools that you specify.
     spot_instance_pools                      = 10                                               # "Number of Spot pools per availability zone to allocate capacity. EC2 Auto Scaling selects the cheapest Spot pools and evenly allocates Spot capacity across the number of Spot pools that you specify."
     spot_max_price                           = ""                                               # Maximum price per unit hour that the user is willing to pay for the Spot instances. Default is the on-demand price
-    use_spot_instances                       = false                                            # set to true to use spot instances with a LAUNCH TEMPLATE
+    market_type                              = null                                             # set to "spot" to use spot instances without defining pools
   }
 
   workers_group_defaults = merge(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
  - Option to set a KMS key for the log group and encrypt it (by @till-krauss)
  - Output the name of the cloudwatch log group (by @gbooth27)
+ - Option to use spot instances with launch templates without defining pools, especially useful for GPU instance types (@onur-sam-gtn-ai)
  - Write your awesome addition here (by @you)
 
 ### Changed
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -222,10 +222,10 @@ resource "aws_launch_template" "workers_launch_template" {
   }
 
   instance_market_options {
-    market_type =   lookup(
-        var.worker_groups_launch_template[count.index],
-        "use_spot_instances",
-        local.workers_group_defaults["use_spot_instances"]) ? "spot" : null
+    market_type = lookup(
+      var.worker_groups_launch_template[count.index],
+      "use_spot_instances",
+    local.workers_group_defaults["use_spot_instances"]) ? "spot" : null
   }
 
   block_device_mappings {
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -224,8 +224,9 @@ resource "aws_launch_template" "workers_launch_template" {
   instance_market_options {
     market_type = lookup(
       var.worker_groups_launch_template[count.index],
-      "use_spot_instances",
-    local.workers_group_defaults["use_spot_instances"]) ? "spot" : null
+      "market_type",
+      local.workers_group_defaults["market_type"],
+    )
   }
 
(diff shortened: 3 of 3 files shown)


max-rocket-internet:
### Checklist

- [ ] CI tests are passing

max-rocket-internet:
What difference does this actually make? Because the Terraform documentation says:

> The market type. Can be `spot`.

Like it only has a single option?

onur-sam-gtn-ai:
If "instance_market_options" is not provided, it creates on-demand instances. If instance_market_options are provided **and** market_type is spot, it creates spot instances only.

local.tf
```diff
@@ -80,6 +80,7 @@ locals {
     spot_allocation_strategy                 = "lowest-price"                                   # The only valid value is lowest-price, which is also the default value. The Auto Scaling group selects the cheapest Spot pools and evenly allocates your Spot capacity across the number of Spot pools that you specify.
     spot_instance_pools                      = 10                                               # "Number of Spot pools per availability zone to allocate capacity. EC2 Auto Scaling selects the cheapest Spot pools and evenly allocates Spot capacity across the number of Spot pools that you specify."
     spot_max_price                           = ""                                               # Maximum price per unit hour that the user is willing to pay for the Spot instances. Default is the on-demand price
+    use_spot_instances                       = false                                            # set to true to use spot instances with a LAUNCH TEMPLATE
```
max-rocket-internet: Can we just have `null` here and then no conditional? i.e.

```
market_type = lookup(
      "use_spot_instances",
    local.workers_group_defaults["use_spot_instances"])
  }
```

max-rocket-internet:
Sorry for the delay!

>  If instance_market_options are provided and market_type is spot, it creates spot instances only.

With this PR, it's always provided, it's just to either `null` or `spot`. Are you sure that when it's `null`, there is no error and on-demand instances are used? Can you test that?

onur-sam-gtn-ai:
It was in the back of my head and I really wanted to test it but I've been assigned to another task for the next couple of weeks. I'll keep you updated.

onur-sam-gtn-ai:
Seems like my lazy implementation actually works:

_Purchasing option: Not specified_
As per the AWS Console:
> The purchasing request. The valid values are Spot Instance or Not specified. If this value is not specified in the template, EC2 launches an On-Demand instance by default.

local.tf
onur-sam-gtn-ai: I don't exactly get why... I just wanted a simple flag it to tell it to use spot in the launch template?

max-rocket-internet: Because now you have `value` > `conditional` > `other value`.

e.g. `true/false` > `conditional` > `string(spot)/null`

But it's simpler, and more like most of the other attributes, to just have the value itself with a default of `""`. See what I mean?

max-rocket-internet:
> Purchasing option: Not specified

Nice 🎉

Just address my comment in `local.tf` and we can merge.

local.tf
onur-sam-gtn-ai: so... a `market_type = "spot" | ""` instead? Less code change but less user friendly? Telling the user to leave it as `""` to use on-demand and use `"spot"` for spot instances, did I catch that right? Sorry I'm too rookie with Terraform, hence all the questions.

max-rocket-internet: Yes, exactly that.

onur-sam-gtn-ai:
tested with both default null and "spot", working fine.

onur-sam-gtn-ai:
> I'm very sorry I didn't see this before but can you also add this item to the `aws_launch_template.workers_launch_template` resource so it's consistent?

Huh? Sorry I'm confused. Isn't that what I already submitted?

`workers_launch_template.tf`
```
  instance_market_options {
    market_type = lookup(
      "market_type",
      local.workers_group_defaults["market_type"],
    )
  }
```
Or did you mean the `_mixed` one?

max-rocket-internet:
> Or did you mean the _mixed one?

Yes, I meant the `_mixed` one. Apologies, I'm doing 10 things at once right now 🙂

onur-sam-gtn-ai:
> > Or did you mean the _mixed one?
> 
> Yes, I meant the `_mixed` one. Apologies, I'm doing 10 things at once right now 🙂

Isn't the `mixed` strategy literally `a mix of on-demand and spot instances`? Which contradicts what this PR does (running with **spot-only** instances)?

max-rocket-internet:
Nope, it's for mixed instance types for spot instances. For diversity so spot requests are definitely fulfilled. Example here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/spot_instances/main.tf

onur-sam-gtn-ai:
That's not what AWS documents say:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-mixedinstancespolicy.html
`With mixed instances, your Auto Scaling group can provision a combination of On-Demand Instances and Spot Instances across multiple instance types.`
Even then, it still doesn't make sense to add this to the `Mixed` file. Mixed policy already says it will create **spot** instances using pools and mixed instance types.  (ex. `on_demand_percentage_above_base_capacity`, `spot_allocation_strategy`, etc.) The only option being going 0% on the on-demand capacity which doesn't make sense to mix with a single-type launch template which is the purpose of my change.

My change allows the use of spots with no percentage of on-demand, no pools and single instance type. It's a single-type approach as per autoscaling group design. It doesn't make sense to me to add a "spot_instances" flag to a launch template that already creates mixed instance types that inherently create spot instances based on the percentage anyway. Besides, `market_type="spot"` shouldn't be provided when using pools and mixed instance types.

max-rocket-internet:
> That's not what AWS documents say

Sorry, I meant that `workers_launch_template_mixed.tf` was written specifically to run multiple spot instance types in a single ASG or a mix of on-demand and spot, but using the `override_instance_types`.

> Even then, it still doesn't make sense to add this to the Mixed file. Mixed policy already says it will create spot instances using pools and mixed instance types.

OK cool.

> My change allows the use of spots with no percentage of on-demand, no pools and single instance type.

OK understood.

max-rocket-internet:
@onur-sam-gtn-ai I don't think you tested this properly. It causes changes to be applied on every run:

```
  # module.eks.aws_launch_template.workers_launch_template[1] will be updated in-place
  ~ resource "aws_launch_template" "workers_launch_template" {
        arn                     = "arn:aws:ec2:us-west-2:051129233020:launch-template/lt-089cfc4750f99604f"
...
        instance_type           = "t2.medium"
      ~ latest_version          = 2 -> (known after apply)
        name                    = "test-eks-lt-1urG22xZ-worker-group-22019073013374180530000000f"
...
      + instance_market_options {
        }
....
```

With worker group like this:

```hcl
    {
      asg_desired_capacity = 2
      public_ip            = true
    },
    {
      name                 = "worker-group-2"
      asg_desired_capacity = 1
      public_ip            = true
    },
  ]
```

I will revert it in the new release.

onur-sam-gtn-ai:
That's because of the null you've told me to create. The parent object "instance_market_options" is requested to be created with "null" and accepted by the API but the actual object created completely ignores "instance_market_options" because it's null. If you proceed and apply it, you'll see that it'll try to add an "instance_market_options" with nothing inside and get accepted but the actual template itself will not be changed (except for a version bump).
My initial method was not creating "instance_market_options" at all if the flag was "false".
I guess I'll have to implement it that way again.

max-rocket-internet:
> The only workaround for this is to not provide "instance_market_options" at all

Exactly.

This problem is super common in Terraform 😃 I think this is improved in Terraform 0.12 though: https://www.terraform.io/upgrade-guides/0-12.html#attributes-vs-blocks

So perhaps you can try again using the special `dynamic` block type.

TarekAS:
I added this to ```workers_launch_template``` and it seems to work:

```
dynamic "instance_market_options" {
    for_each = lookup(
      "use_spot_instances",
    local.workers_group_defaults["use_spot_instances"]) ? [1] : []
    content {
      market_type = "spot"
    }
  }
```
The only thing left would be support extra spot parameters like ```spot_max_price``` for non-mixed policy launch templates.

onur-sam-gtn-ai:
> I added this to `workers_launch_template` and it seems to work:
> 
> ```
> dynamic "instance_market_options" {
>     for_each = lookup(
>       var.worker_groups_launch_template[count.index],
>       "use_spot_instances",
>     local.workers_group_defaults["use_spot_instances"]) ? [1] : []
>     content {
>       market_type = "spot"
>     }
>   }
> ```
> 
> The only thing left would be support extra spot parameters like `spot_max_price` for non-mixed policy launch templates.
> 
> Edit: turns out this also causes changes to be applied on every run. The only other real solution is to just create another `workers_launch_template_spot.tf` with spot enabled.

Instead of `? [1] : []` , maybe `? [1] : null` would work as expected? Since `[]` is still an object, I guess the dynamic block is still called, while `for_each = null` shouldn't trigger anything (at least in theory)?

TarekAS:
@onur-sam-gtn-ai You are correct. Check out: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/463

Issue #253, opened 2019-01-28
No No method of specifying a credit specification for burstable performance worker instances


bmacauley-reward:
# I have issues

## I'm submitting a...

 feature request

## What is the current behavior?

No method of specifying a credit specification for burstable performance worker instances
eg t2.medium ...defaults to standard, would like to set it to unlimited

## What's the expected behavior?
Configuration parameter for credit specification...

aws_launch_template ...rather than aws_launch_configuration?
https://www.terraform.io/docs/providers/aws/r/launch_template.html

## Environment details

* Affected module version: v2.1.0
* OS: MacOS
* Terraform version: 0.11.11

max-rocket-internet:
Cool. Feel free to make a PR 🙂

a-shink:
I started coding, have result and will send Merge request ASAP.

For T3 instances this parameter is "unlimited" by default. As result peoples can have additional charges even in Free tier or "surprises" like 68$ price per instance instead 3.6$ expected. Check detailed explanation at https://advancedweb.hu/2019/03/05/t3_unlimited/
The module still don't configure this param so better do it. I think better default is have "limited" mode.

This parameter can be configured for Launch Templates only, don't work for Launch Configuration even via AWS API.

a-shink:
I just did it #426

PR #426, opened 2019-07-07
Configure 'cpu_credits' for workers (#253)

a-shink:
# PR o'clock

## Description

This PR makes configurable 'T2/T3 unlimited' param for workers.
Used 'standard' mode as default to avoid unexpected paying higher costs.
This parameter can be configured for Launch Templates only, don't work for Launch Configuration even via AWS API.

### Checklist

- [X] `terraform fmt` and `terraform validate` both work from the root and `examples/*` directories
- [X] CI tests are passing
- [X] I've added my change to CHANGELOG.md and highlighted any breaking changes

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -49,7 +49,6 @@ locals {
     bootstrap_extra_args          = ""                         # Extra arguments passed to the bootstrap.sh script from the EKS AMI.
     additional_userdata           = ""                         # userdata to append to the default userdata.
     ebs_optimized                 = true                       # sets whether to use ebs optimization on supported types.
-    cpu_credits                   = "standard"                 # T2/T3 unlimited mode, can be 'standard' or 'unlimited'. Used 'standard' mode as default to avoid paying higher costs
     enable_monitoring             = true                       # Enables/disables detailed monitoring.
     public_ip                     = false                      # Associate a public ip address with a worker
     kubelet_extra_args            = ""                         # This string is passed directly to kubelet if set. Useful for adding labels or taints.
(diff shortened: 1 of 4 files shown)


local.tf
```diff
@@ -49,6 +49,7 @@ locals {
     bootstrap_extra_args          = ""                         # Extra arguments passed to the bootstrap.sh script from the EKS AMI.
     additional_userdata           = ""                         # userdata to append to the default userdata.
     ebs_optimized                 = true                       # sets whether to use ebs optimization on supported types.
+    cpu_credits                   = "standard"                 # T2/T3 unlimited mode, can be 'standard' or 'unlimited'. Used 'standard' mode as default to avoid paying higher costs
```
nauxliu: Use `unlimited` as default is better because T3 instances start in Unlimited mode by default

max-rocket-internet: Hmm to quote the TF doc:

> T3 instances are launched as unlimited by default. T2 instances are launched as standard by default.

So not sure what default we want. Maybe stick `standard` as it's a safer option in terms of costs.

max-rocket-internet:
Thanks for the PR @a-shink. Can you please add your change to the changelog?

local.tf
nauxliu: @a-shink This variable should move to line 76  which is Settings for launch templates

a-shink: @nauxliu Done 00faa0e

a-shink: @nauxliu Is something blocking approving the PR? I don't change defaults because @max-rocket-internet thinks in similar way. Many peoples thinks about 'unlimited' default as money trap.

PR #446, opened 2019-07-25
VPC: `enable_dns_hostnames = true` in examples

karolinepauls:
# PR o'clock

## Description

AWS docs state:

> Your VPC must have DNS hostname and DNS resolution support. Otherwise, your worker nodes cannot register with your cluster. For more information, see Using DNS with Your VPC in the Amazon VPC User Guide. 

- https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html

From that it seems like it should always be enabled, no matter what kind of VPC is used. I added it to the basic example which I used and had error messages with DNS resolution appearing in the logs on the nodes. Interestingly enough, without DNS resolution the cluster appears to sometimes work and sometimes not. It's hard to run experiments though, so as per the docs, it should be enabled.

Question: do we want to play it safe and enable DNS resolution in all examples (even public VPC)?

### Checklist

- [v] `terraform fmt` and `terraform validate` both work from the root and `examples/*` directories
- [ ] CI tests are passing
- [v] I've added my change to CHANGELOG.md and highlighted any breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
  - Update default override instance types to work with Cluster Autoscaler (by @nauxliu on behalf of RightCapital)
+ - Examples now specify `enable_dns_hostnames = true`, as per [EKS documentation](https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html) (by @karolinepauls)
  - Write your awesome change here (by @you)
 
 # History
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Read the [AWS docs on EKS to get connected to the k8s dashboard](https://docs.aw
 
 * You want to create an EKS cluster and an autoscaling group of workers for the cluster.
 * You want these resources to exist within security groups that allow communication and coordination. These can be user provided or created within the module.
-* You've created a Virtual Private Cloud (VPC) and subnets where you intend to put the EKS resources.
+* You've created a Virtual Private Cloud (VPC) and subnets where you intend to put the EKS resources. The VPC satisfies [EKS requirements](https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html).
 * If `manage_aws_auth = true`, it's required that both [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl) (>=1.10) and [`aws-iam-authenticator`](https://github.com/kubernetes-sigs/aws-iam-authenticator#4-set-up-kubectl-to-use-authentication-tokens-provided-by-aws-iam-authenticator-for-kubernetes) are installed and on your shell's PATH.
 
 ## Usage example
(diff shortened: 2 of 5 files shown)


max-rocket-internet:
Strange though. That parameter for the VPC module has always defaulted to `false`. And we haven't had a problem with the examples previously. I have previously run each of the examples from scratch and tested that nodes joined the cluster 🤔

cmrust:
I ran into this as well and found a similar recommendation from a different AWS page:

> In order for the private hosted zone to properly route traffic to your API server, your VPC must have enableDnsHostnames and enableDnsSupport set to true, and the DHCP options set for your VPC must include AmazonProvidedDNS in its domain name servers list.

- https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html

In my experience, the workers not joining was a result of setting `cluster_endpoint_private_access = true` in terraform-aws-eks without setting `enable_dns_hostnames = true` and `enable_dns_support   = true` in terraform-aws-vpc. The DHCP option referenced is already set as a default with the vpc module.

I'm wondering if @karolinepauls was enabling the private endpoint as well? Wish I had more time to draw up a test matrix and exercise all these different configs.

karolinepauls:
@cmrust I haven't changed `cluster_endpoint_private_access`, kept the defaults.

The below is what I had, and added `only enable_dns_hostnames = true` since.

The differences from the basic example is using the launch template and only a single security group (for SSH).

```
data "aws_availability_zones" "available" {}

resource "aws_security_group" "worker_management" {
  name_prefix = "all_worker_management"                   
  vpc_id      = module.vpc.vpc_id

  ingress {             
    from_port = 22                                        
    to_port   = 22
    protocol  = "tcp"
                         
    cidr_blocks = [                                       
      "10.0.0.0/8",                                     
      "172.16.0.0/12",                       
      "192.168.0.0/16",                   
    ]
  }
}

module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "2.6.0"

  name                 = var.cluster_name
  cidr                 = "10.0.0.0/16"
  azs                  = data.aws_availability_zones.available.names
  private_subnets      = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24"]
  public_subnets       = ["10.0.4.0/24", "10.0.5.0/24", "10.0.6.0/24"]
  enable_nat_gateway   = true
  single_nat_gateway   = true

  tags = {
    "kubernetes.io/cluster/${var.cluster_name}" = "shared"
  }

  public_subnet_tags = {
    "kubernetes.io/cluster/${var.cluster_name}" = "shared"
  }

  private_subnet_tags = {
    "kubernetes.io/cluster/${var.cluster_name}" = "shared"
    "kubernetes.io/role/internal-elb"           = "true"
  }
}

module "eks" {
  version = "5.0.0"

  cluster_name = var.cluster_name
  subnets      = module.vpc.private_subnets

  vpc_id = module.vpc.vpc_id

    {
      asg_desired_capacity = "2"
      asg_min_size         = "0"
      asg_max_size         = "5"
      key_name             = aws_key_pair.bastion-key.key_name
    },
  ]

  worker_additional_security_group_ids = [aws_security_group.worker_management.id]
  map_roles                            = []
  map_users                            = []
  map_accounts                         = []

  tags = {
    environment = "dev"
  }
}
```

@cmrust Can you chart the test matrix? I may automate/run the tests, depending on how much of a social failure I'll end up being this weekend :D

karolinepauls:
@max-rocket-internet done adding DNS resolution to other examples.

PR #458, opened 2019-07-31
Basic example: correct ELB tags

karolinepauls:
# PR o'clock

## Description

Added the "kubernetes.io/role/elb" tag to the public subnets in the
basic example.

Per documentation, ELB tag values are supposed to be "1". ELB tag values
being "true" are known not to work with aws-alb-ingress-controller.

https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
https://docs.aws.amazon.com/eks/latest/userguide/alb-ingress.html

### Checklist

- [ ] CI tests are passing


max-rocket-internet:
Travis job is failing. Can you run `terraform fmt` from root of project and commit changes?

karolinepauls:
@max-rocket-internet Sure, when I've got time, probs this evening.

This is kind of strange because I'm sure I ran `find . -name '*.tf' | xargs -n1 terrafrom fmt`. Now I see it's a failure from master.

I'm likely going to have time to create a PR fixing master this evening.

PR #463, opened 2019-08-05
Allow launch template spot instances without mixed policy

TarekAS:
# PR o'clock

## Description

Allow the creation of spot nodegroups using Launch Templates without resorting to Mixed Policy (due to certain incompatibilities with cluster autoscaler, for example). 

Edit: Switched to using `dynamic` block, instead of a separate `workers_launch_template_spot`.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -229,6 +229,14 @@ resource "aws_launch_template" "workers_launch_template" {
     )
   }
 
+  dynamic instance_market_options {
+    iterator = item
+    for_each = lookup(var.worker_groups_launch_template[count.index], "market_type", null) == null ? [] : list(lookup(var.worker_groups_launch_template[count.index], "market_type", null))
+    content {
+      market_type = item.value
+    }
+  }
+
   block_device_mappings {
     device_name = lookup(
       var.worker_groups_launch_template[count.index],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
- - Write your awesome addition here (by @you)
+ - Added `market_type` to `workers_launch_template.tf` allow the usage of spot nodegroups without mixed instances policy.
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


sppwf:
Hi @TarekAS 

Try next dynamic Block in the `workers_launch_template.tf` 
```hcl
  dynamic instance_market_options {
    iterator = item
    for_each = lookup(var.worker_groups_launch_template[count.index], "market_type", null) == null ? [] : list(lookup(var.worker_groups_launch_template[count.index], "market_type", null))
    content {
      market_type = item.value
    }
  }
````

Its quite ugly, however it will work. Conditional with for_each

Thanks

**Comment by TarekAS** [comment]
@sppwf I have already tried that as seen [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/409#issuecomment-518141906). The issue is that it was causing new changes upon each terraform run.

Making its own option was the only working solution I found.

**Comment by sppwf** [comment]
@TarekAS i have used your code as well, and it does not cause new changes upon each run. 
U am using terragrunt, however it should not make any difference. 
I have used this default value: 
```hcl
locals {
    use_spot_instances   = false
  }
}
```
`null` value for the list passed to for_each should not trigger dynamic block to create `instance_market_options` block for launch_templates with  `use_spot_instances`   set to `false`

Thanks

TarekAS:
@sppwf I see. Then it must have been something else from my side causing the change. I'll check again and confirm, if so I'll close this PR and create a new one with the better approach.

TarekAS:
Looks like the dynamic block worked without causing new changes. I have updated this PR with the proper changes.

workers_launch_template.tf
```diff
@@ -229,6 +229,14 @@ resource "aws_launch_template" "workers_launch_template" {
     )
   }
 
+  dynamic instance_market_options {
```
max-rocket-internet: Awesome! Finally seeing some great use of the new features in TF 0.12. Elegant 💅

max-rocket-internet:
Could we use the same method from this PR to remove `workers_launch_template_mixed.tf` and have the logic in `workers_launch_template.tf` without breaking anything?

Issue #462, opened 2019-08-05
How to handle worker upgrade automatically ?


barryib:
# I have issues

This module is great for deploying EKS clusters, but it has taken the decision to leave the worker upgrade out of its scope. This is ok for certain users, but for us, dealing manually with worker upgrade is a painful and repetitive task, mostly when you have a lot of workers.

This issue, is more like a discussion to decide if we want to implement this and how we should handle it.

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [x] support request
* [ ] kudos, thank you, warm fuzzy

## Are you able to fix this problem and submit a PR? Link here if you have already.

Yes, I'll be happy to submit PRs for this. But before that, I want to know what direction I (we) should take for this.

To handle this, I would like to use [autocaling group lifecycle hooks](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html) to drain nodes during scale in. I want to use a lambda function which will subscribe to `autoscaling:EC2_INSTANCE_TERMINATING` events and drain nodes before ASG terminates EC2 instances.

There is already a good proof of concept in aws-samples, called [amazon k8s node drainer](https://github.com/aws-samples/amazon-k8s-node-drainer).

By using ASG lifecycle hooks, we can achieve what @max-rocket-internet proposed https://github.com/terraform-aws-modules/terraform-aws-eks/issues/412#issuecomment-507270902.

And by using both hooks and cloud formation, we can tackle https://github.com/terraform-aws-modules/terraform-aws-eks/issues/333#issuecomment-480238089.

So, my point is NOT to handle all these with this module, but I think it should allow users to decide whether or not to scale in nodes after an LT change and let them handle node draining.

My questions here are :

1. How to handle node scale in ? Can I add a variable like `worker_recreate_asg_when_lt_changes` to let terraform recreate the ASG when the `aws_launch_template` changes ? This will only prefix the ASG name with LT name.
1. For users who wants to have more control on upgrades, are cloud formation templates welcome? This can add rolling upgrades (node by node) and can fit nicely with autocaling group lifecycle hooks.

## Any other relevant info

here are additional links for ASG lifecycle hooks :
- https://circleci.com/blog/graceful-shutdown-using-aws/
- https://devblog.xero.com/automating-ecs-cluster-upgrades-with-cloudformation-and-lambda-2b4f1bec575d
[...]

PR #466, opened 2019-08-07
add initial lifecycle hooks for autosacling groups

barryib:
# PR o'clock

## Description

This PR will allow user to create initial life cycle hooks for autoscaling groups.

Related to #462

### Checklist

- [x] CI tests are passing


max-rocket-internet:
Worth noting that these can exist outside of the module already: https://www.terraform.io/docs/providers/aws/r/autoscaling_lifecycle_hooks.html

barryib:
It can, but as mentioned in the link you are referring to ...

> If you need hooks to run on all instances, add them with initial_lifecycle_hook in aws_autoscaling_group, but take care to not duplicate those hooks with this resource.

In other word :
- if your lifecycle hook should run on node in terminating state (old nodes), so it can be handled outside of this module
- if your lifecycle hook should run on new nodes (or old one), so it has to be handled by this module

barryib:
@max-rocket-internet I updated the changelog.

Issue #101, opened 2018-08-22
aws_auth config fails to apply while getting started


tonyxiao:
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

Terraform apply fails with
```
* module.eks.null_resource.update_config_map_aws_auth: Error running command 'kubectl apply -f ./config-map-aws-auth_beam-eks.yaml --kubeconfig ./kubeconfig_beam-eks': exit status 1. Output: error: unable to recognize "./config-map-aws-auth_beam-eks.yaml": Unauthorized
```

## If this is a bug, how to reproduce? Please include a code sample if relevvant.

This is my configuration for the eks module.

I have a really basic vpc created via `terraform-aws-modules/vpc/aws`. 

```tf
module "eks" {
  source       = "terraform-aws-modules/eks/aws"
  cluster_name = "beam-eks"
  subnets      = "${module.vpc.public_subnets}"
  vpc_id       = "${module.vpc.vpc_id}"
}
```

## What's the expected behavior?

Apply succeeds

## Environment details

* Affected module version: "1.4.0"
* OS: MacOS 10.13.3 (17D47)
Terraform v0.11.8
+ provider.aws v1.33.0
+ provider.http v1.0.1
+ provider.local v1.1.0
+ provider.null v1.0.0
+ provider.template v1.0.0

tonyxiao:
Upon further investigation, it appears that this is only an issue when I have `manage_aws_auth` set to `true`. It turns out that my aws `provider` uses `assume_role`, and this module uses local exec loses the assumed role and thus isn't able to execute the operation.

max-rocket-internet:
Hi @tonyxiao 
We also use assumed roles, like this:

```
$ aws sts get-caller-identity
{
    "UserId": "AROAXXXXXXXXXXX:max.xxxxxx@xxxxxx.com",
    "Account": "99999999999",
    "Arn": "arn:aws:sts::99999999999:assumed-role/sso-xxxxxxx/max.xxxxxx@xxxxxx.com"
}
```

To get it to work with the kubeconfig from this terraform module I did this:

1. Ensure your assumed role details are set in a profile. e.g. in `~/.aws/config`:

```
[profile profile1]
region = eu-west-1
google_config.ask_role = True
google_config.duration = 43200
[...]

PR #467, opened 2019-08-11
Add AWS profile variable to examples

RSO:
# PR o'clock

## Description

When trying to run the examples I ran into issues creating the
`aws_auth_configmap.yaml` file:

```
module.eks.null_resource.update_config_map_aws_auth[0] (local-exec): error: unable to recognize "aws_auth_configmap.yaml": Unauthorized
```

In https://github.com/terraform-aws-modules/terraform-aws-eks/issues/101#issuecomment-415323404
@max-rocket-internet explains that the `AWS_PROFILE` should be set in
`kubeconfig_aws_authenticator_env_variables`. Doing that solved my
issues locally, so I decided to add this bit to the examples so that it
is clear for others as well.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/examples/basic/main.tf
+++ b/examples/basic/main.tf
@@ -5,6 +5,7 @@ terraform {
 provider "aws" {
   version = ">= 2.11"
   region  = var.region
+  profile = var.profile
 }
 
 provider "random" {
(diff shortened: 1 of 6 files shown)


max-rocket-internet:
Thanks for the PR but I'm not sure we need this?

The examples are supposed to be as simple as possible. The setting for `AWS_PROFILE` in relation to kubeconfig is mentioned literally in the main README. You can't get much more obvious than that 😃 And using an AWS profile for the provider is a [terraform feature](https://www.terraform.io/docs/providers/aws/index.html#profile), not really related to this module.

PR #496, opened 2019-09-05
Allow to set user defined workers role name

ivanich:
# PR o'clock

## Description
Allow to set workers role name. Use case is a cross account policies where predictable role names are required.

### Checklist

- [ ] CI tests are passing


max-rocket-internet:
Like it.

Do we need the boolean variable also? Couldn't we just have the role name that defaults to blank?

ivanich:
> Like it.
> 
> Do we need the boolean variable also? Couldn't we just have the role name that defaults to blank?

Makes sense, PR updated. Thanks

lisfo4ka:
Hi folks, @ivanich @max-rocket-internet 

Could you clarify, please, why it's so different from the cluster IAM management approach?

I'm confused a little about realization at the moment.
Am I understand correctly, that if we set `manage_worker_iam_resources=true `and define `workers_role_name` with the required name we'll get IAM role for all worker nodes and there is no way to redefine it for separate `worker_groups_launch_template` block?
I've assumed that we can override `workers_role_name` on the worker group level, but I can't find the way.
I'll be really grateful if somebody can assist with this question.

PR #503, opened 2019-09-11
Improve CI pipeline

nauxliu:


max-rocket-internet:
@Vlaaaaaaad you made and deleted an interesting comment. Feel free to make a PR if you think your work can improve our use of Github actions here 😃

Vlaaaaaaad:
> @Vlaaaaaaad you made and deleted an interesting comment. Feel free to make a PR if you think your work can improve our use of Github actions here 😃

Yeah, it felt too spammy 😅 I'm still hoping to have it ready this week and I'll open-source it and make a PR.

barryib:
@max-rocket-internet @nauxliu 
> Do you think we can add a check that ensures people have updated the variables and outputs in the README like the pre-commit hook does? Maybe something with terraform-docs?

I personally run pre-commit hooks (`pre-commit run -a`) to achieve that. pre-commit hooks will fail with an exit code different from 0 if there is a change. I don't know if the best way to do that, but It works as expected.

RothAndrew:
+1 for running `pre-commit run -a` as a pipeline step.

nauxliu:
@max-rocket-internet Terraform docs check added. Ready for review.

variables.tf
```diff
@@ -207,7 +207,7 @@ variable "worker_create_security_group" {
 variable "permissions_boundary" {
   description = "If provided, all IAM roles will be created with this permissions boundary attached."
   type        = string
-  default     = ""
+  default     = null
```
barryib: This won't break `aws_iam_role` ? Because, it's a null string (`"null"`) and not `null` value.

nauxliu: it's not null string, 0.12 introduces the null value for conditionally omitting a value.

nauxliu: https://www.hashicorp.com/blog/terraform-0-12-conditional-operator-improvements

barryib: So it's probably terraform-doc is writing it as string in docs

> | permissions\_boundary | If provided, all IAM roles will be created with this permissions boundary attached. | string | `"null"` | no |

nauxliu: There is no native support for Terraform 0.12 in terraform-docs. See https://github.com/segmentio/terraform-docs/issues/62

barryib:
@nauxliu Thank you for work. PR is merged.

variables.tf
barryib: I know. I was thinking about the [awk version](https://github.com/antonbabenko/pre-commit-terraform/blob/master/terraform_docs.sh) in the pre-commit hooks.

barryib: Maybe here https://github.com/antonbabenko/pre-commit-terraform/blob/master/terraform_docs.sh#L138

barryib: Never mind, all types are "quoted" by terraform-docs.sh. I using this for a long time now, but I didn't notice that until today :s

PR #506, opened 2019-09-13
Add URL for OIDC issuer to allow IAM roles for Pods

russwhelan:
# PR o'clock

## Description

AWS now support a native solution for grant Pods IAM roles with OIDC tokens. More details can be found here: https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

This PR adds a new output for the ODIC Issuer URL which is needed to create the required IAM resources to allow Pods to assume roles. 

### Checklist

- [x] CI tests are passing


max-rocket-internet:
We would also need a [iam_openid_connect_provider](https://www.terraform.io/docs/providers/aws/r/iam_openid_connect_provider.html).

I as far as I understand, this resource is directly associated with the cluster. i.e. it's per cluster and not per region or per AWS account.

So, do you think it's worth including this resource in this module?

Reference: https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

russwhelan:
I had a similar thought this morning but decided against it as by including it we are making the assumption that the user of the module will want to use the new OIDC method over say kube2iam. Plus a user at this point can still create 1.11 or 1.12 cluster which don't support the OIDC method so a module update might break things for those users as the iam_openid_connect_provider resource would reference an output that may not exist.

mKeRix:
There is an example configuration for the new IAM functionality [here](https://github.com/terraform-providers/terraform-provider-aws/issues/9995#issuecomment-528313360). Maybe it would make sense to only create the additional resources based on a boolean input?

In that case I would add two more outputs: the ID of the openid_connect_provider and the aws_iam_policy_document data source for the needed assume role policy. This would allow roles to be created and mapped quickly together with this module.

dpiddockcmp:
We should not be creating an assume role policy in this module with a hard coded source serviceaccount. You need an assume role policy for each source k8s service account. It's how you're going to restrict which pod can assume which IAM role. I don't know if it's in scope for this module to be creating the `iam_policy_document`s or `iam_role`s given a list of serviceaccounts.

See the [AWS blog](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) for more

mKeRix:
> We should not be creating an assume role policy in this module with a hard coded source serviceaccount. You need an assume role policy for each source k8s service account. It's how you're going to restrict which pod can assume which IAM role. I don't know if it's in scope for this module to be creating the `iam_policy_document`s or `iam_role`s given a list of serviceaccounts.

Yep, you're right - overlooked that. In that case I think it may make more sense to put this into a separate module. Either way I would still include the connect provider here with an optional flag.

russwhelan:
Thanks everyone for the input. :)

My view here is that given as discussed a large part of the IAM parts should sit outside the module I am of the view that there is then no benefit in creating the openid_connect_provider for the module user as it only saves them one resource is what would be a load more anyway.

So to summarise I think the initial PR stands, as it provides the missing piece that is required to create all the roles etc that would be need in a complex environment. Make sense?

mKeRix:
Makes sense to me - either this would at least be a good first iteration for this feature. Would love to see this merged and released soon :)

abdennour:
in which version this feature will be released ? 

BTW, [Terraform AWS provider has already released it since version 2.28.0](https://github.com/terraform-providers/terraform-provider-aws/issues/9995#issuecomment-530876028)

dpiddockcmp:
This was included in v6.0.0.

v6.0.1 included a small fix for backwards compatibility with old clusters that don't support the feature.

PR #508, opened 2019-09-13
Fix launch template market option expansion

to266:
# PR o'clock

## Description

This fixes the `market_type` passing when generating the `workers_launch_template`. The previous version would not populate the `instance_market_options.market_type` even if the corresponding value was passed.

### Checklist

- [ ] CI tests are passing


max-rocket-internet:
Please add your change to the change log.

to266:
Sorry, added.

PR #515, opened 2019-09-17
Gpu workers support

rvoitenko:
# PR o'clock

## Description

Current implementation doesn't allow to use Amazon EKS-Optimized AMI with GPU Support(https://docs.aws.amazon.com/eks/latest/userguide/gpu-ami.html), because of hardcoded AMI filter prefix "amazon-eks-node".
This patch allow to define different AMI filter prefix, i.e. "amazon-eks-gpu-node", but default is still set to "amazon-eks-node"

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Write your awesome addition here (by @you)
+ - Added support for different workers AMI's, i.e. with GPU support (by @rvoitenko)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


max-rocket-internet:
Does the GPU AMI use the same init script? I asked about this a long while ago when trying to implement something for this module: https://github.com/awslabs/amazon-eks-ami/issues/45

rvoitenko:
> Does the GPU AMI use the same init script? I asked about this a long while ago when trying to implement something for this module: [awslabs/amazon-eks-ami#45](https://github.com/awslabs/amazon-eks-ami/issues/45)

yes, it's using same script. Thank you for merge!

Issue #527, opened 2019-09-25
Again - Provider produced inconsistent final plan


timota:
# I have issues
With a new module version 6.0 we still have an issue with:
`Error: Provider produced inconsistent final plan`

This issue previously has been address in: [https://github.com/terraform-providers/terraform-provider-aws/issues/9882](url) and [https://github.com/terraform-aws-modules/terraform-aws-eks/issues/481](url) but in version 6.0 of Module it back again

## What is the current behavior?
```
Error: Provider produced inconsistent final plan

When expanding the plan for
module.eks.aws_autoscaling_group.workers_launch_template[0] to include new
values learned so far during apply, provider "aws" produced an invalid new
value for .initial_lifecycle_hook: planned set element
cty.ObjectVal(map[string]cty.Value{"default_result":cty.UnknownVal(cty.String),
"heartbeat_timeout":cty.UnknownVal(cty.Number),
"lifecycle_transition":cty.UnknownVal(cty.String),
"name":cty.UnknownVal(cty.String),
"notification_metadata":cty.UnknownVal(cty.String),
"notification_target_arn":cty.UnknownVal(cty.String),
"role_arn":cty.UnknownVal(cty.String)}) does not correlate with any element in
actual.
```

## If this is a bug, how to reproduce? Please include a code sample if relevant.
```
  worker_groups_launch_template = [
    {
      instance_type        = "c5.large"
      key_name             = aws_key_pair.worker_management.key_name
      subnets              = var.subnet_ids
      asg_desired_capacity = var.minimal_cluster_size
      asg_min_size         = var.minimal_cluster_size
      asg_max_size         = "5"
      pre_userdata = <<EOFT
     }
  ]

module "eks" {
  source                               = "terraform-aws-modules/eks/aws"
  cluster_name = local.cluster_name
  cluster_version = var.cluster_version
  subnets         = concat(var.subnet_ids, var.underlay_subnet_ids)
  tags            = local.tags
  vpc_id          = var.vpc_id
  worker_groups_launch_template        = local.worker_groups_launch_template
  worker_additional_security_group_ids = [aws_security_group.all_worker_mgmt.id]
  map_roles                            = var.map_roles
  map_users                            = var.map_users
  map_accounts                         = var.map_accounts
  config_output_path                   = "./generated/"
  cluster_endpoint_private_access      = var.private
  permissions_boundary                 = var.private ? <ARN> : ""

}
```

## What's the expected behavior?
[...]

PR #532, opened 2019-09-26
Add option to enable lifecycle hooks creation

barryib:
# PR o'clock

## Description

Related to https://github.com/terraform-aws-modules/terraform-aws-eks/issues/527

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
- - Write your awesome change here (by @you)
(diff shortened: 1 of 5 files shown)


adilsonmenechini:
Estou tendo o mesmo problema 

2020/10/16 11:52:44 [WARN] Provider "registry.terraform.io/hashicorp/aws" produced an unexpected new value for module.eks.aws_autoscaling_group.workers_launch_template[0], but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .launch_configuration: was null, but now cty.StringVal("")
      - .placement_group: was null, but now cty.StringVal("")

PR #539, opened 2019-09-27
Make "dangerous" policy attachments optional

dpiddockcmp:
# PR o'clock

## Description

The IAM policies for the node autoscaler and AWS CNI have write permissions and are assigned to the nodes by default. These could potentially be dangerous to the stability of the cluster if accessed by a rogue pod as they are the default role received.

I'm currently looking at the new OIDC IAM integration. The [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/restrict-ec2-credential-access.html) has an example for using iptables to block access to the metadata service but this broke some pods like the alb-ingress-controller. Not a great situation. I could put pods that need access to the metadata service on the host network but I don't fancy that solution. The only other option is to make the default node role as harmless as possible.

Also made creation of the autoscaling policy optional for users who don't want it at all.

Greater flexibility in the module can't harm anybody.

Defaults set to current behaviour for backwards compatibility.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
  - Add option to enable lifecycle hooks creation (by @barryib)
  - Remove helm chart value `sslCertPath` described in `docs/autoscaling.md` (by @wi1dcard)
+ - Attaching of IAM policies for autoscaler and CNI to the worker nodes now optional (by @dpiddockcmp)
 
 # History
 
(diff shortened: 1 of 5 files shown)


max-rocket-internet:
> These could potentially be dangerous to the stability of the cluster if accessed by a rogue pod as they are the default role received.

Agreed.

> Greater flexibility in the module can't harm anybody.

Agreed.

> Defaults set to current behaviour for backwards compatibility.

Nice.

> The AWS documentation has an example for using iptables to block access to the metadata service

This or use KAIM or kube2iam.

Ultimately I've been waiting for the [EKS service account feature](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) to be completed since EKS was created. And since the beginning, just piling all the policies on the worker EC2 role, which is very bad practice. And now the this feature is available, we still have to wait for all the SDKs to support it and then update all apps to use the new versions of the SDKs. Quite time consuming 😃

PR #541, opened 2019-09-29
WIP: Docs fixer in GitHub Actions

Vlaaaaaaad:
# PR o'clock

Creating this PR to start a discussion on wether this is desired or not. The user experience for this is not perfect.

## Description

There are checks for formatting and docs, but no easy way to fix it. To get the docs up to date/ fix formatting contributors have to setup `pre-commit` properly.

This PR adds a GitHub Action that runs `pre-commit` and commits the result back.

Example workflow:
- PR is open, all checks are green
- commits are added, docs are now out of date
- `/fix` comment on the PR Review( not on the _Conversation_ as GitHub Actions has... limitation about events and what data is in there -- like the branch name)
- new commit is added to the branch fixing the docs

There is a downside to this though: by default when using `secrets.GITHUB_TOKEN` no actions will run on commits made by GitHub Actions. This is bad especially because checks are required for merging.
An ugly but working fix is to use a _Personal Access Token_ as the secret instead of the default `GITHUB_TOKEN`. What I see in most GitHub orgs is a bot user which is used for this. 

An example with `secrets.GITHUB_TOKEN` [can be seen here](https://github.com/Castravete/repo-for-fork-testing/pull/3).
An example with `secrets.BOT_PERSONAL_ACCESS_TOKEN` [can be seen here](https://github.com/Castravete/repo-for-fork-testing/pull/6).

~~⚠️ Weirdly enough, the `terraform_docs` had different results when the check was running on macOS vs when running on Linux. I have no idea where to even start to debug this. ⚠️~~

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/.github/workflows/autofixers.yml
+++ b/.github/workflows/autofixers.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the source for the PR
-        run: git clone --branch=$GITHUB_BRANCH" "https://github.com/${GITHUB_REPO}" . && git reset --hard "$GITHUB_REV"
+        run: git clone --branch="$GITHUB_BRANCH" "https://github.com/${GITHUB_REPO}" . && git reset --hard "$GITHUB_REV"
         env:
           GITHUB_REV: ${{ github.event.pull_request.head.sha }}
           GITHUB_BRANCH: ${{ github.event.pull_request.head.ref }}
--- a/.github/workflows/autofixers.yml
+++ b/.github/workflows/autofixers.yml
@@ -1,0 +1,46 @@
+name: Autofixers
+
+on: [pull_request_review]
+
+jobs:
+  precommit:
+    name: Pre-commit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the source for the PR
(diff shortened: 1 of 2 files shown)


Vlaaaaaaad:
🤦‍♂ this should've been a _Draft PR_ as I wanted to discuss what's in here first. Can't make it a draft now unfortunately.

RothAndrew:
IMO, a better pattern would be to fail the build if the user hasn't done the required pre-commit stuff.

Vlaaaaaaad:
The build does already fail if the pre-commit checks don't pass. The proposed action would help fix those issues without forcing a contributor to setup `pre-commit`.

barryib:
I'm agree with @RothAndrew. I think the setup process is simple enough. To install pre-commit hooks, it just matter of `brew install pre-coomit` or `pip install pre-commit`. And this can be added in the [CONTRIBUTING guide](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/CONTRIBUTING.md). Running pre-commit can be done even manually (if you don't want to `pre-commit install`) with `pre-commit run -a` before pushing your contribution.

This work is great and it would have a lot of benefit on a complicated workflow or setup.

Vlaaaaaaad:
Well, based on feedback I am closing the PR. It makes sense considering the workflow here is rather short.

PR #544, opened 2019-10-01
Fix accidental background filling of config files

james-portman-contino:
# PR o'clock

## Description

It looks like someone has accidentally backgrounded the calls to fill out kube_config.yaml and aws_auth_configmap.yaml
Using a single ampersand will run the echo commands in the background and immediately run kubectl (at the same time essentially) so the config files can/will often be empty
I assume, and it makes sense to use double ampersand so that each command runs only if the previous one worked.

Also might be able to remove the 10x loop depending on why that was added.
If it was added because the code randomly worked/not (not surprised) then I have probably fixed that and removed the need for the loop

### Checklist

- [x] CI tests are passing

### DIFF
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -13,8 +13,8 @@ resource "null_resource" "update_config_map_aws_auth" {
 
     command = <<EOS
 for i in `seq 1 10`; do \
-echo "${null_resource.update_config_map_aws_auth[0].triggers.kube_config_map_rendered}" > kube_config.yaml & \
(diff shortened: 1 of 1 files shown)


max-rocket-internet:
Sorry @james-portman-contino I missed this PR 😐

It was addressed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/592 anyway.

james-portman-contino:
OK brill, thanks

PR #545, opened 2019-10-02
remove the launch template mixed variable from examples

harshal-shah:
# PR o'clock

## Description
Removing the `worker_groups_launch_template_mixed` variable from examples.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/docs/spot-instances.md
+++ b/docs/spot-instances.md
@@ -73,6 +73,7 @@ Launch Template support is a recent addition to both AWS and this module. It mig
     }
   ]
 
+/* The variable worker_groups_launch_template_mixed is not suported from v6.0.0 onwards */
   worker_groups_launch_template_mixed = [
     {
       name                     = "spot-1"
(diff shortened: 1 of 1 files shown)


docs/spot-instances.md
```diff
@@ -73,6 +73,7 @@ Launch Template support is a recent addition to both AWS and this module. It mig
     }
   ]
 
+/* The variable worker_groups_launch_template_mixed is not suported from v6.0.0 onwards */
```
max-rocket-internet: I think just remove it.

We should make the examples reflect the current version only. If we have to write notes about backwards and forwards compatibility this is too much work and too complicated 😃

harshal-shah: Sounds good, thanks for the review!

max-rocket-internet:
Resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/546

PR #549, opened 2019-10-04
Allow `config_output_path` to specify the full path

joshuaspence:
# PR o'clock

## Description

Currently the `kubectl` configuration file is always written to a file named `kubeconfig_${var.cluster_name}`. I want to write to `~/.kube/config` instead, so allow `var.config_output_path` to refer to a file as well as a directory.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
- - Write your awesome change here (by @you)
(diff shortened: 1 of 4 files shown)


joshuaspence:
The module doesn't write this file for this own usage, as otherwise it wouldn't work with `write_kubeconfig = false`. Of course I can use the `local_file` resource myself, that's what I am doing at the moment.

barryib:
@max-rocket-internet any thoughts ?

Issue #480, opened 2019-08-25
data aws_ami "eks_worker" is failing in ap-east-1


vranystepan:
# I have issues
We are not able to create EKS cluster in `ap-east-1` due to issues with AMI names.

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [x] support request
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

With any ami we are getting

```
Error: Your query returned no results. Please change your search criteria and try again.

  on .terraform/modules/<redacted>/terraform-aws-modules-terraform-aws-eks-6ea3582/data.tf line 19, in data "aws_ami" "eks_worker":
  19: data "aws_ami" "eks_worker" {
```

We suppose that EKS module is [always](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/local.tf#L19) trying to find images (before `workers_group_defaults_defaults` is being merged with `var.workers_group_defaults`) with following [pattern](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/data.tf#L22) but apparently there are no such images in the HK region. 

```
amazon-eks-node-${var.cluster_version}-${var.worker_ami_name_filter}
```

I'm not able to receive any results even with CLI:

```
aws ec2 describe-images \                                                                       
    --filters \
        Name=owner-id,Values=602401143452 \
        Name=virtualization-type,Values=hvm \
        Name=root-device-type,Values=ebs \
        Name=is-public,Values=true \
        Name=name,Values=amazon-eks-node-\* \
    --output text \
    --query 'max_by(Images[], &CreationDate).ImageId' \
--profile=<redacted> \
--region=ap-east-1
None
```

## Environment details

* Affected module version: v5.1.0
* OS: any
* Terraform version: 0.12.7

## Any other relevant info
Perhaps there is different owner in HK region? https://github.com/awslabs/amazon-eks-ami/pull/305/files/fd9cf4458dffa1e9c01a3a42fca1cc7690abc504

max-rocket-internet:
> Perhaps there is different owner in HK region?

Looks like it. Strange.

vranystepan:
Internally we've applied temp. workaround in `data.tf`:

```hcl
...
owners = ["602401143452", "800184023465"]
...
```

but this is not the solution as 800184023465 does not contain images with relevant names :)

bsakweson:
I seem to be having the same issue in us-east-1. Here is what I am seeing?

```
[...]

PR #551, opened 2019-10-07
Allow for specifying a custom AMI for the worker nodes.

### LINKED ISSUE #540: Support for specifying worker AMIs other than the defaults
# I have issues

## I'm submitting a...

* [ ] bug report
* [ x ] feature request
* [ ~ ] kudos, thank you, warm fuzzy <<<<<<*Thank You!*

At the moment, the ami can be customised using the `worker_ami_name_filter_prefix` and `worker_ami_name_filter` vars. However the template is limited to the `owner = 602401143452`, and with a name format of `prefix-version-name`:
```
data "aws_ami" "eks_worker" {
  filter {
    name   = "name"
    values = ["${var.worker_ami_name_filter_prefix}-${var.cluster_version}-${var.worker_ami_name_filter}"]
  }

  most_recent = true

  # Owner ID of AWS EKS team
  owners = ["602401143452"]
}
```

I'd like to experiment with the ubuntu EKS images (https://cloud-images.ubuntu.com/aws-eks/) and so need to be able to override this. Happy to raise a PR, but not sure on policy on backwards compatibility. Would the easiest thing be to have a `var.worker_ami_id` that if specified overrides the existing options?

The ubuntu images are owned by `099720109477`, and have the format `ubuntu-eks/k8s_1.13/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-20190823.1`...

bmcstdio:
# PR o'clock

## Description

In some environments one may be required to customise the AMI used by the worker nodes.
This PR allows for specifying a custom AMI to use for the worker nodes in a backwards-compatible way (and I think it closes #540).

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- **Breaking:** Allow for specifying a custom AMI for the worker nodes. (by @bmcstdio)
 - Write your awesome addition here (by @you)
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -142,8 +142,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | tags | A map of tags to add to all resources. | map(string) | `{}` | no |
 | vpc\_id | VPC where the cluster and workers will be deployed. | string | n/a | yes |
 | worker\_additional\_security\_group\_ids | A list of additional security group ids to attach to worker instances | list(string) | `[]` | no |
-| worker\_ami\_name\_filter | Additional name filter for AWS EKS worker AMI. Default behaviour will get latest for the cluster_version but could be set to a release from amazon-eks-ami, e.g. "v20190220" | string | `"v*"` | no |
-| worker\_ami\_name\_filter\_prefix | Name prefix filter for AWS EKS worker AMI. Default behaviour will get regular EKS-Optimized AMI but could be set to a EKS-Optimized AMI with GPU Support, e.g. "amazon-eks-gpu-node", or custom AMI | string | `"amazon-eks-node"` | no |
+| worker\_ami\_name\_filter | Name filter for AWS EKS worker AMI. If not provided, the latest official AMI for the specified 'cluster_version' is used. | string | `""` | no |
+| worker\_ami\_owner\_id | The ID of the owner for the AMI to use for the AWS EKS workers. Valid values are an AWS account ID, 'self' (the current account), or an AWS owner alias (e.g. 'amazon', 'aws-marketplace', 'microsoft'). | string | `"602401143452"` | no |
 | worker\_create\_initial\_lifecycle\_hooks | Whether to create initial lifecycle hooks provided in worker groups. | bool | `"false"` | no |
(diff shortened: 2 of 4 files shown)


max-rocket-internet:
Do you think it's perhaps too complicated having 4 variables just for filtering the AMI? As opposed to users just define the data resource outside the module?

bmcstdio:
I think so, yes, but I guess it's pretty much the only way to keep backwards-compatibility (isn't it?). If we were to break that, I'd prefer repurposing `worker_ami_name_filter` and getting rid of the `worker_ami_name_filter_prefix` one, leaving only two as we currently have. I think that defining the data source outside of the module makes things less user-friendly TBH.

bmcstdio:
@max-rocket-internet I've rebased against `master` and fixed conflicts 🙂 Please let me know what additional changes you'd like me to make.

max-rocket-internet:
Hey @bmcstdio! Thanks for the PR!

I think with this PR it's simply too complicated now. 4 different variables, all with quite complicated descriptions, just to set 2 attributes on the `data.aws_ami.eks_worker` resource.

I know it might seem like a minor gripe but we are always fighting to keep complexity low with this module otherwise it's kind of like "death by a thousand cuts" if you know what I mean 🙂

> pretty much the only way to keep backwards-compatibility

Don't worry about that. If it's a breaking change, just make sure you note it clearly in the changelog. When we next do a release we'll have to make sure it's a major one.

So, how about we try this:
1. Keep the `worker_ami_owner_id` variable as this will solve https://github.com/terraform-aws-modules/terraform-aws-eks/issues/480
2. Make a single variable called `worker_ami_name_filter` that defaults to `""`
3. Set the filter to default to a local that is `"amazon-eks-node-${var.cluster_version}-v*"` unless `worker_ami_name_filter` is set, in which case use that.

How does that sound?

bmcstdio:
@max-rocket-internet sure thing, I've updated the code to reflect that. 🙂

apatap:
How about removing/deprecating _worker_ami_name_filter_ and _worker_ami_name_filter_prefix_ by introducing _worker_node_ami_?

The documentation could be extended by an example how to write a filter. This solution would reduce the number of variables and give more flexibility.

Here an completely **untested DRAFT** of the needed code changes. Maybe the naming should be aligned with _workers_default_ami_id_ in output.tf.

#### Code changes:
local.tf

```
locals {
  ...
      ...
        ami_id = var.worker_node_ami = "" ?  data.aws_ami.eks_worker.id  : var.worker_node_ami.id
      ...
  }
  ...
}
```

variables.tf

```
variable "worker_node_ami" {
  description = "AMI for the default eks worker nodes. If none is provided, the latest version of Amazon EKS-Optimized Linux AMI is used."
  type        = any
  default     = ""
}
```


#### Code snippet for documentation:

```
data "aws_ami" "my_custom_eks_ami" {
  owners           = ["self"]
  most_recent      = true
  name_regex       = "^mycustomeksworkernodeami"
}

module "my-cluster" {
  source       = "terraform-aws-modules/eks/aws"
  ...
  worker_node_ami= data.aws_ami_id.my_custom_eks_ami.id
  ...
}
```

max-rocket-internet:
> How about removing/deprecating worker_ami_name_filter and worker_ami_name_filter_prefix by introducing worker_node_ami

Also possible. Basically you just mean the user must pass in an AMI ID, always? I think most users, myself included, are only interested in the using the official AMIs. But maybe I'm wrong?

PR #556, opened 2019-10-10
Remove region from vars mapping for kubeconfig

miguelaferreira:
# PR o'clock

## Description

The template file data source was injecting a region variable that is not defined in the template. Once removed the variable the aws region data source was not used anywhere else, so it is removed as well.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -2,9 +2,6 @@ locals {
   worker_ami_name_filter = var.worker_ami_name_filter != "" ? var.worker_ami_name_filter : "amazon-eks-node-${var.cluster_version}-v*"
 }
 
-data "aws_region" "current" {
-}
-
 data "aws_iam_policy_document" "workers_assume_role_policy" {
   statement {
     sid = "EKSWorkerAssumeRole"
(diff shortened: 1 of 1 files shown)


max-rocket-internet:
Please observe the checklist and rebase 🙂

miguelaferreira:
I'm going to rebase, and I think it's safe to say that, given the change, fmt and validate will be happy, but I'll run it just to provide evidence. Is CI running the examples or am I expected to run them myself on my own account?

miguelaferreira:
Rebased.

To that checklist:

- `terraform fmt` on the module produces no change
- `terraform init && terraform validate` on all examples produces no errors

Since this PR did not change anything in the module's interface, or even logic (this PR removes dead code), the README does not need an update, and I'm not sure about the CHANGELOG entry

max-rocket-internet:
hmmmm now I'm thinking someone might be using the region in these parts? https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/templates/kubeconfig.tpl#L26-L28

>  I'm not sure about the CHANGELOG entry

Always a good idea 😃

max-rocket-internet:
> I don't think I follow

Ah you are right, I thought users could provide their own template for the  `data.template_file.kubeconfig` resource. But this isn't possible. My mistake!

PR #572, opened 2019-10-25
Attach EBS volumes to worker nodes #233

szpuni:
# PR o'clock

## Description

This PR updates functionality of ASG to allow external Launch Configuration to be used rather than one provided by module. 
This was requested in #233 

### Checklist

- [ ] CI tests are passing


szpuni:
I think there might be better way of doing this with help of Dynamic block.
```
variable "ebs" {
  default = ["/dev/sdf"]
}

  dynamic ebs_block_device {
    for_each = var.ebs
    content {
      device_name = ebs_block_device.value
      volume_size = "8"
      delete_on_termination = true
      volume_type = "gp2"
    }
```
I was goofing around and found that if you run above code with single instance you will get Root and EBS volume. If you give empty list to above then plan will show that EBS will be created but there is only Root device is created in AWS.

This might be a good solution to this problem with small enough changes to code.

Ideas?

max-rocket-internet:
> I think there might be better way of doing this with help of Dynamic block

Agreed.

PR #573, opened 2019-10-28
Update PULL_REQUEST_TEMPLATE.md

max-rocket-internet:
- I don't think we need the notes about "terraform fmt and terraform validate" as these are covered in the CI tests anyway, right?
- I want to emphasise adding to changelog as many people are skipping this because they think their change is minor but I think every change should be added.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,6 @@ Please explain the changes you made here and link to any relevant issues.
 
 ### Checklist
 
-- [ ] `terraform fmt` and `terraform validate` both work from the root and `examples/*` directories
+- [ ] Change added to CHANGELOG.md. All changes must be added and breaking changes and highlighted
 - [ ] CI tests are passing
-- [ ] I've added my change to CHANGELOG.md and highlighted any breaking changes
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,7 +6,6 @@ Please explain the changes you made here and link to any relevant issues.
 
 ### Checklist
 
-- [ ] `terraform fmt` and `terraform validate` both work from the root and `examples/*` directories
+- [ ] Change added to CHANGELOG.md. All changes must be added and breaking changes and highlighted
 - [ ] CI tests are passing
-- [ ] I've added my change to CHANGELOG.md and highlighted any breaking changes
 - [ ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation
```
barryib: doc generation is also tested by the CI https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/workflows/lint.yml#L22-L30. Maybe we should delete this line also ?

max-rocket-internet: It's tested, yes, but I think a user still needs to be reminded to how install doc-generation stuff as it's more complicated than fmt/validate.

PR #570, opened 2019-10-24
Increase minimum terraform version to include yamlencode

toadjaune:
https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#0122-june-12-2019

# PR o'clock

## Description

Bump minimum terraform version in order to avoid the following errors :
```
Error: Call to unknown function

  on .terraform/modules/my-cluster/terraform-aws-modules-terraform-aws-eks-1be1a02/aws_auth.tf line 82, in data "template_file" "config_map_aws_auth":
  82:     map_users    = yamlencode(var.map_users),

There is no function named "yamlencode".


Error: Call to unknown function

  on .terraform/modules/my-cluster/terraform-aws-modules-terraform-aws-eks-1be1a02/aws_auth.tf line 83, in data "template_file" "config_map_aws_auth":
  83:     map_roles    = yamlencode(var.map_roles),

There is no function named "yamlencode".


Error: Call to unknown function

  on .terraform/modules/my-cluster/terraform-aws-modules-terraform-aws-eks-1be1a02/aws_auth.tf line 84, in data "template_file" "config_map_aws_auth":
  84:     map_accounts = yamlencode(var.map_accounts)

There is no function named "yamlencode".
```

### Checklist

- [ ] `terraform fmt` and `terraform validate` both work from the root and `examples/*` directories
- [x] CI tests are passing
- [ ] I've added my change to CHANGELOG.md and highlighted any breaking changes
- [ ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #575, opened 2019-10-29
Add Action check against minimum versions

dpiddockcmp:
# PR o'clock

## Description

Add a Github Action check to try to validation the minimum versions listed. This won't catch everything but it does catch the two recent cases:
- `yamlencode()` use from terraform 0.12.2 #570 
- tags on `aws_eks_cluster` in aws 2.31.0 #550 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@master
       - name: TFLint
         uses: docker://wata727/tflint
-  
+
   fmt:
     name: Code Format
     runs-on: ubuntu-latest
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - run: terraform fmt --recursive -check=true
-  
+
   docs:
     name: Docs
     runs-on: macOS-latest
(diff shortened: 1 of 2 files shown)


max-rocket-internet:
Good idea @dpiddockcmp 

Maybe we should just set all `hashicorp/terraform:latest` to `hashicorp/terraform:0.12.2`?

dpiddockcmp:
Hi Max.

Do you mean instead of this check? As this check is also validating the providers' minimum versions.

max-rocket-internet:
I meant we are already running the `fmt` job with `hashicorp/terraform:latest`:  https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/workflows/lint.yml#L17

And `validate`  with `hashicorp/terraform:latest`:  https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/workflows/lint.yml#L36

So I was thinking we just switch all of these to `0.12.2` but now I think about it, you're saying we should validate for current latest AND our current minimum. Is that correct?

dpiddockcmp:
Personally, I hate the use of `:latest` as it can easily add unpredictable breakages to CI runs, like every time the terraform minor version increases. No doubt 1.0 or 0.13 will cause issues again.

I think it would be a good idea to set static version numbers for the other tests. This may not necessarily be the minimum of the module:
- `terraform fmt` from 0.12.2 incorrectly complains about `aws_auth.tf` and `data.tf` but 0.12.6 and later do not.
- `examples` include other modules which may have their own terraform restrictions. Currently only VPC is used which does not specify versions.

max-rocket-internet:
> Personally, I hate the use of :latest as it can easily add unpredictable breakages 

I agree. It's especially bad in docker 😅

> I think it would be a good idea to set static version numbers for the other tests

I get the concept but the problem is that people will just `brew install terraform`, get the latest version and then use this module, and perhaps have an issue. But we need to know if the latest version has a problem with this module.

So I think we merge this and see how it goes unless someone has a better idea.

max-rocket-internet:
> Personally, I hate the use of :latest as it can easily add unpredictable breakages 

I agree. It's especially bad in docker 😅

> I think it would be a good idea to set static version numbers for the other tests

I get the concept but the problem is that people will just `brew install terraform`, get the latest version and then use this module, and perhaps have an issue. But we need to know if the latest version has a problem with this module.

So I think we merge this and see how it goes unless someone has a better idea.

PR #576, opened 2019-10-30
Prepare for new release 6.1.0

barryib:


max-rocket-internet:
> If this contains breaking changes should it not have been 7.0.0?

Oh yes, you are correct! My mistake. Maybe we should make a new release and delete the current tag to be consistent @barryib? I can do it

PR #584, opened 2019-11-06
Userdata parts

thomas-brx:
# PR o'clock

## Description

Support added for `userdata_parts` in both `worker_groups` and `worker_groups_launch_template`. This changes the `user_data` to use [multipart MIME configuration](https://www.terraform.io/docs/providers/template/d/cloudinit_config.html), instead of a single script.

This allows for a finer control of all the phases of cloud-init.
```
 worker_groups = [
    {
      ...
      userdata_parts = [
        {
          content_type = "text/cloud-config"
          filename = "cloud-config"
          merge_type = "list(append)+dict(recurse_array)+str()"

          content = <<CLOUD_CONFIG
packages:
  - strace

bootcmd:
  - echo "ENVIROMNET=${var.environment}" >> /etc/environment
CLOUD_CONFIG
        },
      ]
    }
  ]
```
### Checklist

- [ ] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Test against minimum versions specified in `versions.tf` (by @dpiddockcmp)
+- Added support for customizing userdata with `userdata_parts` list in `worker_groups` and `worker_groups_launch_template` (by @thomas_brx)
 
 ### Changed
 
--- a/data.tf
+++ b/data.tf
@@ -217,3 +217,25 @@ data "aws_iam_instance_profile" "custom_worker_group_launch_template_iam_instanc
     local.workers_group_defaults["iam_instance_profile_name"],
   )
 }
+
+data "template_cloudinit_config" "worker_group_cloudinit_config" {
+  count         = local.worker_group_count
+  base64_encode = true
+  gzip          = false
+
+  part {
+    content_type = "text/x-shellscript"
+    content      = data.template_file.userdata.*.rendered[count.index]
+  }
+}
+
+data "template_cloudinit_config" "worker_group_launch_template_cloudinit_config" {
+  count         = local.worker_group_launch_template_count
+  base64_encode = true
+  gzip          = false
+
+  part {
+    content_type = "text/x-shellscript"
+    content      = data.template_file.launch_template_userdata.*.rendered[count.index]
+  }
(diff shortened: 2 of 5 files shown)


max-rocket-internet:
In principal I like the idea as it's very elegant if you are doing complicated things in the user data. But it's a deviation away from the AWS design: https://amazon-eks.s3-us-west-2.amazonaws.com/cloudformation/2019-10-08/amazon-eks-nodegroup.yaml

```json
      UserData: !Base64
        "Fn::Sub": |
          #!/bin/bash
          set -o xtrace
          /etc/eks/bootstrap.sh ${ClusterName} ${BootstrapArguments}
          /opt/aws/bin/cfn-signal --exit-code $? \
                   --stack  ${AWS::StackName} \
                   --resource NodeGroup  \
                   --region ${AWS::Region}
```

I know we have already somewhat deviated away from the AWS design but this PR could perhaps break something for people that are using a custom AMI?

Let's see if other people have opinions either way.

max-rocket-internet:
Another option is we just simply allow supplying your own complete user data and then you can have the `data.template_cloudinit_config` resource outside of the module.

thomas-brx:
> I know we have already somewhat deviated away from the AWS design but this PR could perhaps break something for people that are using a custom AMI?

I guess it could be a problem if the custom AMI does not support multipart MIME. That could be fixed by not using `data.template_cloudinit_config` in the module if `userdata_parts` is not set.

I can fix that if you feel it's worth going forward with this PR.

> Another option is we just simply allow supplying your own complete user data and then you can have the `data.template_cloudinit_config` resource outside of the module.

That could possibly work, the only problem I see with that is the responsibility of the actual bootstrapping (`/etc/eks/bootstrap.sh --b64-cluster-ca ...`), which should really be part of the module? In that case, the module would still need logic to add that part to the user data, which could be tricky, as it will not really know the format of the supplied custom user data.

max-rocket-internet:
Hi @thomas-brx

I've given it thought and I think we should go in the direction of just allowing a user to provide the full user data from outside the module: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/601

The reasons are:

- It's simpler as we don't have to manage an extra resource in this module
- We are trying to reduce complexity
- It doesn't deviate from the AWS user-data by default

I hope that makes sense!

If you fancy making a PR to allow full user-data per worker group, that would be awesome 😃

PR #588, opened 2019-11-08
Add the 'create_cluster' variable.

bmcstdio:
# PR o'clock

## Description

Sometimes one needs to have a way to create an AWS EKS cluster conditionally, but Terraform does not allow the usage of `count` inside a `module` block. The workaround (inspired by [`terraform-aws-vpc`](https://github.com/terraform-aws-modules/terraform-aws-vpc#conditional-creation)) is to support a `create_cluster` variable which can be set to `false` to disable the creation of the AWS EKS cluster and its associated resources.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Test against minimum versions specified in `versions.tf` (by @dpiddockcmp)
+- Allow for conditionally creating the AWS EKS cluster and associated resources via `create_cluster` (by @bmcstdio)
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


max-rocket-internet:
What's your use case @bmcstdio?

I think this is essentially the same as https://github.com/terraform-aws-modules/terraform-aws-eks/pull/580 ?

bmcstdio:
> What's your use case @bmcstdio?

The main use case is to be able to specify whether a cluster exists or not in a given AWS account (being that the same "code" is run against two different AWS accounts using two different TFE workspaces).

> I think this is essentially the same as #580 ?

I wasn't aware of that PR, and it didn't pop up my searches. 😞 Judging by a quick look, the end result may be the same, yes.

max-rocket-internet:
> The main use case is to be able to specify whether a cluster exists or not in a given AWS account (being that the same "code" is run against two different AWS accounts using two different TFE workspaces).

Makes sense!

I'm happy to merge either PR but I think in the other one we decided on calling the option `create_eks`

max-rocket-internet:
I tested this from `examples/basic` with `create_cluster = false` and I get lots of these errors:

```
Error: Invalid index

  on ../../local.tf line 22, in locals:
  22:   default_ami_id_linux   = data.aws_ami.eks_worker.0.id
    |----------------
    | data.aws_ami.eks_worker is empty tuple

The given key does not identify an element in this collection value.


Error: Invalid index

  on ../../local.tf line 23, in locals:
  23:   default_ami_id_windows = data.aws_ami.eks_worker_windows.0.id
    |----------------
    | data.aws_ami.eks_worker_windows is empty tuple

The given key does not identify an element in this collection value.
```

max-rocket-internet:
This functionality was added in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/580 😃

PR #602, opened 2019-11-20
Support for EKS Managed Node Groups

wmorgan6796:
# PR o'clock

## Description
This adds support for the newly announced AWS EKS Managed Node Groups. Though the allowed config options are pretty limited, this adds initial support.

### Checklist

- [x ] Change added to CHANGELOG.md. All changes must be added and breaking changes and highlighted
- [ x] CI tests are passing
- [x ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation


wbertelsen:
I tried this out, im not sure its compatible with `manage_worker_iam_resources: true`
Getting
```
Error: error creating EKS Node Group (redacted:redacted): InvalidParameterException: The provided nodeRole is invalid.
	status code: 400, request id:xx
```

wmorgan6796:
> `manage_worker_iam_resources`

This is due to the fact that Managed Node Groups will not work without using a role that has certain policies attached. The policies that need to be attached can be seen here: [EKS Node IAM Role](https://docs.aws.amazon.com/eks/latest/userguide/worker_node_IAM_role.html). I can add a change to make it so that the role will be made correctly.

wmorgan6796:
@max-rocket-internet I have added the example and fixed a bug that @wbertelsen found as well.

pierresteiner:
We are also really looking forward to this PR, awesome work.

I saw nothing in the changes regarding the aws auth config map. I didn't play much with the product, but I was expecting some simplification in that regard, as the master knows about the worker.

wmorgan6796:
@pierresteiner the aws auth configmap isn't for the workers, its for delegating access to the cluster for IAM roles. Managed Node Groups don't simplify that.

lcycon:
Looks like there is an issue with the `remote_access` attribute. By my read of things, it appears that a `remote_access` block with empty `key_name` and security group ID list is rewritten to a null remote access block by the AWS API. Because of this, Terraform believes it must recreate the workers on every apply.

One work around is to delete the `remote_access` block entirely if it is unused.

wmorgan6796:
I’m unsure if it’s possible to completely remove a block if unused. Maybe it would be a good case to note this in the docs? This seems like more of a bug either on the AWS provider or the AWS API. Any suggested edits would be welcome

lcycon:
I agree this seems like a bug with the provider.

I think one potential workaround for the module is to declare two versions of each worker group (one with the remote access and one that omits it entirely). You may then be able to conditionally choose one over the other (via count trickery) depending on whether a key name is set.

pierresteiner:
> @pierresteiner the aws auth configmap isn't for the workers, its for delegating access to the cluster for IAM roles. Managed Node Groups don't simplify that.

It is actually both, prior to managed node it was necessary to also grant permission to workers role so that they could join the cluster. I don't know if it  is still needed (https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html) though

wmorgan6796:
> It is actually both, prior to managed node it was necessary to also grant permission to workers role so that they could join the cluster. I don't know if it is still needed (https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html) though

Hmm,  I wasn't aware of that. I'm hesitant to change too many things as I'm not familiar with the true inner workings of some parts of this module.

wmorgan6796:
> I agree this seems like a bug with the provider.
> 
> I think one potential workaround for the module is to declare two versions of each worker group (one with the remote access and one that omits it entirely). You may then be able to conditionally choose one over the other (via count trickery) depending on whether a key name is set.

I think I mitigated this issue with my latest commit. I just made it so that if the ssh key isn't set, then automatically make the `source_security_group_ids` an empty list, regardless of what the user puts there.

splieth:
Shouldn't the worker group also have a `create_before_destroy = true` on it? I guess right now all nodes would be destroyed before new ones would be created.

wmorgan6796:
@splieth terraform throws an error if I add create_before_destroy to the resource. I don't want to make any changes to any of the other worker groups as it would be outside the scope of this PR.

@max-rocket-internet What else needs to be done to get this merged?

wbertelsen:
Just tried again from `wmorgan6796:master`

Still seeing replacement forced on my worker groups due to `remote_access` / `remote_access_security_group_ids = []` getting turned into `""` by aws.

```
      ~ resources       = [
          - {
              - autoscaling_groups              = [
                  - {
                      - name = "eks-REDACTED"
                    },
                ]
              - remote_access_security_group_id = ""
            },
        ] -> (known after apply)
      ~ status          = "ACTIVE" -> (known after apply)
```

When making

splieth:
@wmorgan6796 I just tested it myself with adding 

```
lifecycle {
    create_before_destroy = true
}
```

to the resource (`aws_eks_node_group`) locally. Worked for me without any errors from terraform.

Personally I'm fine with opening a pull request for that. But actually I don't get why it is out of the scope of this PR.

For example if I would  alter the scaling configuration after the first deployment, this would lead the deletion of the current autoscaling group and a recreation of it afterwards with the current implementation of the module. For large clusters, this would mean a significant downtime.

On the other hand, having set the lifecycle policy the new autoscaling group would be created beforehand and the old nodes drained and terminated afterwards. For me this is the major advantage of using managed nodes and also what I would expect when using this module.

wmorgan6796:
I was mistaken about how to add it. I’ll add it to the PR.

wbertelsen:
Sorry to report so many issues with this much appreciated PR but just a note: to make `create_before_destroy` work you need to use dynamically generated  nodegroup names or AWS will complain
```
Error: error creating EKS Node Group (cluster-name:cluster-name-group-name): ResourceInUseException: NodeGroup already exists with name cluster-name-group-name)and cluster name cluster-name
	status code: 409, request id: xxx
```

wbertelsen:
terraform apply

wmorgan6796:
I believe that I've fixed the issue now

variables.tf
```diff
@@ -293,3 +293,9 @@ variable "attach_worker_cni_policy" {
   type        = bool
   default     = true
 }
+
+variable "worker_group_managed_node_groups" {
+  description = "A list of maps defining worker group configurations to be defined using AWS EKS Managed Node Groups. See workers_group_defaults for valid keys."
+  type        = any
```
eytanhanig: The type should probably be either [list(any)](https://www.terraform.io/docs/configuration/types.html#list-) or, better yet, [list(object({for_examples = follow_this_link}))](https://www.terraform.io/docs/configuration/variables.html#declaring-an-input-variable).

workers_node_groups.tf
```diff
@@ -0,0 +1,176 @@
...
+}
+
+resource "aws_iam_role_policy_attachment" "managed_node_groups_AmazonEC2ContainerRegistryReadOnly" {
+  count      = local.worker_group_managed_node_group_count > 0 ? 1 : 0
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
+  role       = aws_iam_role.managed_node_groups[0].name
+}
+
+resource "aws_iam_role_policy_attachment" "managed_node_groups_additional_policies" {
+  count      = local.worker_group_managed_node_group_count > 0 ? length(var.workers_additional_policies) : 0
```
eytanhanig: This would be a really good place to use `for_each` instead of `count` so that reordering or removing a policy doesn't cause it to re-attach every single one. You can [find more info here](https://www.terraform.io/docs/configuration/resources.html#using-sets).

It would probably end up looking something like this:
```hcl
resource "aws_iam_role_policy_attachment" "managed_node_groups_additional_policies" {
  for_each = toset(var.workers_additional_policies)
  
  role       = aws_iam_role.managed_node_groups[0].name
  policy_arn = each.key
}
```

I suspect you can replace almost every instance of `count` with something like this, however it's less important for cases when at most there where be one entity.

variables.tf
wmorgan6796: I would make it list(any) but none of the other 2 config lists (`worker_groups` or `worker_groups_launch_template`) are listed that way. I usually think that its better for the code to fit in vs. be more technically correct. That way its a cohesive whole

workers_node_groups.tf
```diff
@@ -0,0 +1,176 @@
...
+resource "aws_iam_policy" "managed_node_groups_autoscaling" {
+  count       = var.manage_worker_autoscaling_policy && local.worker_group_managed_node_group_count > 0 ? 1 : 0
+  name_prefix = "eks-worker-autoscaling-${aws_eks_cluster.this.name}"
+  description = "EKS worker node autoscaling policy for cluster ${aws_eks_cluster.this.name}"
+  policy      = data.aws_iam_policy_document.worker_autoscaling.json
+  path        = var.iam_path
+}
+
+resource "random_pet" "managed_node_groups" {
+  count = local.worker_group_managed_node_group_count
```
eytanhanig: Every occurrence of `count = local.worker_group_managed_node_group_count` is strong candidate for switching to `for_each = toset(var.workers_additional_policies)`.

variables.tf
wmorgan6796: If I am going to change that, I should change the other 2 as well. Would that fall under this PR or be a different one?

workers_node_groups.tf
wmorgan6796: Can I replace that look up though? I would still need to provide a default value

examples/managed_node_groups/main.tf
```diff
@@ -0,0 +1,104 @@
+terraform {
+  required_version = ">= 0.12.0"
```
eytanhanig: This should match what's in the versions.tf file (which actually should have been bumped to `>= 0.12.6` when the module started using `for_each` in `workers.tf` and `workers_launch_template.tf`).

workers_node_groups.tf
wmorgan6796: This is much simpler. I will make this change though it will differentiate the code of the managed node groups from the rest of the worker groups in terms of code

examples/managed_node_groups/main.tf
wmorgan6796: done

workers_node_groups.tf
eytanhanig: Good point. Given that [node_group_name](https://www.terraform.io/docs/providers/aws/r/eks_node_group.html) is a required argument for the eks_node_group resource it seems reasonable to ask that module users specify it along with things like the max size.

variables.tf
eytanhanig: Do a separate commit. Altering the existing resources will be a breaking change for users and therefore require a major version increment, whereas introducing it here is 100% safe given that it's brand new functionality.

workers_node_groups.tf
wmorgan6796:
```
The given "for_each" argument value is unsuitable: "for_each" supports maps
and sets of strings, but you have provided a set containing type object.
```
Can't use for each on the random_pets

variables.tf
wmorgan6796: I resolved this for managed worker groups but didn't touch this for any others

wmorgan6796:
@eytanhanig @max-rocket-internet I updated the minimum versions of Terraform so that we can make use of the for_each blocks for resources. But the linting job for the Github CI only has terraform 0.12.2. Should that be updated or should I revert that change?

eytanhanig:
> @eytanhanig @max-rocket-internet I updated the minimum versions of Terraform so that we can make use of the for_each blocks for resources. But the linting job for the Github CI only has terraform 0.12.2. Should that be updated or should I revert that change?

It appears that you'll need to update the image used in lint.yml.

workers_node_groups.tf
eytanhanig: Node groups must have unique names, therefore you can use those as an index. Try creating a local variable that maps the name to the dictionary and instead reference that everywhere:

```hcl
locals {
  node_groups = { for obj in toset(var.worker_group_managed_node_groups) : obj.name => obj }
}
```

max-rocket-internet:
> I updated the minimum versions of Terraform so that we can make use of the for_each blocks for resources. But the linting job for the Github CI only has terraform 0.12.2. Should that be updated or should I revert that change?

Yes please also update that: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/workflows/lint.yml#L59

variables.tf
dpiddockcmp: Hi. This really should *not* be `list(any)`. `any` is a stupid name for what this type does. It really means "any of the same type" and Terraform will try to convert items to the same type. With `list(any)` it will try to convert to a single type, usually `string` if one is present. Or throw errors when incompatible types are used, say only booleans and numbers. This list can contain incompatible mixed types like sublists and numbers. When used as a bareword like originally, terraform will convert to an Object using the types given. 

`list(object())` is even worse. When you specify an `object` the caller has to exactly match the keys listed. It would be a massive pain for all users of this module.

`type = any` is the only correct answer here.

eytanhanig: I stand corrected. `type = any` it is then.

workers_node_groups.tf
wmorgan6796: @eytanhanig I can't seem to fix this error when using for_each:
```
Error: Invalid function argument

  on ../../node_groups.tf line 88, in resource "aws_eks_node_group" "workers":
  87: 
  88:       [
  89: 
  90: 
  91: 
  92: 
  93: 
    |----------------
    | count.index is 0
    | random_pet.node_groups is object with 1 attribute "example"
    | var.node_groups is tuple with 1 element

Invalid value for "list" parameter: element 2: string required.
```
Can you take a look at it and see what you can find?

eytanhanig: Because you're switching from count to for_each, `var.node_groups[count.index]` becomes `each.value`.

If we require that each node_group have a unique name then you can ditch the use of `random_pet`, resulting in this block simplifying even further:

```hcl
resource "aws_eks_node_group" "workers" {
  for_each = local.node_groups

  cluster_name    = var.cluster_name
  node_group_name = join("-", [var.cluster_name, each.key])
  node_role_arn   = lookup(each.value, "iam_role_arn", aws_iam_role.node_groups[0].arn)
  subnet_ids      = lookup(each.value, "subnets", local.workers_group_defaults["subnets"])

  scaling_config {
    desired_size = lookup(each.value, "node_group_desired_capacity", local.workers_group_defaults["asg_desired_capacity"])
    max_size     = lookup(each.value, "node_group_max_capacity", local.workers_group_defaults["asg_max_size"])
    min_size     = lookup(each.value, "node_group_min_capacity", local.workers_group_defaults["asg_min_size"])
  }

  disk_size       = lookup(each.value, "root_volume_size", null)
  instance_types  = [lookup(each.value, "instance_type", null)]
  labels          = lookup(each.value, "node_group_k8s_labels", null)
  release_version = lookup(each.value, "ami_release_version", null)

  remote_access {
    ec2_ssh_key               = lookup(each.value, "key_name", "") != "" ? each.value["key_name"] : null
    source_security_group_ids = lookup(each.value, "key_name", "") != "" ? lookup(
      each.value,
      "source_security_group_ids",
      []
    ) : null
  }

  tags = lookup(each.value, "node_group_additional_tags", null)

  lifecycle {
    create_before_destroy = true
  }
}
```

eytanhanig:
I believe that I've found a bug in the AWS provider resource `aws_eks_node_group`: Idempotency is broken - and therefore subsequent applies attempt to create & destroy the node group - when `remote_access` is used in the following way:
- `remote_access {}`
- `remote_access {}` with `ec2_ssh_key = ""`
- `remote_access {}` with `ec2_ssh_key = null`

It **does become idempotent** when used with a **valid** SSH key. I've not tested it with permutations of the `source_security_group_ids` argument.

To account for this I believe that that we are limited to the following options:
- Completely remove the `remote_access` configuration block.
- Require the inclusion of a valid SSH key.
- Wait for a fix to become available before merging this PR.

node_groups.tf
```diff
@@ -92,7 +88,7 @@ resource "aws_eks_node_group" "workers" {
       [
         aws_eks_cluster.this.name,
         lookup(var.node_groups[count.index], "name", count.index),
-        random_pet.node_groups[count.index].id
+        random_pet.node_groups[lookup(var.node_groups[count.index], "name", count.index)]
```
saviogrossi: Hello, I'm getting this error:
```
Error: Invalid function argument

  on .terraform/modules/eks/node_groups.tf line 88, in resource "aws_eks_node_group" "workers":
  87: 
  88:       [
  89: 
  90: 
  91: 
  92: 
  93: 
    |----------------
    | aws_eks_cluster.this.name is "toro-dev2"
    | count.index is 0
    | random_pet.node_groups is object with 1 attribute "standard"
    | var.node_groups is tuple with 1 element

Invalid value for "list" parameter: element 2: string required.
```

looks like `.id` is missing here:
```
random_pet.node_groups[lookup(var.node_groups[count.index], "name", count.index)].id
```
tks!

eytanhanig: resource "random_pet" "node_groups" {
  for_each = local.node_groups

  separator = "-"
  length    = 2

  keepers = {
    instance_type = lookup(each.value, "instance_type", local.workers_group_defaults["instance_type"])

    ec2_ssh_key = lookup(each.value, "key_name", local.workers_group_defaults["key_name"])

    source_security_group_ids = join("-", compact(
      lookup(each.value, "source_security_group_ids", local.workers_group_defaults["source_security_group_id"]
    )))

    node_group_name = join("-", [var.cluster_name, each.value["name"]])
  }
}

resource "aws_eks_node_group" "workers" {
  for_each = local.node_groups

  node_group_name = join("-", [var.cluster_name, each.key, random_pet.node_groups[each.key].id])

  cluster_name    = var.cluster_name
  node_role_arn   = lookup(each.value, "iam_role_arn", aws_iam_role.node_groups[0].arn)
  subnet_ids      = lookup(each.value, "subnets", local.workers_group_defaults["subnets"])

  scaling_config {
    desired_size = lookup(each.value, "node_group_desired_capacity", local.workers_group_defaults["asg_desired_capacity"])
    max_size     = lookup(each.value, "node_group_max_capacity", local.workers_group_defaults["asg_max_size"])
    min_size     = lookup(each.value, "node_group_min_capacity", local.workers_group_defaults["asg_min_size"])
  }

  disk_size       = lookup(each.value, "root_volume_size", null)
  instance_types  = [lookup(each.value, "instance_type", null)]
  labels          = lookup(each.value, "node_group_k8s_labels", null)
  release_version = lookup(each.value, "ami_release_version", null)

  # This sometimes breaks idempotency as described in https://github.com/terraform-providers/terraform-provider-aws/issues/11063
  remote_access {
    ec2_ssh_key               = lookup(each.value, "key_name", "") != "" ? each.value["key_name"] : null
    source_security_group_ids = lookup(each.value, "key_name", "") != "" ? lookup(each.value, "source_security_group_ids", []) : null
  }

  version = aws_eks_cluster.this.version

  tags = lookup(each.value, "node_group_additional_tags", null)

  lifecycle {
    create_before_destroy = true
  }
}

wmorgan6796: @eytanhanig I added the changes you suggested and the terraform plan seems to work.

wmorgan6796:
whats the status of this?

Issue #624, opened 2019-12-10
Terraform apply breaks communications between workers and masters


xposix:
# I have issues
Described below
## I'm submitting a...

* [X] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
Every time I execute a terraform apply, it tries to make the following changes
```
  # module.eks-cluster.kubernetes_config_map.aws_auth[0] will be updated in-place
  ~ resource "kubernetes_config_map" "aws_auth" {
        binary_data = {}
      ~ data        = {
            "mapAccounts" = jsonencode([])
          ~ "mapRoles"    = <<~EOT
              - - groups:
              -   - system:bootstrappers
              -   - system:nodes
              -   rolearn: arn:aws:iam::1234567890:role/mytests-managed-node-groups
              -   username: system:node:{{EC2PrivateDNSName}}
              +
              +
            EOT
            "mapUsers"    = jsonencode([])
        }
        id          = "kube-system/aws-auth"

        metadata {
            annotations      = {}
            generation       = 0
            labels           = {}
            name             = "aws-auth"
            namespace        = "kube-system"
            resource_version = "35037"
            self_link        = "/api/v1/namespaces/kube-system/configmaps/aws-auth"
            uid              = "781f4175-1b42-11ea-b157-02d81c51517c"
        }
    }
```
after that,  the node becomes 'NotReady'. 
```
NAME                                         STATUS     ROLES    AGE   VERSION
ip-10-10-10-180.eu-west-1.compute.internal   NotReady   <none>   55m   v1.14.7-eks-1861c5
```

When looking into the managed node logs at `/var/log/messages`:
```
Dec 10 17:32:49 ip-10-10-10-180 kubelet: E1210 17:32:49.651699    3821 reflector.go:126] k8s.io/kubernetes/pkg/kubelet/kubelet.go:442: Failed to list *v1.Service: Unauthorized
Dec 10 17:32:50 ip-10-10-10-180 kubelet: E1210 17:32:50.130065    3821 reflector.go:126] k8s.io/kubernetes/pkg/kubelet/kubelet.go:451: Failed to list *v1.Node: Unauthorized
[...]

PR #627, opened 2019-12-11
Fix aws-auth config map for managed node groups

wbertelsen:
# PR o'clock

## Description

This change adds the IAM role used for each managed node group to the
aws-auth config map. This fixes an issue where managed nodes can not
access the EKS kubernetes API server.

This should fix https://github.com/terraform-aws-modules/terraform-aws-eks/issues/624
### Checklist

- [x] CI tests are passing


andres-de-castro:
How can I help this get approved?

max-rocket-internet:
I'm still not sure how to integrate the managed node groups into this module.

See here for some background: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635

The thing is, we have this now:

- Worker groups with LC
- Work groups with LT
- Managed node groups
- Fargate (coming soon)

And I have no clusters using Managed node groups to test with.

wbertelsen:
@max-rocket-internet I've been watching the discussion in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635 and I buy the idea that the worker/node groups could be submodules. 

However, managed node groups are broken on master without this patch, so I think we should work with other testers to get this merged, or revert the original node group PR (https://github.com/terraform-aws-modules/terraform-aws-eks/pull/602) as broken.

I've tested this myself on a real cluster, but obviously there should be another set of eyes on it.

@andres-de-castro: Are you able to test this out on a cluster and report back if it works?
If you're not already aware, you can specify any git ref when sourcing the module:

```
 source          = "github.com/wbertelsen/terraform-aws-eks?ref=fix-node-group-aws-auth"
```

TBeijen:
I'll take a look. Was trying out latest master and indeed noticed a second plan attempting to delete the managed-nodes from `aws-auth`. Probably added by AWS EKS itself (similar to how it updates VPC & subnet tags)

max-rocket-internet:
@TBeijen that would be great. I think then we merge this and make a new release with the node group stuff included.

TBeijen:
Ok, ran some tests:

* Added maanged node group to a cluster with 1 launch-template based worker group using latest master 583c32d
* Noticed 2nd plan would remove the `rolearn: arn:aws:iam::NNNNNN:role/apps-1-managed-node-groups` from `mapRoles`
* Switched to `github.com/wbertelsen/terraform-aws-eks?ref=fix-node-group-aws-auth`, ran plan & apply. Observed that:

    * Plan showed preserving the `aws-auth` entry
    * `aws-auth` values are now quoted. Looks like valid yaml so...
    * After apply I wasn't locked out of my cluster. Nodes from both launch template and managed node group reported as `Ready`.

* Removed the `node_groups` block from TF code (testing if it renders fine without any managed node groups):

    * Plan and apply looked and ran ok.
    * After apply I wasn't locked out of my cluster. Nodes reported `Ready`, `aws-auth` looks good.

So, based on the above I'd say this fixes `aws-auth` as it's expected to do.

PR #186, opened 2018-11-07
Add ability to pass computed values to cluster_security_group_id and worker_security_group_id

rmakram-ims:
# PR o'clock

## Description

Since we cannot pass computed values to cluster_security_group_id or worker_security_group_id, I specified a cluster_create_security_group and worker_create_security_group.

### Checklist

- [X] `terraform fmt` and `terraform validate` both work from the root and `examples/eks_test_fixture` directories (look in CI for an example)
- [x] Tests for the changes have been added and passing (for bug fixes/features)
- [x] Test results are pasted in this PR (in lieu of CI)
- [X] Docs have been updated using `terraform-docs` per `README.md` instructions
- [X] I've added my change to CHANGELOG.md
- [x] Any breaking changes are highlighted above

PR #631, opened 2019-12-12
Control plane security group always whitelist worker security group and revert #186

ryanooi:
# PR o'clock

## Description

1. Main change on cluster.tf `aws_security_group_rule.cluster_https_worker_ingress` intended to whitelist port 443 for newly created security group or self attach security group from worker's security group. Self attach control plane security group will not able to whitelist worker security group before it get created therefore additional step will required.
2. Revert PR #186 as Terraform v0.12.x fixed "count cannot be computed" issue
2.1  Remove `cluster_create_security_group` and `worker_create_security_group` to improve consistency. There is no guarantee/enforcement when you set `cluster_create_security_group` or `worker_create_security_group` to `false`, user will also provide `cluster_security_group_id` or `worker_security_group_id`. 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Fix cluster_oidc_issuer_url output from list to string (by @chewvader)
 - Fix idempotency issues for node groups with no remote_access configuration (by @jeffmhastings)
 - Added support to create IAM OpenID Connect Identity Provider to enable EKS Identity Roles for Service Accounts (IRSA). (by @alaa)
-- Improve UX of create new EKS cluster (by @ryanooi)
+- **Breaking:** Change logic of security group whitelisting. Will always whitelist worker security group on control plane security group either provide one or create new one. See Important notes below for upgrade notes (by @ryanooi)
 
 #### Important notes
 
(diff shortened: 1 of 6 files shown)


ryanooi:
Need some explanation? or?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Updated minimum version of Terraform to avoid a bug (by @dpiddockcmp)
 - Fix cluster_oidc_issuer_url output from list to string (by @chewvader)
 - Fix idempotency issues for node groups with no remote_access configuration (by @jeffmhastings)
+- Improve UX of create new EKS cluster (by @ryanooi)
```
max-rocket-internet: Can you update this to describe your changes more accurately? i.e. "changed logic around SG and SG rule creation"

max-rocket-internet:
Actually, this is a breaking change as you are removing some vars. So also please write details under `Important notes` that clearly explain how to migrate after this change 🙂

CHANGELOG.md
ryanooi: Updated. Please help review again :)

Issue #621, opened 2019-12-10
Error: Post https://xxxx.eks.amazonaws.com/api/v1/namespaces/kube-system/configmaps: dial tcp xxx:443: i/o timeout


max-rocket-internet:
On every example on first run:

```
Error: Post https://C39B1571479D46ED13A85BC0105510E2.gr7.us-west-2.eks.amazonaws.com/api/v1/namespaces/kube-system/configmaps: dial tcp 44.229.104.205:443: i/o timeout

  on ../../aws_auth.tf line 45, in resource "kubernetes_config_map" "aws_auth":
  45: resource "kubernetes_config_map" "aws_auth" {
```

This is because theres a small between when the cluster endpoint is returned after creation but is not available.

If you rerun `terraform apply` the configmap is created fine.

Quite annoying 😅

dpiddockcmp:
Is this an EKS bug that's worth reporting to AWS container team, or an interplay between the aws and kubernetes providers?

xposix:
I keep on having the same issue. IMO, the kubernetes provider should wait / retry few times before coming back with an error.

max-rocket-internet:
Related: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/96

max-rocket-internet:
> Is this an EKS bug that's worth reporting to AWS container team, or an interplay between the aws and kubernetes providers?

I think a socket error should be retried by the provider.

In the mean time, could we make a work around somehow? Like a null resource that sleeps for 10 seconds after cluster creation?

ivanich:
I'm using kubectl hack for this but it requires kubectl and a lot of `depends_on` and in general it ugly
```resource "null_resource" "wait-eks-api" {
  provisioner "local-exec" {
    command = "until kubectl --kubeconfig ${path.root}/kubeconfig_${var.cluster_id} -n kube-system get pods >/dev/null 2>&1;do echo 'Waiting for EKS API';sleep 5;done"
  }
}
```
in the meantime I'm going to fork eks module with kubernetes provider commits reverted because it's not usable for our usecases.

max-rocket-internet:
> in the meantime I'm going to fork eks module with kubernetes provider commits reverted because it's not usable for our usecases.

You can just use release `v7.0.1`, no?

ivanich:
> You can just use release v7.0.1, no?

[...]

PR #639, opened 2019-12-16
Wait cluster responsive

shaunc:
# PR o'clock

## Description

I experience intermittent problems applying config_map as cluster is not immediately ready to respond to kubectl after create. This PR adds a null resource that sleeps until the cluster responds to kubectl.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -63,7 +63,7 @@ resource "null_resource" "wait_for_kubernetes" {
 
 resource "kubernetes_config_map" "aws_auth" {
   depends_on = [null_resource.wait_for_kubernetes]
-  count = var.create_eks && var.manage_aws_auth ? 1 : 0
+  count      = var.create_eks && var.manage_aws_auth ? 1 : 0
 
   metadata {
     name      = "aws-auth"
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -52,7 +52,17 @@ data "template_file" "node_group_arns" {
   }
 }
 
+resource "null_resource" "wait_for_kubernetes" {
+  depends_on = [local_file.kubeconfig]
+  provisioner "local-exec" {
+    command = <<EOT
+    until kubectl version --kubeconfig ${local.kubeconfig_filename} >/dev/null; do sleep 4; done
+  EOT
+  }
+}
+
 resource "kubernetes_config_map" "aws_auth" {
+  depends_on = [null_resource.wait_for_kubernetes]
   count = var.create_eks && var.manage_aws_auth ? 1 : 0
 
   metadata {
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -41,6 +41,7 @@ data "template_file" "worker_role_arns" {
     )
   }
 }
+<<<<<<< HEAD
 
 data "template_file" "node_group_arns" {
   count    = var.create_eks ? local.worker_group_managed_node_group_count : 0
@@ -52,6 +53,11 @@ data "template_file" "node_group_arns" {
   }
 }
 
+=======
+locals {
+  kubeconfig_filename = concat(local_file.kubeconfig.*.filename, [""])[0]
+}
+>>>>>>> fixup missing local for kubeconfig_filename
 resource "null_resource" "wait_for_kubernetes" {
   depends_on = [local_file.kubeconfig]
   provisioner "local-exec" {
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -41,7 +41,6 @@ data "template_file" "worker_role_arns" {
     )
   }
 }
-<<<<<<< HEAD
 
 data "template_file" "node_group_arns" {
   count    = var.create_eks ? local.worker_group_managed_node_group_count : 0
@@ -53,22 +52,8 @@ data "template_file" "node_group_arns" {
   }
 }
 
-=======
-locals {
-  kubeconfig_filename = concat(local_file.kubeconfig.*.filename, [""])[0]
-}
->>>>>>> fixup missing local for kubeconfig_filename
-resource "null_resource" "wait_for_kubernetes" {
-  depends_on = [local_file.kubeconfig]
-  provisioner "local-exec" {
-    command = <<EOT
-    until kubectl version --kubeconfig ${local.kubeconfig_filename} >/dev/null; do sleep 4; done
-  EOT
-  }
-}
-
 resource "kubernetes_config_map" "aws_auth" {
-  depends_on = [null_resource.wait_for_kubernetes]
+  depends_on = [aws_eks_cluster.this]
   count      = var.create_eks && var.manage_aws_auth ? 1 : 0
 
   metadata {
--- a/cluster.tf
+++ b/cluster.tf
@@ -31,6 +31,11 @@ resource "aws_eks_cluster" "this" {
     aws_iam_role_policy_attachment.cluster_AmazonEKSServicePolicy,
     aws_cloudwatch_log_group.this
   ]
+  provisioner "local-exec" {
+    command = <<EOT
+    until curl -k ${aws_eks_cluster.this[0].endpoint}/healthz >/dev/null; do sleep 4; done
+  EOT
+  }
 }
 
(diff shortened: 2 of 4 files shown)


max-rocket-internet:
This solves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/621

But, in this PR we are running `kubectl`, which kind of defeats the point of using the Kubernetes Terraform provider in the first place 😅

Let's see what others think about this approach or if there's a way we can do it more elegantly?

shaunc:
@max-rocket-internet ... well ... it undermines it, but doesn't defeat it entirely: after this completes, the provider abstraction is perfectly usable. However, I see your point. I wonder if there is a way of health-checking with curl... hmm... https://success.docker.com/article/how-to-poll-kubernetes-health-with-curl ?... also isn't so pretty....

barryib:
FWIW, you can get the server version with `curl -k https://<k8s-api-endpoint>/version?timeout=30s`. I don't know exactly if this is a blocking http call or not, but it worth to try to get this endpoint with the [http provider](https://www.terraform.io/docs/providers/http/data_source.html) 🤞 .

I also noticed that the `/healthz` is readable without creds. We can also test this endpoint.

max-rocket-internet:
Cool ideas @barryib. 

@shaunc could you test that?

shaunc:
I'll check that. I'm also thinking that, instead of a separate null_resource, it would be best to include this as a provisioner on the cluster itself, so anything that depends on the cluster doesn't have to separately specify "cluster" and "cluster is ready". Of course, in theory this could slow up create time for configuration that only depends on knowing cluster attributes (e.g. configuring assume role w/ the OIDC endpoint)... but only a minute or two compared with ~>10min for cluster create anyway.

shaunc:
Health check w/o credentials seems to work -- thanks @barryib! As in comment above, got rid of separate null_resource in favor of provisioner for the cluster itself. I found that terraform would figure out it could start creation of other things that depended on the cluster (that needed to use it) prematurely, and this PR didn't expose the null_resource anyway. In theory we could have a separate output for "cluster_ready" but IMHO not worth the complexity cost for the benefit.

shaunc:
(rebased)

cluster.tf
```diff
@@ -31,6 +31,11 @@ resource "aws_eks_cluster" "this" {
     aws_iam_role_policy_attachment.cluster_AmazonEKSServicePolicy,
     aws_cloudwatch_log_group.this
   ]
+  provisioner "local-exec" {
```
barryib: Any chance to avoid `local-exec` ?  The http provider doesn't work ? We wanted to be agnostic from the OS. Using the kubernetes provider was a huge improvement in that way and would like to avoid going back to local-exec with curl if possible.

shaunc: Ah ... missed that part of your comment; sorry! Hmm... I'd worry about three things: (1) That the http provider would sometimes timeout, and it doesn't provide any control over timeouts, (2) when not combined with the cluster resource itself, terraform could schedule other resources that depend on the cluster before its ready, and outside the module, its a bit opaque to know what to do (though it could be documented). (3) Doc says they verify chain of trust for https, and certificate is likely to be unchained.

I guess (2) could be possibly solved by having (say) cluster_id depend on an expression involving some output from http resource. (3) just testing will confirm/deny :)....

vs (1) -- I'm a terraform newbie ... do you know anything about http data source timeout behavior? Would seem to be a risk to me, but I'll go ahead and try it if you guys want to go that way.

shaunc: Most general would be to have variable for command be a template required to embed (say) ${URL}. I guess the shell varies by platform, but I guess this type of variable substitution should be broadly compatible? I'm not sure if the "until" loop is available "everywhere"? In any case, AFAIK the options are to code up a custom provider (or patch the http provider, to accept options for waiting, and for insecure https), or to try to make local-exec be as cross-platform as possible.

max-rocket-internet: > hen perhaps changing the kubernetes provider to wait until endpoint is up would also be an option.

https://github.com/terraform-providers/terraform-provider-kubernetes/issues/96

max-rocket-internet: I'm not sure what we should do now but open to opinions. Either merge with a new local-exec or think of something else 🙂

shaunc: I would think easiest way to fix something without upstream changes would be to use local-exec, but make it configurable for cross-platform support. Do you have an idea how you'd like to configure it? Do you want me to propose something? (Adding variables for configuration -- what program to use for "curl", for instance, or possibly whether to install curl appropriate for platform.)?

barryib: From what I'm reading from stackoverflow, it sounds like since powershell 3.O, `wget` and `curl` are aliases for `Invoke-WebRequest` http://support.moonpoint.com/os/windows/PowerShell/wget-curl.php. It means, we can probably use `curl` in both OS, but i'm not sure for arguments.

Any windows guy around ?

max-rocket-internet: I think we forget about Windows. It's proved to be too much effort to support both platforms natively in the past. Windows users can use Docker I think.

versions.tf
```diff
@@ -2,10 +2,11 @@ terraform {
...
-    local    = ">= 1.2"
-    null     = ">= 2.1"
-    template = ">= 2.1"
-    random   = ">= 2.1"
+    aws        = ">= 2.38.0"
+    local      = ">= 1.2"
+    null       = ">= 2.1"
+    template   = ">= 2.1"
+    random     = ">= 2.1"
+    kubernetes = "~> 1.10"
```
dpiddockcmp: We should only use `=>` here to avoid conflicts with users and other module definitions. 2.0 might be compatible with our resource. We can't know yet.

Our minimum compatible version is 1.6.2, the first correctly tagged with TF 0.12 support.
```suggestion
    kubernetes = "=> 1.6.2"
```

aws_auth.tf
```diff
@@ -41,9 +41,13 @@ data "template_file" "worker_role_arns" {
     )
   }
 }
+locals {
+  kubeconfig_filename = concat(local_file.kubeconfig.*.filename, [""])[0]
```
dpiddockcmp: Is this local needed for anything?

aws_auth.tf
```diff
@@ -41,9 +41,13 @@ data "template_file" "worker_role_arns" {
...
+}
 
 resource "kubernetes_config_map" "aws_auth" {
-  count = var.create_eks && var.manage_aws_auth ? 1 : 0
+  depends_on = [aws_eks_cluster.this]
```
dpiddockcmp: Is this explicit `depends_on` required? The provider doesn't work before the cluster's creation.

aws_auth.tf
shaunc: Nope ... will remove

aws_auth.tf
shaunc: It depends on how the kubernetes provider is set up(?) I wasn't sure it wasn't impossible make dependence on cluster opaque to terraform, depending on how the provider is setup, so thought explicit dependency would be useful to avoid bug in edge case.

shaunc:
(rebased)

aws_auth.tf
dpiddockcmp: Very true. It's not necessarily implicit from other variables.

cluster.tf
```diff
@@ -31,6 +31,11 @@ resource "aws_eks_cluster" "this" {
...
+    command = <<EOT
+    until curl -k ${aws_eks_cluster.this[0].endpoint}/healthz >/dev/null; do sleep 4; done
```
max-rocket-internet: This with no limit is OK?

max-rocket-internet:
@dpiddockcmp @barryib shall we merge this then?

cluster.tf
shaunc: (I thought that overall terraform timeout would be by default how user would control)

barryib:
LGTM.

dpiddockcmp:
LGTM. Create timeout is 30 minutes. Can always add fail-fast logic later if people complain.

Issue #633, opened 2019-12-12
Multiple node groups break if they don't all specify same keys


ioben:
# I have issues

## I'm submitting a...

* [x] bug report

## What is the current behavior?

If the following is specified, the error below occurs:

```
module "eks" {
  source = "github.com/terraform-aws-modules/terraform-aws-eks"

  cluster_name = local.cluster_name
  cluster_endpoint_public_access = false
  cluster_endpoint_private_access = true

  vpc_id = module.vpc.vpc_id
  subnets = module.vpc.private_subnets

  tags = {
    Name = local.cluster_name
  }

  node_groups = [
    {
      name = "test1"
      instance_type = "t3.2xlarge"
    },
    {
      name = "test2"
      instance_type = "t3.2xlarge"

      key_name = "test"
    },
  ]
}
```

```
$ terraform plan -refresh=false

Error: Invalid function argument

  on .terraform/modules/eks/node_groups.tf line 62, in resource "random_pet" "node_groups":
  62:       lookup(each.value, "source_security_group_ids", local.workers_group_defaults["source_security_group_id"]
  63:
    |----------------
    | local.workers_group_defaults["source_security_group_id"] is empty tuple

Invalid value for "default" parameter: the default value must have the same
type as the map elements.


Error: Invalid function argument

  on .terraform/modules/eks/node_groups.tf line 62, in resource "random_pet" "node_groups":
  62:       lookup(each.value, "source_security_group_ids", local.workers_group_defaults["source_security_group_id"]
  63:
    |----------------
    | local.workers_group_defaults["source_security_group_id"] is empty tuple

Invalid value for "default" parameter: the default value must have the same
type as the map elements.
```

If a few more node_group specific keys are added, then the below error occurs:

```
module "eks" {
  source = "github.com/terraform-aws-modules/terraform-aws-eks"

  cluster_name = local.cluster_name
  cluster_endpoint_public_access = false
  cluster_endpoint_private_access = true

  vpc_id = module.vpc.vpc_id
  subnets = module.vpc.private_subnets

  tags = {
    Name = local.cluster_name
  }

  node_groups = [
    {
      name = "test1"
      instance_type = "t3.2xlarge"

      node_group_desired_capacity = "3"
      node_group_min_size = "3"
      node_group_max_size = "3"

      node_group_k8s_labels = {
        "test" = "test"
      }

      node_group_additional_tags = {
        Name = "test"
      }
    },
    {
      name = "test2"
[...]

PR #644, opened 2019-12-22
Additional managed node outputs

TBeijen:
# PR o'clock

## Description

Work in progress for #643 

Additional node group output.

### Checklist

- [ ] Change added to CHANGELOG.md. All changes must be added and breaking changes and highlighted
- [ ] CI tests are passing
- [ ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

Issue #635, opened 2019-12-13
Is the complexity of this module getting too high?


max-rocket-internet:
A general question for users and contributors of this module 🙂

My feeling is that complexity getting too high and quality is suffering somewhat. We are squeezing a lot of features in a single module. This is by far the most complex module in the Terraform AWS modules org. But perhaps I just feel the pressure since I'm a maintainer?

Some recent examples:

- Adding global `create_eks` variable: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/580
- Adding Managed Node Groups: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/602
- Use kubernetes provider: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/355
- Add Windows support: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/555

And now:

- Adding IRSA: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/632
- Adding Fargate profiles: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/634

The amount of new resources being added to this module is just getting higher. And I don't see this slowing down as AWS is doubling down on EKS, as is everyone on k8s in general.

Don't get me wrong, these are all awesome contributions 💙 but I feel there's a price here. On every new release there's always bunch of new issues related to the recent changes.

Let me know your thoughts 😃

RothAndrew:
Yep. I've felt this way for quite a while now. And it's it's already too late. Too many new "features" have been added without letting things settle down and stabilize.

And, it's not fair to you as the maintainer, or to me as the user trying to run this thing in production. I'm cringing right now thinking about upgrading my prod cluster to v7.x. There's just no way I can reasonably safely do it.

IMO: deprecate this module and make several smaller, focused ones. Think long and hard about each new addition. Solicit feedback from multiple sources before leaping into a new feature. Even if somebody shows up with a PR with it all done if it ups the complexity of the module that should be considered carefully.

Another idea I have is to stop trying to support windows. There are unfortunately instances where a local_exec needs to run, and lots of complexity has been added trying to support windows alongside *nix. Instead, we can say "this is supported by linux (and probably mac). If you are on Windows, here is instructions on how to run this thing in a docker container.

RothAndrew:
Examples of small, focused modules:

1. Worker group module, separate from the main module(s)
1. Main EKS Cluster module
1. Fargate k8s module

[...]

PR #650, opened 2019-12-27
Node groups submodule

dpiddockcmp:
# PR o'clock

## Description

Move `node_group` resources to a submodule for code isolation. As generally discussed and bounced about in #635.

This also changes the `node_groups` variable to be a map of maps and associated resources use `for_each` instead of `count`. This was more complicated than expected, thanks Terraform. Idea originally came from a comment on a different PR: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/644#issuecomment-569111711

I think this is feature compatible with the current setup. Still some improvements to do.

### Checklist

- [x] CI tests are passing

### WIP
- [x] Add `node_groups` defaults to make user life easier
- [x] Clean up `local.workers_group_defaults_defaults`
- [ ] Defaults in the submodule? Only relevant if we want it to be easily usable standalone
- [x] Look at issues raised in #633 

### Related
Fixes: #633


max-rocket-internet:
Nice! I think we create a new release when this is done 🙂

dpiddockcmp:
I think it's ready for others to test and review.

Solved the race condition with `kubernetes_config_map` with a little use of a `null_data_source`.

For people using node_groups from the master branch it would be a good idea to target apply the node_group submodule first before a full apply that will delete the original node_group resources.

```
terraform apply -target=module.eks.module.node_groups
terraform apply
```

max-rocket-internet:
Question: A user can still use the submodule completely separately right? Like to allow using mixed versions of the main and MNG module? Like using the basic example but add this:

```hcl
module "mng1" {
  source                 = "git://github.com/terraform-aws-modules/terraform-aws-eks.git/modules/node_groups?ref=v999"
  create_eks             = true
  cluster_name           = module.eks.outputs.cluster_id
  cluster_version        = module.eks.outputs.cluster_version
  default_iam_role_arn   = module.eks.outputs.worker_iam_role_arn
  node_groups = {
    example = {
      desired_capacity = 1
      max_capacity     = 10
      min_capacity     = 1
      instance_type = "m5.large"
    }
  }
}
```

Second question: What do you think about forcing this approach? i.e. no direct connection between the core and mng module, i.e. no variables of `variable.node_groups*`.

modules/node_groups/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+  type        = map(string)
+}
+
+variable "node_groups_defaults" {
+  description = "map of maps of node groups to create. See default for valid keys and type. See source for extra comments"
+  type        = any
+
+  # This will be always be overriden by the value from the top level module
+  default = {
+    iam_role_arn              = ""           # IAM Role ARN for workers. If unset: uses `var.default_iam_role_arn`
```
barryib: By defining defaults here, can we just overwrite one (or some) default(s) ?

node_groups.tf
```diff
@@ -1,112 +1,28 @@
...
-
-    ec2_ssh_key = lookup(each.value, "key_name", local.workers_group_defaults["key_name"])
-
-    source_security_group_ids = join("-", compact(
-      lookup(each.value, "source_security_group_ids", local.workers_group_defaults["source_security_group_id"]
-    )))
-
-    node_group_name = join("-", [var.cluster_name, each.value["name"]])
+# Hack to ensure ordering of resource creation. Do not create node_groups
+# before other resources are ready. Removes race conditions
```
barryib: Ohhh.. Good to know. Thanks.

variables.tf
```diff
@@ -282,10 +282,16 @@ variable "create_eks" {
...
 
+variable "node_groups_defaults" {
+  description = "Map of values to be applied to all node groups. See `node_groups_defaults` variable in the `node_groups` submodule for more details"
+  type        = any
+  default     = {}
+}
+
 variable "node_groups" {
-  description = "A list of maps defining node group configurations to be defined using AWS EKS Managed Node Groups. See workers_group_defaults for valid keys."
+  description = "Map of map of node groups to create. See `node_groups` variable in the `node_groups` submodule for more details"
```
barryib: Just thinking... Is it better to have a map or maps or a list of maps with a key `name` which will converted into a map during [expansion](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/650/files#diff-52226e549f1fac8f550ec975e0bf901eR7) ? 

```hcl
  node_groups_expanded = { for k, v in var.node_groups : v.name => merge(
    var.node_groups_defaults,
    v,
  ) }
```

dpiddockcmp:
> Question: A user can still use the submodule completely separately right? Like to allow using mixed versions of the main and MNG module? Like using the basic example but add this:

They could but that might cause fun issues if the interface between the two is changed. My primary aim was to have the module as self contained code for readability rather than public use.

All variables need setting to something, including keys that "fall back" to `workers_group_defaults`. So you'd end up nearer to:
```hcl
module "mng1" {
  source                 = "git://github.com/terraform-aws-modules/terraform-aws-eks.git/modules/node_groups?ref=v999"
  create_eks             = true
  cluster_name           = module.eks.cluster_id
  cluster_version        = module.eks.cluster_version
  default_iam_role_arn   = module.eks.worker_iam_role_arn
  tags                   = {}
  node_groups_defaults   = {}
  workers_group_defaults = {}

  node_groups = {
    example = {
      subnets          = module.vpc.private_subnets
      desired_capacity = 1
      max_capacity     = 10
      min_capacity     = 1
      instance_type    = "m5.large"
      key_name         = ""
    }
  }
}
```

All the `node_groups` keys could be put in `node_groups_defaults` if creating more than one group.

> Second question: What do you think about forcing this approach? i.e. no direct connection between the core and mng module, i.e. no variables of `variable.node_groups*`.

It's an option. But we'd have to modify the root module for interaction with the `aws-auth` ConfigMap. Need to get any additional worker roles into it and document how to deal with the race condition when first creating the cluster. Possibly exposing an output variable from the `kubernetes_config_map` resource for dependency ordering? Or have a `null_data_source` that depends on `aws-auth` be the source for the `cluster_name` output? Potential for dependency cycle if users not careful.

modules/node_groups/locals.tf
```diff
@@ -0,0 +1,11 @@
+locals {
+  # Trying to use `? var.node_groups : {}` does not work
+  node_groups_keys  = toset(var.create_eks ? keys(var.node_groups) : [])
```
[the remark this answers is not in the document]
dpiddockcmp: I thought it was cleaner to have `for_each = local.node_groups_keys` instead of `for_each = toset(var.create_eks ? keys(var.node_groups) : [])` on every resource. 

Cannot directly use `for_each = var.node_groups` as that doesn't support `var.create_eks`.

Cannot use `for_each = var.create_eks ? var.node_groups : {}` because Terraform gets very confused and throws errors depending on what you have in `var.node_groups`. "Object does not equal Object" or the maps are not the same size.

modules/node_groups/locals.tf
```diff
@@ -0,0 +1,11 @@
...
+  node_groups_count = length(local.node_groups_keys)
```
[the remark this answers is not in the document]
dpiddockcmp: I think that was used by the IAM resource creation. You're right, no longer needed.

modules/node_groups/node_groups.tf
```diff
@@ -0,0 +1,49 @@
+resource "aws_eks_node_group" "workers" {
+  for_each = local.node_groups_keys
```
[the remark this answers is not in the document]
dpiddockcmp: `node_groups_expanded` is a map. `[]` is a list/set. Terraform will throw a type-mismatch error. Can't use `{}` as terraform will throw errors about map size when `node_groups_expanded` has more than 1 item. Or "Object does not equal Object". I saw so many stupid and frustrating errors. Objects are not very ternary friendly.

modules/node_groups/variables.tf
dpiddockcmp: I now fear that this attempt at abusing automatic documentation may actually cause more confusion than it solves 😅 

The values in this default map are rubbish. And will cause errors/issues if actually used. They're to indicate the set of expected keys and their data types. When only used by the parent module this will always be set to an empty map or whatever the user passes in. This default will need removing if we're expecting users to directly use the module. 

Maybe the whole lot should be moved to README.md. Then the extended comments will be more visible too without having to dive into the source code.

variables.tf
dpiddockcmp: I don't really see the advantage to this. It came up in discussions on #644. The output there was a list that had no relation to the order of the input list. This then makes life difficult for users if they want to associate the outputs to the items in their inputs.

Terraform will throw an error about duplicate keys if the name is specified twice. Not the most obvious error.

`for_each` would be much more useful if we weren't worrying about a `create_eks` variable or if Terraform worked better with Objects in ternary statements.

modules/node_groups/locals.tf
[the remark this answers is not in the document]
barryib: I see. Is something like would work ?

```hcl
node_groups_expanded = { for k, v in var.node_groups :  k => merge(
    var.node_groups_defaults,
    v,
  ) if create_eks }
```

I don't test it, but from doc it should generate an empty map if creat_eks is false.

variables.tf
barryib: Agree.

modules/node_groups/node_groups.tf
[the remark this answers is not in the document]
barryib: I see. Terraform is a terrible can be annoying sometimes.

But is something like this in [locals](url) ok for you ?

```hcl
  node_groups_expanded = { for k, v in var.node_groups : k => merge(
    var.node_groups_defaults,
    v,
  ) if var.create_eks }
```
This should create an empty map if `var.create_eks` is set to `false`

modules/node_groups/locals.tf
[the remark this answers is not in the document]
dpiddockcmp: 🤦‍♂ I see what you mean now. I've made changes.

modules/node_groups/variables.tf
dpiddockcmp: Made changes and moved this weird default and the commends into the module's README.

modules/node_groups/node_groups.tf
```diff
@@ -0,0 +1,49 @@
+resource "aws_eks_node_group" "workers" {
+  for_each = random_pet.node_groups
```
barryib: A last question. So now we don't have intermediate locals anymore, can't we just loop over `local.node_groups_expanded` instead of `random_pet.node_groups` ? So we can use `each.value` (the current node group) in the ressource defintion ? I think it will be more readable, will avoid using split and different variable name for the same thing.

Furthermore, we only need the random pet for the `node_group_name`, so we can have something like this:

```hcl
node_group_name = join("-", [var.cluster_name, each.key, random_pet.node_groups[each.key].id])
```

@dpiddockcmp what do you think ?

modules/node_groups/README.md
```diff
@@ -0,0 +1,55 @@
...
+`iam_role_arn` must be specified in either `var.node_groups_defaults` or `var.node_groups` if the default parent IAM role is not being created for whatever reason, for example if `manage_worker_iam_resources` is set to false in the parent.
+
+## `node_groups` and `node_groups_defaults` keys
+`node_groups_defaults` is a map that can take the below keys. Values will be used if not specified in individual node groups.
+
+`node_groups` is a map of maps. Key of first level will be used as unique value for `for_each` resources and in the `aws_eks_node_group` name. Inner map can take the below values.
+
+| Name | Description | Type | If unset |
+|------|-------------|:----:|:-----:|
+| additional\_tags | Additional tags to apply to node group | map(string) | Only `var.tags` applied |
```
barryib: Great. I think it's better this way with all needed variable in the readme.

modules/node_groups/node_groups.tf
```diff
@@ -0,0 +1,49 @@
...
+      source_security_group_ids = compact(split("|", each.value.keepers.source_security_group_ids))
+    }] : []
+
+    content {
+      ec2_ssh_key               = remote_access.value["ec2_ssh_key"]
+      source_security_group_ids = remote_access.value["source_security_group_ids"]
+    }
+  }
+
+  version = var.cluster_version
```
eytanhanig: Allowing a user to set a node group's value means that they can incrementally replace their node groups as part of an upgrade process rather than having every single node group replaced simultaneously, which is what this currently does. I strongly recommend that instead we do something along the lines of the following (some tweaking may be required):
```suggestion
  version = lookup(var.node_groups[each.value], "version", null)
```

Note that we're able to use `null` since [the default is for node groups to match the EKS cluster version](https://www.terraform.io/docs/providers/aws/r/eks_node_group.html#version):
> version – (Optional) Kubernetes version. Defaults to EKS Cluster Kubernetes version. Terraform will only perform drift detection if a configuration value is provided.

modules/node_groups/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+      id        = v.id
+      status    = v.status
+      resources = v.resources
+    }
+  }
+}
+
+output "aws_auth_roles" {
+  description = "Roles for use in aws_auth ConfigMap"
+  value = [
```
eytanhanig: Without some sort of ordering function you're going to hit consistency problems when attempting to convert an unordered map into a list. Is there any reason _not_ to use a map instead of this?

modules/node_groups/outputs.tf
```diff
@@ -0,0 +1,25 @@
+output "node_groups" {
+  description = "Outputs from EKS node groups. Map of maps, keyed by var.node_groups keys"
+  value = {
```
eytanhanig: Keep it simple, keep it versatile:
```suggestion
  value = aws_eks_node_group.workers
```

modules/node_groups/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Node Groups' IAM Role
+The role ARN specified in `var.default_iam_role_arn` will be used by default. In a simple configuration this will be the worker role created by the parent module.
+
+`iam_role_arn` must be specified in either `var.node_groups_defaults` or `var.node_groups` if the default parent IAM role is not being created for whatever reason, for example if `manage_worker_iam_resources` is set to false in the parent.
+
+## `node_groups` and `node_groups_defaults` keys
+`node_groups_defaults` is a map that can take the below keys. Values will be used if not specified in individual node groups.
+
+`node_groups` is a map of maps. Key of first level will be used as unique value for `for_each` resources and in the `aws_eks_node_group` name. Inner map can take the below values.
```
eytanhanig: I have mixed feelings about this: On the one hand it more closely mirrors the output they'll be given, on the other it is a lot more confusing than just having a list of maps with the note that "`base_name` is a required value and must be unique for each node group". What are your thoughts?

modules/node_groups/node_groups.tf
eytanhanig: I agree with @barryib. Terraform is less error-prone when you share a loop index between resources. In fact the simplest solution is to just use
```suggestion
  for_each = var.node_groups
```

aws_auth.tf
```diff
@@ -43,13 +43,10 @@ data "template_file" "worker_role_arns" {
 }
 
 data "template_file" "node_group_arns" {
-  count    = var.create_eks ? local.worker_group_managed_node_group_count : 0
+  count    = var.create_eks ? length(module.node_groups.aws_auth_roles) : 0
```
eytanhanig: Perhaps use a `for_each` loop here instead of count?

modules/node_groups/node_groups.tf
dpiddockcmp: I agree that the join and split is ... not great. But `random_pet.keepers` can only take a plain map(string). This is a limitation from terraform's provider API which does not support dynamic typing.

We need to pass in all those variables to the `random_pet` because they trigger a recreation on the `aws_eks_node_group` resource.

I then used random_pet.keepers in the aws_eks_node_group to reduce duplication of logic. For example on the instance_type:

```HCL
resource "random_pet" "node_groups" {
  for_each = local.node_groups_expanded
  keepers = {
    instance_type = lookup(each.value, "instance_type", var.workers_group_defaults["instance_type"])
  }
}

resource "aws_eks_node_group" "workers" {
  for_each = random_pet.node_groups
  instance_types = [each.value.keepers.instance_type]
}
```

Any change to the logic on random_pet is automatically received by the aws_eks_node_group. I did this as I made mistakes when migrating the old code which had duplication of logic.

We can go back to duplication of logic if you want but it will need careful review of any PRs in future that touch either of those resources 🤷‍♂ 

`labels` and `tags` refer to `var.node_groups_defaults` and `var.node_groups[each.key]` as terraform does not do a deep merge of maps.

modules/node_groups/outputs.tf
dpiddockcmp: Terraform processes the keys in a map in lexical ordering. There aren't consistency problems here.

This is the list of `vars` maps given to the `worker-role.tpl` template file in `aws_auth.tf`. The API is set by that code instead of needing a conversion elsewhere.

I did experiment with generating the aws-auth snippet directly in this module but it felt wrong. I didn't like the use of string concatenation. I think improvements could be made to how the aws-auth ConfigMap is generated but that's out of scope for this PR.

aws_auth.tf
dpiddockcmp: Then you're just pushing the flattening of the map to further down the chain. Currently it's consistent with how other worker resources are generated.

I feel the whole aws-auth file generation could do with improvement but it's out of scope for this PR.

modules/node_groups/README.md
dpiddockcmp: Terraform often does stupid things with lists. You can see it when changing the length of the `workers_groups` list and suddenly all the workers are scheduled for re-creation.

We should be going the other way and converting old inputs from lists to maps with for_each.

Having this as a map makes it pretty obvious what the key on created resources will be. It's a much more natural flow than converting from a list to a map using a subkey.

modules/node_groups/node_groups.tf
dpiddockcmp: I agree. This is existing logic. I'll change it to be user configurable.

modules/node_groups/node_groups.tf
```diff
@@ -0,0 +1,49 @@
...
+      source_security_group_ids = remote_access.value["source_security_group_ids"]
+    }
+  }
+
+  version = lookup(local.node_groups_expanded[each.key], "version", null)
+
+  labels = merge(
+    lookup(var.node_groups_defaults, "k8s_labels", {}),
+    lookup(var.node_groups[each.key], "k8s_labels", {})
+  )
```
eytanhanig: I believe that you can replace this entire block with something like the following:
```suggestion
  labels = lookup(local.node_groups_expanded, "k8s_labels", null)
```

You can combine lines 42 and 43 below by doing something similar.

modules/node_groups/node_groups.tf
eytanhanig: If looping over `var.node_groups` you can still reference `random_pet.node_groups`, simply do the following replacement inside the resource block:
`each.value` => `random_pet.node_groups[each.key]`

Better yet, if you loop over local `local.node_groups_expanded` instead - which will make it more consistent with random_pet's index - you can do the following replacement as well:
`local.node_groups_expanded[each.key]` => `each.value`

modules/node_groups/node_groups.tf
```diff
@@ -0,0 +1,49 @@
...
+
+    content {
+      ec2_ssh_key               = remote_access.value["ec2_ssh_key"]
+      source_security_group_ids = remote_access.value["source_security_group_ids"]
+    }
+  }
+
+  version = lookup(local.node_groups_expanded[each.key], "version", null)
+
+  labels = merge(
```
eytanhanig: I believe that you can replace this entire block with something like the following:
```suggestion
  labels = lookup(local.node_groups_expanded, "k8s_labels", null)
```

You can combine lines 42 and 43 below by doing something similar.

dpiddockcmp: Those are functionally two different things as Terraform does not do deep merging of lists:
```
> merge({a = {a = "a"}}, {a = {aa = "aa"}})
{
  "a" = {
    "aa" = "aa"
  }
}
```

I thought the workers merged the default tags with the per workers tags but on double checking the code it's a simple fallback: [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/c17d17e2c251ce9f57fcab6a6822ca138f01f035/workers.tf#L132..L135). Although that's harder as they are lists of maps and merging there isn't easy.

modules/node_groups/node_groups.tf
dpiddockcmp: There are currently 8 references to `each.value.keepers` and only 5 references to `local.node_groups_expanded`. Switching to the local for the `for_each` would make the code more verbose. 

The random provider also makes reference to "reading" variables through the keepers map in the example to ensure they change together: [docs](https://www.terraform.io/docs/providers/random/index.html#resource-quot-keepers-quot-)

modules/node_groups/node_groups.tf
barryib: Right.

modules/node_groups/node_groups.tf
barryib: I think we can resolve this conversation to let @max-rocket-internet ship a new release and get feedbacks from users. This conversation is more a coding style discussion and we can change it easily if needed in a next release.

barryib:
@dpiddockcmp please rebase master.

modules/node_groups/node_groups.tf
dpiddockcmp: Had a thought. I've moved the `var.workers_group_defaults` merging into the `local.node_groups_expanded` map. Allowed to greatly simplify the logic in `random_pet`. And thus no need to use it as a pass through.

barryib: Love it !

max-rocket-internet:
OK are we good to merge?

node_groups.tf
```diff
@@ -1,112 +1,29 @@
...
+    cluster_name = var.cluster_name
+
+    # Ensure these resources are created before "unlocking" the data source.
+    # `depends_on` causes a refresh on every run so is useless here.
+    # [Re]creating or removing these resources will trigger recreation of Node Group resources
+    aws_auth         = coalescelist(kubernetes_config_map.aws_auth[*].id, [""])[0]
+    role_NodePolicy  = coalescelist(aws_iam_role_policy_attachment.workers_AmazonEKSWorkerNodePolicy[*].id, [""])[0]
+    role_CNI_Policy  = coalescelist(aws_iam_role_policy_attachment.workers_AmazonEKS_CNI_Policy[*].id, [""])[0]
+    role_Container   = coalescelist(aws_iam_role_policy_attachment.workers_AmazonEC2ContainerRegistryReadOnly[*].id, [""])[0]
+    role_auotscaling = coalescelist(aws_iam_role_policy_attachment.workers_autoscaling[*].id, [""])[0]
```
pierresteiner: Typo: should be role_autoscaling I guess

dpiddockcmp: Fixed, thank you.

barryib:
@max-rocket-internet Yes ... 🚀

node_groups.tf
```diff
@@ -1,112 +1,29 @@
...
+  inputs = {
+    cluster_name = var.cluster_name
+
+    # Ensure these resources are created before "unlocking" the data source.
+    # `depends_on` causes a refresh on every run so is useless here.
+    # [Re]creating or removing these resources will trigger recreation of Node Group resources
+    aws_auth         = coalescelist(kubernetes_config_map.aws_auth[*].id, [""])[0]
+    role_NodePolicy  = coalescelist(aws_iam_role_policy_attachment.workers_AmazonEKSWorkerNodePolicy[*].id, [""])[0]
+    role_CNI_Policy  = coalescelist(aws_iam_role_policy_attachment.workers_AmazonEKS_CNI_Policy[*].id, [""])[0]
+    role_Container   = coalescelist(aws_iam_role_policy_attachment.workers_AmazonEC2ContainerRegistryReadOnly[*].id, [""])[0]
```
barryib: Just got a thought, should we create distincts IAM policies for each submodules ? This will make them totally self contained from each others.

@max-rocket-internet @dpiddockcmp any thoughts ?

dpiddockcmp: I don't think so by default. Everything can use the same single worker role between all workers. AWS seems to be pushing towards this role being as basic as possible with pods using OIDC where necessary. For example see the recently updated [blog post about ALB ingress](https://aws.amazon.com/blogs/opensource/kubernetes-ingress-aws-alb-ingress-controller/)

If users really need different IAM roles per worker group they can always override in one of the maps.

barryib: In a near future we will split the terraform-aws-eks module into multiple submodules. Probably like this:
- `modules/node_groups` (the current PR)
- `modules/fargate_profile`
- `modules/workers`
- `modules/cluster` ??? (this can stay in the root module)

The root module will probably only create the cluster. So I was thinking to move `
aws_iam_role_policy_attachment.*` into each submodules and with their own iam_role. In that way, user can provide aditionnal policy to a particular module. I see this already in cloudpose modules [aws-eks-workers](https://github.com/cloudposse/terraform-aws-eks-workers/blob/master/main.tf#L38-L73), [aws-eks-node-group](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/master/main.tf#L35-L64), etc.

In other words, my thoughts were: should we split IAM role and profile per submodule in the near future or not ? If yes, should we start now with the node_groups submodule ?

max-rocket-internet: TBH I am with @barryib on this. Let's aim for as much separation as possible.

> In a near future we will split the terraform-aws-eks module into multiple submodules

Well, let's see how PR/module goes first 😅

node_groups.tf
```diff
@@ -1,112 +1,29 @@
...
-
-  tags = lookup(each.value, "node_group_additional_tags", null)
-
-  lifecycle {
-    create_before_destroy = true
-  }
+module "node_groups" {
+  source                 = "./modules/node_groups"
+  create_eks             = var.create_eks
+  cluster_name           = coalescelist(data.null_data_source.node_groups[*].outputs["cluster_name"], [""])[0]
```
max-rocket-internet: Can't this just be `var.cluster_name`?

node_groups.tf
max-rocket-internet: Ahh but then the IAM roles need to be passed back out of the NG module to be used in `aws-auth` configmap, which might create a circular dependency?

barryib: > Ahh but then the IAM roles need to be passed back out of the NG module to be used in aws-auth configmap, which might create a circular dependency?

Hummm... I don't think so. Terraform will handle the dependency graph. The aws-auth templating will only wait for the IAM role creation and not for the whole module.

dpiddockcmp: Why do we need all that code duplication?

If you want to use cloudposse modules why not use them instead of slowly migrating this one to being a copy? It makes slightly more sense for them as they're totally separate module repos.

This module was a one shot for setting up everything EKS. From that point it makes more sense to only have the one worker role.

node_groups.tf
dpiddockcmp: This sets up a dependency on the `kubernetes_config_map.aws_auth` resource to avoid a race condition.

node_groups.tf
barryib: @dpiddockcmp The idea here is not mimicking those modules, but discussing how we should migrate slowly from on main modules to a multiple submodules with all his pros and cons.

> It makes slightly more sense for them as they're totally separate module repos.

Make sense.

Let merge this PR and get feedbacks from users. We'll improve it if needed.

> Well, let's see how PR/module goes first 😅

@max-rocket-internet I know we are not in the hurry, but if we start splitting this module, we have to continue with other features (fargate and actual workers).

Thanks all for your time for discussions and work.

max-rocket-internet: > Let merge this PR and get feedbacks from users. We'll improve it if needed.

Agreed!

max-rocket-internet:
I'll merge this now, run some test and then create a new release.

PR #651, opened 2019-12-30
[conceding to 653] add output for aws_iam_openid_connect_provider arn

jonathancolby-olx:
# PR o'clock

## Description

The OIDC provider ARN is required for the assume role policy as the "Federated" principal of the iam-for-pods role.  This change adds the OIDC provider ARN as a module output.

usage example:

```
data "aws_iam_policy_document" "this" {
  statement {
    actions = ["sts:AssumeRoleWithWebIdentity"]

    principals {
      type        = "Federated"
      identifiers = [module.eks.aws_iam_openid_connect_provider_arn]
    }
```

Please explain the changes you made here and link to any relevant issues.

### Checklist

- [X] Change added to CHANGELOG.md. All changes must be added and breaking changes and highlighted
- [X] CI tests are passing
- [X] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #653, opened 2020-01-01
Add the OIDC provider ARN to outputs

eytanhanig:
# PR o'clock

## Description

Please explain the [changes](url) you made here and link to any relevant issues.

This PR adds the OICD Provider ARN to outputs when `enable_irsa = true`. This needed when creating IAM policy documents/roles for Service Accounts, an example of which can be seen [here](https://www.terraform.io/docs/providers/aws/r/eks_cluster.html#enabling-iam-roles-for-service-accounts).

Note: The output is conditionally generated only when `enable_irsa = true`. When `enable_irsa = false` no output will be given, which [is preferable](https://github.com/hashicorp/terraform/issues/23222#issuecomment-547421898) to it returning a bad value, such as the empty string, which could then lead to misconfigured downstream resources.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -63,6 +63,11 @@ output "kubeconfig_filename" {
   value       = concat(local_file.kubeconfig.*.filename, [""])[0]
 }
 
+output "oidc_provider_arn" {
+  description = "The ARN of the OIDC Provider if `enable_irsa = true`."
+  value       = var.enable_irsa ? aws_iam_openid_connect_provider.oidc_provider[0].arn : null
+}
+
 output "workers_asg_arns" {
   description = "IDs of the autoscaling groups containing workers."
   value = concat(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Fix aws-auth config map for managed node groups (by @wbertelsen)
 - Added support to create IAM OpenID Connect Identity Provider to enable EKS Identity Roles for Service Accounts (IRSA). (by @alaa)
 - Adding node group iam role arns to outputs. (by @mukgupta)
+- Added the OIDC Provider ARN to outputs. (by @eytanhanig)
 - **Breaking:** Change logic of security group whitelisting. Will always whitelist worker security group on control plane security group either provide one or create new one. See Important notes below for upgrade notes (by @ryanooi)
 
 #### Important notes
--- a/README.md
+++ b/README.md
@@ -219,6 +219,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | kubeconfig | kubectl config file contents for this EKS cluster. |
 | kubeconfig\_filename | The filename of the generated kubectl config. |
 | node\_groups\_iam\_role\_arns | IAM role ARNs for EKS node groups |
+| oidc\_provider\_arn | The ARN of the OIDC Provider if `enable_irsa = true`. |
 | worker\_autoscaling\_policy\_arn | ARN of the worker autoscaling IAM policy if `manage_worker_autoscaling_policy = true` |
 | worker\_autoscaling\_policy\_name | Name of the worker autoscaling IAM policy if `manage_worker_autoscaling_policy = true` |
 | worker\_iam\_instance\_profile\_arns | default IAM instance profile ARN for EKS worker groups |


jonathancolby-olx:
hi. did you see this ?  https://github.com/terraform-aws-modules/terraform-aws-eks/pull/651

outputs.tf
```diff
@@ -63,6 +63,11 @@ output "kubeconfig_filename" {
   value       = concat(local_file.kubeconfig.*.filename, [""])[0]
 }
 
+output "oidc_provider_arn" {
+  description = "The ARN of the OIDC Provider if `enable_irsa = true`."
+  value       = var.enable_irsa ? aws_iam_openid_connect_provider.oidc_provider : null
```
dpiddockcmp: This isn't returning just the ARN. It's returning a list of `aws_iam_openid_connect_provider` outputs.

For example:
```
oidc = [
  {
    "arn" = "arn:aws:iam::111111111111:oidc-provider/oidc.eks.us-west-2.amazonaws.com/id/518F3C3FCEEB9B22064F6EE41A9BDC39"
    "client_id_list" = [
      "sts.amazonaws.com",
    ]
    "id" = "arn:aws:iam::111111111111:oidc-provider/oidc.eks.us-west-2.amazonaws.com/id/518F3C3FCEEB9B22064F6EE41A9BDC39"
    "thumbprint_list" = [
      "9e99a48a9960b14926bb7f3b02e22da2b0ab7280",
    ]
    "url" = "oidc.eks.us-west-2.amazonaws.com/id/518F3C3FCEEB9B22064F6EE41A9BDC39"
  },
]
```

```suggestion
  value       = var.enable_irsa ? aws_iam_openid_connect_provider.oidc_provider[0].arn : null
```

Tried with 0.12.9 and 0.12.18. Appears to work correctly. Having indexes in outputs scare me a little due to historical issues but it seems the lazy evaluation of the ternary operator is working now. Also tried manually deleting the resource, and running destroy on an empty state with `enable_irsa = true`.

Setting an output to `null` causes it to not be listed in the state's output. This may cause issues for people using the `terraform_remote_state` data source to chain configs together.

eytanhanig: Good catch and suggested change, I'll push an update. I'm also not a fan of how using the ternary operator plus count forces you to refer to an index and have been pondering whether switching to a for_each makes more sense, however the place to make that switch is where the resource is being created. Do you think that is the smarter way to do it?

Setting up `terraform_remote_state` will work regardless of whether `enable_irsa` is true or false, however anything referencing this output _will_ fail when it is set to false. I believe that **this is desirable behavior** since an early Terraform error message is preferable to a resource being created with bad data downstream. Say for example that we use the empty string when `enable_irsa = false` and the user attempts to deploy a Kubernetes manifest with this value. The user would likely see no errors in Terraform and end up wasting a bunch of time trying to debug it as a Kubernetes problem.

eytanhanig:
> hi. did you see this ? #651
> 
> I personally like your usage of `var.enable_irsa` better. i can close mine.

@ jonathancolby-olx Apparently I missed it. I'm OK closing my PR as well. Let me know which you'd prefer.

> what do you think about naming the output `cluster_oidc_connect_provider_arn` to make it appear next to `cluster_oidc_issuer_url`

I've been pondering the same thing. The issue is that nearly every output starting with `cluster_` - including `cluster_oidc_issuer_url` - is referring to attributes associated with and output from the actual `aws_eks_cluster` resource. It ultimately comes down to two things:

1. Is the fact that it's related to the cluster implicit given what this module is?
2. If it isn't implicit, then would we rather have clarity (starts with `cluster_`), or precision (leave as above since it's not output from the `aws_eks_cluster` resource)?

jonathancolby-olx:
> > hi. did you see this ? #651
> > I personally like your usage of `var.enable_irsa` better. i can close mine.
> 
> @ jonathancolby-olx Apparently I missed it. I'm OK closing my PR as well. Let me know which you'd prefer.
> 

@eytanhanig - Let's go with yours.  Your case for using conditional with `enable_irsa` is completely valid.   I will close mine.

> > what do you think about naming the output `cluster_oidc_connect_provider_arn` to make it appear next to `cluster_oidc_issuer_url`
> 
> I've been pondering the same thing. The issue is that nearly every output starting with `cluster_` - including `cluster_oidc_issuer_url` - is referring to attributes associated with and output from the actual `aws_eks_cluster` resource. It ultimately comes down to two things:
> 
> 1. Is the fact that it's related to the cluster implicit given what this module is?
> 2. If it isn't implicit, then would we rather have clarity (starts with `cluster_`), or precision (leave as above since it's not output from the `aws_eks_cluster` resource)?

@eytanhanig - after thinking about this, I also tend to agree that the `cluster_` prefix is redundant.  I'm OK with how you did it too.

PR #679, opened 2020-01-13
Secure kubeconfig permissions

szczad:
# PR o'clock

## Description

Adjustable Kubeconfig file permissions

### Checklist

- [x] CI tests are passing

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -37,6 +37,12 @@ variable "config_output_path" {
   default     = "./"
 }
 
+variable "config_file_permissions" {
+  description = "File mode of the Kubectl config file. For security concerns this defaults to 0600."
+  type        = string
+  default     = "0600"
+}
+
 variable "write_kubeconfig" {
   description = "Whether to write a Kubectl config file containing the cluster configuration. Saved to `config_output_path`."
   type        = bool
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - Write your awesome change here (by @you)
 - Add support for restricting access to the public API endpoint (@sidprak)
+- Add support to adjust permissions of kubeconfig file (@szczad)
 
 # History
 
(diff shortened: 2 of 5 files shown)


variables.tf
```diff
@@ -37,6 +37,12 @@ variable "config_output_path" {
   default     = "./"
 }
 
+variable "config_file_permissions" {
+  description = "File mode of the Kubectl config file. For security concerns this defaults to 0600."
```
max-rocket-internet: What security concerns? There's nothing sensitive in this file?

szczad: That's option if:
* Someone wants to be paranoid
* Someone does not want to share details with other users
* Someone modifies his cluster and adds more sensitive data to it (i.e. someone puts his key in it - `client-key-data`)

Moreover, basic `~/.kube/config` file created by `kubectl` also does not contain any sensitive information until you add some. So to follow best practices let's assume generated file is safe enough for everyone just by design.

max-rocket-internet:
Is this likely to break something for someone now that it defaults to a very tight permission? Is this needless complexity? I'm not sure. @barryib what do you think?

variables.tf
```diff
@@ -37,6 +37,12 @@ variable "config_output_path" {
...
+  type        = string
+  default     = "0600"
```
barryib: I'll set default to `null` in that way we use the default permission set by the local_file resource https://www.terraform.io/docs/providers/local/r/file.html#file_permission. Furthermore we don't change the current module's behavior => avoid breaking something.

barryib:
> Is this likely to break something for someone now that it defaults to a very tight permission? Is this needless complexity? I'm not sure. @barryib what do you think?

I think we can use null instead of `600` as default https://github.com/terraform-aws-modules/terraform-aws-eks/pull/679/files#r368041789

variables.tf
```diff
@@ -37,6 +37,12 @@ variable "config_output_path" {
   default     = "./"
 }
 
+variable "config_file_permissions" {
```
barryib: `kubeconfig_file_permission` as variable name.

variables.tf
barryib: @szczad do you need this or you think some will ?

barryib:
@szczad is this PR still needed ? Are you still working on this ?

barryib:
@szczad closing this PR. Feel free to reopen it when you'll get a chance to work on it.

Issue #680, opened 2020-01-13
Not working on Windows


maganuk:
Hi,

I saw that there was a PR added for Windows Compatibility, however in cluster.tf even this line should have the ability to specify a custom interpretor

    provisioner "local-exec" {
        command = <<EOT
        until curl -k -s ${aws_eks_cluster.this[0].endpoint}/healthz >/dev/null; do sleep 4; done
      EOT
    }

Without that, its not possible to create a cluster

barryib:
@maganuk Thanks for opening this issue. There was a long discussion about this on https://github.com/terraform-aws-modules/terraform-aws-eks/pull/639 and called for help to windows users to help is test this.

Could you please take a time on this issue and open a PR to fix this ? I can help you with this.

Do you have a working curl (or equivalent) command on Windows to do a HTTP GET on the healthz endpoint ?

maganuk:
I need a day or two to figure this out. Curl is not installed by default on Windows and it would be odd to require the user to install any third party packages. Lets try and figure out if we can do this using terraform. Usually the delay will not be more that 8-10 seconds. Can we not delay the execution of the next step by a few seconds, instead of pinging the url to see if its alive?

barryib:
Actually, there is `Invoke-RestMethod` in powershell https://discoposse.com/2012/06/30/powershell-invoke-restmethod-putting-the-curl-in-your-shell/. Maybe we can use that ?

> Usually the delay will not be more that 8-10 seconds.

I think we don't really know how long it'll take. So I'll really prefer pinging until it's alive rather than waiting and hoping that it'll take less than 10 seconds.

sanjeevgiri:
I have a potential fix for this, however I am not able to create a branch for pull request. Is there anyway something I need to do to be able to create a branch?

dpiddockcmp:
@sanjeevgiri 
> I have a potential fix for this, however I am not able to create a branch for pull request. Is there anyway something I need to do to be able to create a branch?

You need to Fork the repo to your own account using the buttons in the github interface, push your changes there and then open a PR. See [github's docs](https://guides.github.com/activities/forking/) for the full details

barryib:
Closing this as the wait command is now configurable.

avoidik:
wouldn't it be better to make `interpreter` item configurable?

PR #701, opened 2020-01-20
Configurable local exec command for waiting until cluster is healthy

sanjeevgiri:
# PR o'clock

## Description

Currently executing this module for windows fails due to local-exec command that waits until the cluster is healthy. The command being used `until curl -k -s %CLUSTER_HEALTH_ENDPOINT% >/dev/null; do sleep 4; done` only works in *nix systems. We could default to using this command, however, it would be great if non *nix users would have to option to specify custom command that would achieve the same. (Even better would be  terraform specific resource that would allow us to wait until a http url becomes available :)).

This PR attempts to add the ability to define os-specific commands to wait for a healthy cluster to be available.

### Checklist

- [ ] CI tests are passing

#680


sanjeevgiri:
Not ready yet. Will re-open soon.

sanjeevgiri:
@maganuk  @barryib @dpiddockcmp @RothAndrew . Would you guys be kind enough to look at his and shed some insight on how I've approached it? I am open to suggestions and discussions that drive to a meaningful solution.

maganuk:
Hey, I don't know whats changed in v8 but v7 use to work just fine without waiting for the endpoint to be active. Maybe because I was doing my own Aws auth, and there is an option to skip Aws auth. For such cases, can we not make this check optional?

sanjeevgiri:
> Hey, I don't know whats changed in v8 but v7 use to work just fine without waiting for the endpoint to be active. Maybe because I was doing my own Aws auth, and there is an option to skip Aws auth. For such cases, can we not make this check optional?

local-exec was removed 

> Hey, I don't know whats changed in v8 but v7 use to work just fine without waiting for the endpoint to be active. Maybe because I was doing my own Aws auth, and there is an option to skip Aws auth. For such cases, can we not make this check optional?

It looks like aws_auth.tf had local-exec to update kube config map if manage_aws_auth was set. This local-exec has been removed in this revision 93636625740c63fd89ad8bc60ad180761288c54d. @stijndehaes can elaborate a little on that.

The check for healthy cluster was added by @shaunc as a part of https://github.com/terraform-aws-modules/terraform-aws-eks/pull/639. They seem to be related though. 

@stijndehaes @shaunc should be able to shed some light on this? Thanks in advance.

shaunc:
We did try terraform specific (like using http resource for health check) -- couldn't get to work. I think the maintainers decided that windows probably had other things broken as well; but certainly PR that allowed customizing command would seem reasonable to me (I am not a maintainer). I started using after the switch from local_exec to kubernetes, so I can't speak to that except that the maintainers do want to shift to terraform-supported if possible. (Just we couldn't figure out how to wait for cluster using only terraform.)

maganuk:
Since we’re only using this for Manage_AWS_Auth, can we make the local_exec
conditional based on the value of Manage_AWS_Auth?

shaunc:
You'll have to get a maintainer to chime in. IMO terraform-built resources are generally supposed to be "ready to use" when apply completes, so they can be used as part of larger builds. From that perspective, waiting for the cluster to be up should be the "normal case", irrespective of what it is used for internally. But having some option to change (or possibly skip) the health check doesn't seem unreasonable.

max-rocket-internet:
This approach with the `%CLUSTER_HEALTH_ENDPOINT%` and so on is quite complicated. Could we just use `environment` on the `local-exec` and put the endpoint in there? then maybe something like this:

```hcl
provisioner "local-exec" {
  command = <<EOT
  ${var.wait_for_cluster_cmd == "" ? "until curl -k -s $${ENDPOINT}/healthz >/dev/null; do sleep 4; done" : var.wait_for_cluster_cmd}
EOT
}
```

> Since we’re only using this for Manage_AWS_Auth, can we make the local_exec conditional based on the value of Manage_AWS_Auth?

That's also a good idea.

sanjeevgiri:
@max-rocket-internet let me explore your advice. :) Thanks for chiming in.

- I am not entirely sure if using dynamically setting ENDPOINT as an environment variable is possible. I checked the docs for local-exec, and it seems like it has a environment parameter. Perhaps that can be leveraged to some extent. The catch is that the environment variable needs to be tied to url that is dynamically generated.

- For conditionally executing the local-exec: (I may be too much of a newbie here :) ), is it possible to use count within the local-exec block or would it need to be externalized into a different resource? I will research myself while I keep you guys confused with my newbie questions heh.

sanjeevgiri:
> * For conditionally executing the local-exec: (I may be too much of a newbie here :) ), is it possible to use count within the local-exec block or would it need to be externalized into a different resource? I will research myself while I keep you guys confused with my newbie questions heh.

- On further reading, looks like count can be used with local-exec (meta-arguments, I will try that)

cluster.tf
```diff
@@ -32,10 +32,18 @@ resource "aws_eks_cluster" "this" {
...
+resource "null_resource" "wait_for_cluster" {
+  depends_on = [
+    aws_eks_cluster.this[0]
+  ]
+
   provisioner "local-exec" {
-    command = <<EOT
-    until curl -k -s ${aws_eks_cluster.this[0].endpoint}/healthz >/dev/null; do sleep 4; done
-  EOT
+    command = var.manage_aws_auth ? var.wait_for_cluster_cmd : ""
```
sanjeevgiri: @max-rocket-internet this would optionally perform a no-op if manage_aws_auth is disabled? I may have to use count, but was unsure on wiring dependencies on the kubernetes_config_map resource.

cluster.tf
```diff
@@ -32,10 +32,18 @@ resource "aws_eks_cluster" "this" {
...
+    environment = {
+      ENDPOINT = aws_eks_cluster.this[0].endpoint
```
sanjeevgiri: I believe this is close to what you were expressing? Not sure :) but it does seem to work.

max-rocket-internet: Yes. This is much cleaner IMO 😃

cluster.tf
max-rocket-internet: Like it but have you tested this in both situations? No errors or issues with having a command of `""`

variables.tf
```diff
@@ -198,10 +198,10 @@ variable "cluster_delete_timeout" {
   default     = "15m"
 }
 
-variable "local_exec_interpreter" {
-  description = "Command to run for local-exec resources. Must be a shell-style interpreter. If you are on Windows Git Bash is a good choice."
-  type        = list(string)
-  default     = ["/bin/sh", "-c"]
+variable "wait_for_cluster_cmd" {
+  description = "Custom local-exec command to execute for determining if the eks cluster is healthy. Cluster endpoint will be available as an ENDPOINT environment variable"
```
max-rocket-internet: Small thing but can you change this to `will be available as an environment variable called ENDPOINT`? Otherwise it sounds like "ENDPOINT" is a special type of env var or something.

cluster.tf
max-rocket-internet: I get this error if `manage_aws_auth = false`:

```
Error: local-exec provisioner command must be a non-empty string
```

So I think you need to do this:

```hcl
resource "null_resource" "wait_for_cluster" {
  count      = var.manage_aws_auth ? 1 : 0

  depends_on = [
    aws_eks_cluster.this[0]
  ]

  provisioner "local-exec" {
    command = var.wait_for_cluster_cmd
    environment = {
      ENDPOINT = aws_eks_cluster.this[0].endpoint
    }
  }
}
```

sanjeevgiri: This is what I wanted to in my initial attempt, but I was not sure how this would be included as a dependency in the aws_auth module. I will try this and test without aws_auth. Thanks again.

sanjeevgiri:
@max-rocket-internet I believe I have addressed all your concerns.

sanjeevgiri:
Anyone has any requests for me? I believe I have addressed the changes requested so far. Can someone help me with doc linter issue. It seems like its failing for everyone.

max-rocket-internet:
I tested with basic example with `manage_aws_auth` set to true and false and works well for me.

variables.tf
sveerabathini: @sanjeevgiri @max-rocket-internet , i am using the code but still i get an error "Error: Error running command 'until curl -k -s $ENDPOINT/healthz >/dev/null; do sleep 4; done': exit status 1. Output: The system cannot find the path specified." unable to get through and create configmap. can any one of you please help me

sanjeevgiri: Hello @sveerabathini, there was a another PR that included a variable local_exec_interpreter. This interpreter uses /bin/sh as the default interpreter. If you are running windows it is likely you will not have a shell installed in that path. If you set the default interpreter value to empty list ([]), then terraform should be able to find a suitable interpreter to use based on the client OS and version (likely power shell for windows). 

I am using git bash, so I added the git bash bin to my windows path environment variable, set the wait_for_cluster_interpreter to []. 

Hope this helps.

PR #710, opened 2020-01-24
Add example for IRSA and cluster-autoscaler

max-rocket-internet:
# PR o'clock

## Description

Many people have requested an example using IRSA (IAM Roles for Service Accounts). So here I am adding an example for the cluster-autoscaler.

### Checklist

- [ ] CI tests are passing


examples/irsa/main.tf
```diff
@@ -0,0 +1,141 @@
...
+    to_port   = 22
+    protocol  = "tcp"
+
+    cidr_blocks = [
+      "10.0.0.0/8",
+      "172.16.0.0/12",
+      "192.168.0.0/16",
+    ]
+  }
+}
```
dpiddockcmp: Are these security groups and random_string needed? They seem to be unused and add noise to the example.

examples/irsa/README.md
```diff
@@ -0,0 +1,70 @@
...
+NAME                                       STATUS   ROLES    AGE     VERSION
+ip-10-0-2-190.us-west-2.compute.internal   Ready    <none>   6m39s   v1.14.8-eks-b8860f
+```
+
+Install Helm:
+
+```
+kubectl -n kube-system create serviceaccount tiller
+kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller
+helm init --service-account=tiller
```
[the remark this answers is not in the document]
barryib: Maybe we can omit tiller installation. And add note that we must have a working helm installation. Because, it's not relevant here.

I personally use terraform with helm provider to perform cluster-autoscaler installation. But it doen't matter how we install it. What we need for the serviceaccount is here https://github.com/terraform-aws-modules/terraform-aws-eks/pull/710/files#diff-5d3e96c50e09c013f8553243cb0be4d6R3-R6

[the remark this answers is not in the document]
max-rocket-internet: I agree 100%. I just haven't used Helm 3 yet 😢

I'll redo next week with Helm 3. I need to it this anyway!

examples/irsa/main.tf
max-rocket-internet: Good point. I just copied the basic example. I'll remove them.

barryib: Agree !

max-rocket-internet:
@barryib do you know why the `terraform_docs` check is failing? I don't get it...

examples/irsa/README.md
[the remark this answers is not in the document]
max-rocket-internet: > Maybe we can omit tiller installation. 

OK good input. I'll remove that part. But I'll keep the `helm install` command for the chart. Is it same for Helm 3?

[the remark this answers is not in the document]
barryib: yep.

barryib:
> @barryib do you know why the `terraform_docs` check is failing? I don't get it...

Nope. But even the `terraform fmt` is failing.

barryib:
@max-rocket-internet  `terraform-doc` 0.8.0 is now in stable. We should probably update terraform pre-commit https://github.com/antonbabenko/pre-commit-terraform/pull/85.

`pre-commit autoupdate` would probably help.

max-rocket-internet:
> We should probably update terraform pre-commit antonbabenko/pre-commit-terraform#85.

OK cool. Next week I'll have a look.

I think we should also remove the `autoscaling_enabled = true` option from workers and remove the autoscaling policies from the module. In a a separate PR. People can do this outside the module now and IRSA is one way to do it.

examples/irsa/main.tf
```diff
@@ -0,0 +1,94 @@
...
+    "kubernetes.io/role/internal-elb"             = "1"
+  }
+}
+
+module "eks" {
+  source       = "../.."
+  cluster_name = local.cluster_name
+  subnets      = module.vpc.private_subnets
+  vpc_id       = module.vpc.vpc_id
+  enable_irsa  = true
```
TBeijen: If I'm not mistaken, using IRSA on cluster-autoscaler, it would be good to add these flags to the cluster variables:

```
# module no longer needs to manage autoscaling policy since this is moved to the service-account IAM role
manage_worker_autoscaling_policy = false
attach_worker_autoscaling_policy = false
```

max-rocket-internet: You are not mistake but I think we just remove the autoscaling policy stuff completely and let users manage that themselves.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
 - repo: git://github.com/antonbabenko/pre-commit-terraform
-  rev: v1.22.0
+  rev: v1.24.0
```
barryib: @max-rocket-internet please change [args](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/710/commits/47d4489889e3dc192277221d982c305aca26be45#diff-82c3b36709926de3a1b27bea0d8258ccL7)  into `args: ["--args=--with-aggregate-type-defaults", "--args=--no-escape"]`

examples/irsa/main.tf
TBeijen: May I suggest: Leave the policy stuff in but change default enabled to `false` in next release (and ofc. mention in release notes). Then remove in subsequent release.

This would allow users to somewhat seamlessly transition to IRSA, without needing to temporary bolt on additional policies.

max-rocket-internet: @TBeijen actually we already have `var.attach_worker_autoscaling_policy`/`var.manage_worker_autoscaling_policy`. I'll just make a PR to set these to default of `false`.

barryib:
Can you please update your branch and upgrade to terraform-docs 0.8.1 ?

examples/irsa/main.tf
```diff
@@ -0,0 +1,91 @@
...
+  token                  = data.aws_eks_cluster_auth.cluster.token
+  load_config_file       = false
+  version                = "~> 1.10"
+}
+
+data "aws_availability_zones" "available" {}
+
+data "aws_caller_identity" "current" {}
+
+resource "random_string" "suffix" {
```
dpiddockcmp: Is this resource used anywhere?

max-rocket-internet: Thanks. I've removed it.

max-rocket-internet:
@barryib done rebase. Thanks for fixing the pre-commit stuff

ankushagarwal:
@max-rocket-internet : Curious as to what is the reason to use `worker_groups` instead of `node_groups` in module "eks". I am not sure what the difference between them is. The `managed_node_groups/main.tf` example uses `node_groups` in module "eks"

morganchristiansson:
worker_groups is the classic unmanaged way of doing things and allows you to use spot instances and provide kubelet args.

I believe both will work with irsa.

PR #715, opened 2020-01-28
Don't create ingress rule if worker security group exists

andjelx:
Don't create ingress rule if worker security group exists

# PR o'clock

## Description

Disable creation of ingress rule on worker security group if exists

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ## [[v8.?.?](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v8.1.0...HEAD)] - YYYY-MM-DD]
 
 - Include ability to configure custom os-specific command for waiting until kube cluster is healthy (@sanjeevgiri)
+- Disable creation of ingress rules if worker nodes security groups are exists (@andjelx)
 
 # History
 
--- a/cluster.tf
+++ b/cluster.tf
@@ -74,7 +74,7 @@ resource "aws_security_group_rule" "cluster_egress_internet" {
 }
 
 resource "aws_security_group_rule" "cluster_https_worker_ingress" {
-  count                    = var.create_eks ? 1 : 0
+  count                    = var.worker_security_group_id == "" && var.create_eks ? 1 : 0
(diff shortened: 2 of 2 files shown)


andjelx:
@max-rocket-internet thanks for quick review.
Can u please share the estimated date when next release of the module is planned ?

max-rocket-internet:
@andjelx very soon: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/720

ryanooi:
Hi. Why want to disable it when worker security group exists? the worker security group will need to another step to whitelist cluster security group manually after EKS created.

andjelx:
@ryanooi The idea is: if you have worker security groups already, you should add proper rules as well before provisioning EKS. If you don't have - all will be managed by terraform-aws-eks module.

ryanooi:
> @ryanooi The idea is: if you have worker security groups already, you should add proper rules as well before provisioning EKS. If you don't have - all will be managed by terraform-aws-eks module.

@andjelx I got your idea. My question is how can you get the cluster's security group (to whitelist on worker node security group) before provision EKS? 

Your steps will be:
1) Create worker node security group
2) Provision cluster and cluster security group
3) Whitelist cluster's security group on worker node security group

Previous steps will be:
1) Create worker node security group
2) Provision cluster and the security group will get whitelisted after cluster created.

that's a security group rule resource which is quite standalone. The previous change was aim to reduce 1 step and it's cleaner(Will get clean up when you destroy the cluster). Unless you're saying this EKS module shouldn't touch security group created outside of the module as of Terraform best practice.

andjelx:
@ryanooi 

Actually yes we can have cluster security group and pass it as **cluster_security_group_id**

> Unless you're saying this EKS module shouldn't touch security group created outside of the module as of Terraform best practice.

That was exactly my point.

ryanooi:
> @ryanooi
> 
> Actually yes we can have cluster security group and pass it as **cluster_security_group_id**
> 
> > Unless you're saying this EKS module shouldn't touch security group created outside of the module as of Terraform best practice.
> 
> That was exactly my point.

That's was your assumption that someone/everyone will want the module to manage both security groups or let the module manage both. But OK. I'm fine with it.  Thanks

PR #716, opened 2020-01-28
Remove of autoscaling IAM policy related stuff

max-rocket-internet:
# PR o'clock

## Description

I think we want to remove the cluster-autoscaler policy from the module. One reason is to reduce complexity but also now that IRSA is GA for EKS, that makes a much more elegant solution.

### Checklist

- [x] CI tests are passing


max-rocket-internet:
@barryib or we just remove the whole lot now? Otherwise it's 2 breaking changes, one setting defaults to `false` and then another removing the vars and policy?

barryib:
Can you please update your branch and upgrade to terraform-docs 0.8.1 ?

barryib:
I think we can remote it and hold the merge for https://github.com/terraform-aws-modules/terraform-aws-eks/pull/710

max-rocket-internet:
OK I think we should just remove it completely in one PR to avoid 2 breaking changes. I've written notes on the release with instructions.

examples/irsa/main.tf
```diff
@@ -70,13 +70,7 @@ module "eks" {
...
       instance_type        = "t2.medium"
       asg_desired_capacity = 1
-      tags = [
-        {
-          "key"                 = "k8s.io/cluster-autoscaler/enabled"
-          "propagate_at_launch" = "false"
-          "value"               = "true"
-        }
-      ]
+      autoscaling_enabled  = true
```
barryib: Why can't we remove completely remove autoscaling management from this module ? We can add in the doc that user should add required tags like https://github.com/terraform-aws-modules/terraform-aws-eks/pull/716/files#diff-a2651bd231ac6e9334ebf2cf3bd8fa76L73-L79

max-rocket-internet: I was thinking just leave that option in as it's an easy way to add the required tags without having to look up what they are. But yeah I guess it's cleaner to just remove this as well.

max-rocket-internet: OK I've removed it also.

workers.tf
```diff
@@ -104,29 +104,6 @@ resource "aws_autoscaling_group" "workers" {
...
-      },
-      {
-        "key" = "k8s.io/cluster-autoscaler/node-template/resources/ephemeral-storage"
-        "value" = "${lookup(
-          var.worker_groups[count.index],
-          "root_volume_size",
-          local.workers_group_defaults["root_volume_size"],
-        )}Gi"
-        "propagate_at_launch" = false
-      },
```
barryib: We don't need to add these tags in the exemple https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/irsa/main.tf#L73 ?

max-rocket-internet: I was trying to keep it as simple as possible and that single tag is all that is needed for autodiscovery to work.....but the docs say `It is recommended to use a second tag` so I'll add the second one. I think we leave `ephemeral-storage` out for now, I've never found good documentation of exactly how this tag works.

sc250024:
@max-rocket-internet Not to bring up an already merged PR, but why was this done? From your description:

> One reason is to reduce complexity but also now that IRSA is GA for EKS, that makes a much more elegant solution.

I understand removing it for complexity, but why does IRSA justify removing the tags from the ASGs? AFAIK, they're still needed for Cluster Autoscaler, no matter what mechanism the pods are using to authenticate with AWS. Did something change in that regard?

max-rocket-internet:
Hey @sc250024 

> why does IRSA justify removing the tags from the ASGs

The cluster-autoscaler needs some IAM policy and some tags on the ASGs.

Because we stopped managing the IAM policy (because IRSA is better for this) then I think it makes sense to move the cluster-autoscaler tags to the normal method that we already have for tagging ASGs: using `worker_groups[]. tags` e.g. [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/irsa/main.tf#L73-L83)

Does that make sense?

max-rocket-internet:
> How would a person keep them if they upgrade to v9.0.0 ?

Just add this to your worker group: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/irsa/main.tf#L73-L83

sc250024:
@max-rocket-internet Got it. I was already using IRSA for Cluster Autoscaler before v9.0.0, so I was already creating the IAM policy myself in another way. However, I was using the ASG tags.

Anyway, what you mentioned makes sense now. Thank you!

PR #731, opened 2020-02-10
Patch kube-proxy metricsBindAddress in configmap

syst0m:
# PR o'clock

## Description

The default metrics bind address for kube-proxy is `127.0.0.1:10249`. 
As a result,`prometheus` cannot scrape the `kube-proxy` metrics, and the bundled `kubernetes-monitoring-mixin` alerts for will always fire for `kube-proxy` objects, even though the `kube-proxy` pods are up.
The solution is to use `0.0.0.0:10249` as the `metricsBindAddress`, which allows `prometheus` to access the `kube-proxy` `/metrics` endpoint.
See: [prometheus-operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator#kubeproxy)

There have been some initaitives in k8s upstream, to change this to `0.0.0.0`, by default.
See: [Expose kube-proxy metrics on 0.0.0.0 by default #74300](https://github.com/kubernetes/kubernetes/pull/74300)

Until this is changed, the only way to fix `kube-proxy` scraping, is to patch the `metricsBindAddress` in the `kube-proxy` configmap. Which is what this PR does.
Kudos to @jcpetruzza, for helping with this.

### Checklist

- [ ] CI tests are passing


max-rocket-internet:
Hey @syst0m 
This looks like an important issue but patching kube-proxy and anything related to monitoring is not something we want to address in the module. We have to draw the line somewhere otherwise complexity and issues get too much. I hope that makes sense!

Issue #726, opened 2020-02-03
EKS create timeouts inconsistent between module and resource (over two repositories)


verdverm:
The default value here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/variables.tf#L189

Is not consistent with the default value here: https://github.com/terraform-providers/terraform-provider-aws/issues/5165

It should be updated to "30m" to reflect the good default value selected else where.

confiq:
I totally agree! 
the defaults should come from upstream. Do you mind to open a PR? If not, I can do it

barryib:
@verdverm @confiq Thanks for opening this issue. Can you please open a PR for this ?

confiq:
I have a feeling that I've hijacked this issue :)
 I've create the PR for this

barryib:
Resolved by #736

PR #736, opened 2020-02-11
changing timeout

confiq:
# PR o'clock

## Description

In issue #726, we concluded that timeout should be the same as from upstream aws provider. 
Therefore, I've changed it to 30m. 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -189,7 +189,7 @@ variable "kubeconfig_name" {
 variable "cluster_create_timeout" {
   description = "Timeout value when creating the EKS cluster."
   type        = string
-  default     = "15m"
+  default     = "30m"
 }
 
 variable "cluster_delete_timeout" {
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -189,13 +189,13 @@ variable "kubeconfig_name" {
...
   type        = string
-  default     = "15m"
+  default     = "30m"
 }
 
 variable "cluster_delete_timeout" {
   description = "Timeout value when deleting the EKS cluster."
   type        = string
-  default     = "15m"
+  default     = "30m"
```
barryib: It sounds like the default for delete timeout is `15m` https://www.terraform.io/docs/providers/aws/r/eks_cluster.html#timeouts

confiq: no idea

I've changed and squashed

PR #751, opened 2020-02-26
Add ASG cooldown and health_check_grace_period options

ArieLevs:
# PR o'clock

## Description

Add [default_cooldown](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#default_cooldown)
and [health_check_grace_period](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#health_check_grace_period)
options

### Checklist

- [ ] CI tests are passing


ArieLevs:
@barryib @max-rocket-internet 
Hi how can i get this PR get reviewed?

version
```diff
@@ -1 +1 @@
-v8.2.0
+v8.2.1
```
max-rocket-internet: Please leave this file as it's only changed when a new release is created.

```suggestion
v8.2.0
```

workers_launch_template.tf
```diff
@@ -72,6 +72,16 @@ resource "aws_autoscaling_group" "workers_launch_template" {
     "termination_policies",
     local.workers_group_defaults["termination_policies"]
   )
+  default_cooldown = lookup(
```
max-rocket-internet: Can you please add these to `workers.tf` also?

max-rocket-internet:
Screen Shot 2020-02-27 at 11 21 30

👈😃👈

ArieLevs:
@max-rocket-internet 
Can you please look into this, 
I cannot find a way to resolve conflicts from my side without reverting the commit on changelog

ArieLevs:
@max-rocket-internet 
Hi,
Anything i can do to promote this?

local.tf
```diff
@@ -32,6 +32,8 @@ locals {
     asg_force_delete              = false                       # Enable forced deletion for the autoscaling group.
     asg_initial_lifecycle_hooks   = []                          # Initital lifecycle hook for the autoscaling group.
     asg_recreate_on_change        = false                       # Recreate the autoscaling group when the Launch Template or Launch Configuration change.
+    default_cooldown              = "300"                       # The amount of time, in seconds, after a scaling activity completes before another scaling activity can start.
```
barryib: Where did you take the `300`  as default ? Maybe, the default should be `null` ?

ArieLevs: As i checked the [health_check_grace_period](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#health_check_grace_period) probably should be set to `null`,

But the [default_cooldown](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#default_cooldown), is not being set by terraform (and should bee 300 as [stated here](https://docs.aws.amazon.com/autoscaling/ec2/userguide/Cooldown.html#cooldown-default)).
That means if set `default_cooldown` to lets say 90 seconds, and later remove this, the value will not get set back to `300` but stay on `90`.

i'm not sure who should enforce defaults (as for example [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.create_auto_scaling_group) has this default),

But let me know what should be the desired behaviour from this module? null?

max-rocket-internet: I would go with null for both as this won't change any behaviour for existing users of the module. I think that is safest option?

ArieLevs: updated

workers.tf
```diff
@@ -73,6 +73,16 @@ resource "aws_autoscaling_group" "workers" {
     "termination_policies",
     local.workers_group_defaults["termination_policies"]
   )
+  default_cooldown = lookup(
+    var.worker_groups_launch_template[count.index],
```
max-rocket-internet: You have copy pasted something wrong here:

```suggestion
    var. worker_groups[count.index],
```

workers.tf
```diff
@@ -73,6 +73,16 @@ resource "aws_autoscaling_group" "workers" {
...
+    "default_cooldown",
+    local.workers_group_defaults["default_cooldown"]
+  )
+  health_check_grace_period = lookup(
+    var.worker_groups_launch_template[count.index],
```
max-rocket-internet: You have copy pasted something wrong here:

```suggestion
    var.worker_groups[count.index],
```

barryib:
@ArieLevs can you please update your branch and resolve conflits.

ArieLevs:
this PR is a total mess containing other files, 
I'll just create a new one.

PR #755, opened 2020-02-28
cluster-autoscaler shouldn't be scheduled on spot

simowaer:
Update documentation to schedule cluster-autoscaler on normal instances.

# PR o'clock

## Description

The documentation states the cluster autoscaler should be scheduled on spot instances. This is incorrect. It's the exact opposite.


simowaer:
@barryib Thanks. updated.

Issue #754, opened 2020-02-28
ami_id never read from workers_group_defaults


ck3mp3r:
If I understand the referenced line of code in [workers_launch_template.tf:237](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/908c656cf5da50d576d77696efac1d76dfb86a07/workers_launch_template.tf#L237)
the `ami_id` provided by `workers_group_defaults` will actually never be applied. I have seen the effect of this in our code after updating to newer version of this module. Instead it immediately defaults to the platform specific ami derived from the ami data lookup. 
Current workaround is to specify the ami_id via `worker_groups_launch_template` for every element in the supplied list...

PR #756, opened 2020-02-28
defaults lookup before falling back on default ami based on platform

ck3mp3r:
## Description
This addresses the missing lookup to provided worker defaults before falling back on platform based ami default.
#754


workers.tf
```diff
@@ -143,7 +143,10 @@ resource "aws_launch_configuration" "workers" {
   image_id = lookup(
     var.worker_groups[count.index],
     "ami_id",
-    lookup(var.worker_groups[count.index], "platform", local.workers_group_defaults["platform"]) == "windows" ? local.default_ami_id_windows : local.default_ami_id_linux,
+    lookup(
+      local.workers_group_defaults, "ami_id",
+      lookup(var.worker_groups[count.index], "platform", local.workers_group_defaults["platform"]) == "windows" ? local.default_ami_id_windows : local.default_ami_id_linux
+    )
```
barryib: I would prefer something easier to read. Something like this [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/908c656cf5da50d576d77696efac1d76dfb86a07/local.tf#L22-L23) :

```hcl
default_ami_id_linux   = lookup(local.workers_group_defaults, "ami_id", data.aws_ami.eks_worker.id)
default_ami_id_windows = lookup(local.workers_group_defaults, "ami_id", data.aws_ami.eks_worker_windows.id)
```
And leave workers.tf and workers_launch_template.tf unchanged. What do you think ?

ck3mp3r: Yeah, I see where you coming from but then the same holds true for worker_launch_template specific overrides... I'd rather have it where it is being used so that when people look at the code and follow the logic they'll see the order of precendence... make sense?

ck3mp3r:
> Thanks @ck3mp3r for your PR.
> 
> Please update the changelog ?

Just added some info on the changelog.

workers.tf
ck3mp3r: Okay, I'll implement it this way, check back in a sec.

ck3mp3r: @barryib as you can see I have implemented your suggestion. One thing that worries me now is that in the plan it doesn't show the default AMI IDs but rather `null` so I'll spend some more time on this.

ck3mp3r: Okay, implemented as you suggested. Instead of `local.workers_group_defaults` I used `var.workers_group_defaults` however.

local.tf
```diff
@@ -19,8 +19,8 @@ locals {
   worker_group_count                 = length(var.worker_groups)
   worker_group_launch_template_count = length(var.worker_groups_launch_template)
 
-  default_ami_id_linux   = data.aws_ami.eks_worker.id
-  default_ami_id_windows = data.aws_ami.eks_worker_windows.id
+  default_ami_id_linux   = lookup(var.workers_group_defaults, "ami_id", data.aws_ami.eks_worker.id)
+  default_ami_id_windows = lookup(var.workers_group_defaults, "ami_id", data.aws_ami.eks_worker_windows.id)
```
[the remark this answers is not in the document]
ck3mp3r: @barryib not sure I follow with using coalesce instead of lookup. With lookup you specify a key to get a value or default to something else, coalece picks first non null element... not sure how this is going to solve the problem of an empty query result from the ami data source... it might be a downstream problem pertaining to our account...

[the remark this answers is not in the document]
ck3mp3r: @barryib I've just implemented your improvement request. Also found out why it tried to use null instead of the default from the data, wasn't checking for empty provided value.
Works as advertised and tested a few permutations. Let me know if you need more changes. Thanks for your input!

local.tf
```diff
@@ -19,8 +19,8 @@ locals {
   worker_group_count                 = length(var.worker_groups)
   worker_group_launch_template_count = length(var.worker_groups_launch_template)
 
-  default_ami_id_linux   = data.aws_ami.eks_worker.id
-  default_ami_id_windows = data.aws_ami.eks_worker_windows.id
+  default_ami_id_linux   = coalesce(local.workers_group_defaults.ami_id, data.aws_ami.eks_worker.id)
+  default_ami_id_windows = coalesce(local.workers_group_defaults.ami_id, data.aws_ami.eks_worker_windows.id)
```
barryib: @ck3mp3r @max-rocket-internet I was just wondering how do we should handle hybrid kubernetes cluster (linux and windows). Do you think it worth to have `local.workers_group_defaults.ami_id` (default for linux) and `local.workers_group_defaults.ami_id_windows` (default for windows) to handle default ami id differently for both platforms ? Or we should let those people specify the ami_id for every worker group like https://github.com/terraform-aws-modules/terraform-aws-eks/issues/754 ?

Any thought ?

ck3mp3r: @barryib good point, I think your suggestion works. If we specify our own defaults we might as well specify them per OS... isn't going to hurt and completes the pattern.

ck3mp3r: @barryib this seems more involved than I anticipated 😆
How about we get this branch done first and then thing about how we could enhance in a next step?

barryib: @max-rocket-internet any thought ?

max-rocket-internet: Sounds good to me.

max-rocket-internet: > handle default ami id differently for both platforms ?
Yep

barryib:
Great.  @ck3mp3r  can you update your branch please so I can merge your PR ?

ck3mp3r:
> Great. @ck3mp3r can you update your branch please so I can merge your PR ?

Done :)

Issue #753, opened 2020-02-27
Tagging spot instances with launch templates


Art3mK:
## I'm submitting a...

* [x] feature request
* [x] kudos, thank you, warm fuzzy

## What is the current behavior?

[Currently](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers_launch_template.tf#L339-L352) there is a way to pass tags to launch templates, but it defines those only for volumes.
 
My use case is deny IAM policy, which doesn't allow to call `ec2:RunInstances` without specifying certain tags for instance. I know there is way for ASG to pass tags to instance, but [AWS Documentation](https://docs.aws.amazon.com/autoscaling/ec2/userguide/autoscaling-tagging.html#tag-lifecycle) states that:
> In most cases, when an Auto Scaling group launches instances, it adds tags to the instances during resource creation rather than after the resource is created.

>  The exception is when you use a launch configuration to launch Spot Instances. For this scenario, your Auto Scaling group adds tags while the instances are in the Pending lifecycle state

> if you need the Auto Scaling group to add tags to instances as part of the same API call that launches the Spot Instances, consider migrating to launch templates. 

But currently there is no way to pass tags for instances down to launch templates.

In my case debugging of that took a lot of time because of terraform showing this error:

```
Error: Error creating AutoScaling Group: AccessDenied: You are not authorized to use launch template: lt-009a3ea50f5e2c231
    status code: 403, request id: 2f09ef93-17a6-4b41-950d-03b3f194411d
      on .terraform/modules/eks/terraform-aws-modules-terraform-aws-eks-dd06c59/workers_launch_template.tf line 3, in resource "aws_autoscaling_group" "workers_launch_template":
       3: resource "aws_autoscaling_group" "workers_launch_template" {
```

So error was unclear, but cause of error was that missing instance tag in launch template.

## What's the expected behavior?

One could specify instance tags via launch template

-----
Thanks!

barryib:
This is now implemented in https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v11.0.0.

See #762 #822 for more info.

PR #762, opened 2020-03-05
fix: Add instance tag specification to launch template

js-timbirkett:
# PR o'clock

## Description

Tags are being applied after instances are launched whils in pending state.

Adding a tag_specification for instance resources allows ensures that
instances are tagged during launch.

Fixes: #753

### Checklist

- [x] CI tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [[v9.?.?](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v9.0.0...HEAD)] - 2020-xx-xx]
 
+- Fix #753 by adding and instance tag spec to launch templates (by @js-timbirkett)
 - Fix doc about spot instances, cluster-autoscaler should be scheduled on normal instances instead of spot (by @simowaer)
 
 # History
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -337,6 +337,21 @@ resource "aws_launch_template" "workers_launch_template" {
   }
 
   tag_specifications {
+    resource_type = "instance"
+
+    tags = merge(
+      {
+        "Name" = "${aws_eks_cluster.this[0].name}-${lookup(
+          var.worker_groups_launch_template[count.index],
+          "name",
+          count.index,
+        )}-eks_asg"
+      },
+      var.tags,
+    )
+  }
+
+  tag_specifications {
     resource_type = "volume"
 
     tags = merge(


max-rocket-internet:
TBH I don't really understand the changes in this PR as I don't have experience here.

@barryib do you know how `tag_specifications` works and what "Tags are being applied after instances are launched" does?

barryib:
@max-rocket-internet I don't use it either, but the use case sounds legit to me. From what I understood in https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html, it's add tags to EC2 instances during their creation.

@Art3mK  does this resolve your issue https://github.com/terraform-aws-modules/terraform-aws-eks/issues/753 ?

Art3mK:
yeah, I think so! Thanks for checking this! 

Use case is that with spot instances ASG will add tags to instances after those are created
> The exception is when you use a launch configuration to launch Spot Instances. For this scenario, your Auto Scaling group adds tags while the instances are in the Pending lifecycle state

Problem with this is that if you have policy, which denies you to launch instance without proper tagging - there is no way currently with this module to do that. And error message from API is very cryptic. See issue. That took me some time to solve that :)

This could be solved with adding that `tag_specification` for instances also in launch template (there is already one for volumes).

I'm not sure though how these tags for instances defined in launch template will work with tags defined on ASG itself. Would be there any conflicts/overwrites if both defines the same tag?

max-rocket-internet:
> I'm not sure though how these tags for instances defined in launch template will work with tags defined on ASG itself. Would be there any conflicts/overwrites if both defines the same tag?

True. I'm not sure. As there would be a lot of overlap. @js-timbirkett do you know? Are they just merged?

barryib:
> > I'm not sure though how these tags for instances defined in launch template will work with tags defined on ASG itself. Would be there any conflicts/overwrites if both defines the same tag?
> 
> True. I'm not sure. As there would be a lot of overlap. @js-timbirkett do you know? Are they just merged?

@js-timbirkett can you confirm please and remove your changes from changelog. It's now auto generated during the release process.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [[v9.?.?](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v9.0.0...HEAD)] - 2020-xx-xx]
 
+- Fix #753 by adding and instance tag spec to launch templates (by @js-timbirkett)
```
barryib: Can you please your changes in changelog ? It's now generated from git commits during release.

Reuuke:
> > I'm not sure though how these tags for instances defined in launch template will work with tags defined on ASG itself. Would be there any conflicts/overwrites if both defines the same tag?
> 
> True. I'm not sure. As there would be a lot of overlap. @js-timbirkett do you know? Are they just merged?
Yep. They just merged. And tags from ASG have a higher priority then tags from LT if they match on tag key (I tested today)

barryib:
Great closing this in favor of #822

PR #784, opened 2020-03-12
improvement: Require k8s provider >=1.11.1

### LINKED ISSUE #713: kubernetes provider does not respect local config, can operate on other clusters
# I have issues
The k8s provider does not seem to reliably work when `load_config_file = false`, as it would be when using this module to create a new cluster. I frequently see Unauthorized and/or attempts to call the endpoint on localhost. In one execution, I actually found that this module deleted the aws-auth config map from a cluster that was defined as the default context in my kubeconfig but was not in any way related to my terraform run.

## I'm submitting a...

* [x] bug report

Without defining your cluster in .kubeconfig, the provider cannot reliably be configured. This seems to make sense based on a bunch of bugs in the terraform-provider-kubernetes project, most specifically:

https://github.com/terraform-providers/terraform-provider-kubernetes/issues/521

Others seem to confirm tons of bugs when setting `load_config_file = false`. This one seemed most relevant since it also pointed toward long standing issues with terraform itself using interpolated values to configure a provider:

https://github.com/hashicorp/terraform/issues/4149

In my execution where it operated on a different cluster, the deletion of the config map was successful, but it attempted to apply the config map back to an endpoint listening on localhost. In other executions, the deletion attempt ran against localhost. This suggests that there is a timing issue where terraform itself has deferred configuring the provider concurrently with the attempt to manage the config map in this module.

I failed to create a cluster using the provided example, unless after cluster creation, I then added the cluster config to my local kubeconfig.

Regardless of any upstream issues with terraform and the k8s provider, this module should _never_ operate on a cluster it didn't define. Defining a local kubeconfig file and pointing the provider to that may be the best option.

I'm not sure what the best fix is here. Given the various bugs in GitHub for this, I personally feel like the only workaround here is to set `manage_aws_auth = false` always until the upstream provider addresses these issues.

## Environment details

* Affected module version: 8.0.0
* OS: macOS
* Terraform version: 0.12.20

### LINKED ISSUE #733: Kubernetes provider setup broken for version 1.11
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [x] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!

Kubernetes provider setup breaks when the provider version is upgraded to v1.11

Provider configuration
```
provider "kubernetes" {
  host                   = data.aws_eks_cluster.cluster.endpoint
  cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
  token                  = data.aws_eks_cluster_auth.cluster.token
  version                = "~>1.9"
}
```

`terraform init`
image

`terraform plan`
image

Plan should work

## Environment details

Terraform v0.12.19
+ provider.aws v2.48.0
+ provider.kubernetes v1.11.0
+ provider.local v1.4.0
+ provider.null v2.1.2
+ provider.random v2.2.1
+ provider.template v2.1.2

* OS: Darwin

dpiddockcmp:
# PR o'clock

## Description

Set the required kubernetes provider version to 1.11.1.

Looks like the original breaking changes in 1.11.0 have been fixed. And hopefully this also resolves the problems people were seeing with the provider randomly trying to modify their default or localhost clusters. Or at least we'll encourage further testing of the provider 😁 

Not sure how you want to classify this. It's not really a breaking change but it may cause terraform to fail to init if users have set `version = "1.10"` like we had in the examples.

Fixes #733
Fixes #713 [probably]

### Checklist

- [x] Add [sementics prefix](#semantic-pull-requests) to your PR or Commits (at leats one of your commit groups)
- [x] CI tests are passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | Name | Version |
 |------|---------|
 | aws | >= 2.52.0 |
-| kubernetes | >= 1.6.2 |
+| kubernetes | >= 1.11.1 |
 | local | >= 1.2 |
 | null | >= 2.1 |
 | random | >= 2.1 |
--- a/examples/basic/main.tf
+++ b/examples/basic/main.tf
@@ -36,7 +36,7 @@ provider "kubernetes" {
   cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
   token                  = data.aws_eks_cluster_auth.cluster.token
   load_config_file       = false
-  version                = "1.10"
+  version                = "~> 1.11"
 }
 
 data "aws_availability_zones" "available" {
--- a/examples/irsa/main.tf
+++ b/examples/irsa/main.tf
@@ -32,7 +32,7 @@ provider "kubernetes" {
   cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
   token                  = data.aws_eks_cluster_auth.cluster.token
   load_config_file       = false
-  version                = "1.10"
+  version                = "~> 1.11"
 }
 
 data "aws_availability_zones" "available" {}
--- a/examples/launch_templates/main.tf
+++ b/examples/launch_templates/main.tf
@@ -36,7 +36,7 @@ provider "kubernetes" {
   cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
   token                  = data.aws_eks_cluster_auth.cluster.token
   load_config_file       = false
-  version                = "1.10"
+  version                = "~> 1.11"
 }
 
 data "aws_availability_zones" "available" {
--- a/examples/managed_node_groups/main.tf
+++ b/examples/managed_node_groups/main.tf
@@ -36,7 +36,7 @@ provider "kubernetes" {
   cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
   token                  = data.aws_eks_cluster_auth.cluster.token
   load_config_file       = false
-  version                = "1.10"
+  version                = "~> 1.11"
 }
 
 data "aws_availability_zones" "available" {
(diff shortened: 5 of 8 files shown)


barryib:
@dpiddockcmp thanks for your PR. Can you please update your branch according to the [new contribution guide](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/CONTRIBUTING.md) and resolve conflicts.

You don't need to modify changelog anymore. It's autogenerated with git-chglog.

dpiddockcmp:
Changes done. 

Is this a potentially breaking change? I guess if users set their kubernetes provider version to 1.10.0 to avoid the broken 1.11.0.

barryib:
> Changes done.
> 
> Is this a potentially breaking change? I guess if users set their kubernetes provider version to 1.10.0 to avoid the broken 1.11.0.

Humm does the 1.11.1 breaks anything from 1.10.x ?

dpiddockcmp:
No. There were no breaking changes in the provider. However people who followed the examples and used `provider "kubernetes" { version = "1.10" }` will get an error on terraform init. Simple enough to fix but a little rude 😅

barryib:
Understood. Looking to way to add breaking change note in commits.

barryib:
I merged this with a typo in the commit. I commited `Improvement: Require kubernetes provider >=1.11.1 (#784)` instead of `improvement: Require kubernetes provider >=1.11.1 (#784)` (Improvement != improvement). So this commit is not correctly added in the changelog.

I opened this https://github.com/git-chglog/git-chglog/issues/63 but I don't think it'll go somewhere.

I was looking for a way to fix this.

- Should I amend on master ? (I don't like it)
- Should I revert this change and Re-open another PR to add this improvement back ?
- Another solution ?

@dpiddockcmp @max-rocket-internet @antonbabenko please advice.

FWIW, since this, I added a more restrictive configuration for sementic pr https://github.com/terraform-aws-modules/terraform-aws-eks/pull/804 and I'll probably add something like https://github.com/tibdex/autosquash to avoid this kind of error happening again.

antonbabenko:
I've made the same mistake a couple of days ago and then realized that "improvement" is not even in spec anymore. See here all allowed/supported types - https://github.com/commitizen/conventional-commit-types/blob/master/index.json

Probably "fix" is more universal for something that existed before while "feat" is for new features that didn't exist before.

I think we can safely roll forward and just learn from this. #804 you've made was important and I think it will be enough.

PR #788, opened 2020-03-13
Fix kitchen test for terraform 0.12.x

bigwheel:
# PR o'clock

## Description

Fix kitchen test for terraform 0.12.x.

Test codes are written 6 months ago and it was broken because it supported only terraform 0.11.x.
In 6 months, eks module's least support version bumped to 0.12.x therefore test codes didn't work completely.

This PR ups test library versions and fix broken point.

### Checklist

- [ ] ~Change added to CHANGELOG.md. All changes must be added and breaking changes and highlighted~
    - I think no need to do it
- [ ] ~CI tests are passing~
    - Unfortunately, in this module CI test is not running. The test with real AWS resources need real money.
- [ ] ~README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation~
    - no need

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -1,6 +1,7 @@
 ---
 driver:
   name: "terraform"
+  command_timeout: 3000
   root_module_directory: "examples/basic"
 
 provisioner:
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 ruby '2.4.6'
 
 source 'https://rubygems.org/' do
-  gem 'awspec', '~> 1.4.2'
-  gem 'kitchen-terraform', '~> 3.2'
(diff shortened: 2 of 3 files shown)


bigwheel:
P.S.

I think that [gruntwork\-io/terratest](https://github.com/gruntwork-io/terratest) is best for terraform module test.
But,  firstly fix existing test. it will become trust.

max-rocket-internet:
I opened a PR to just remove the kitchen stuff https://github.com/terraform-aws-modules/terraform-aws-eks/pull/787

I don't think it's valuable. Would you agree?

bigwheel:
@max-rocket-internet 
Basically agree.
I believe terraform test is valuable but terraform-aws-eks couldn't use it well.
Let's remove.

Issue #777, opened 2020-03-11
Deployment never completes but cluster is active


Tomthi:
# I have issues
The same code was use to build two other clusters and they are working. There was some lag from when the initial clusters were built. 

## I'm submitting a...

* [X] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

The cluster deploy never completes with worker nodes and ASG's.  The Cluster state does show active. Seems like the ASG' and worker nodes are never deployed. 

## If this is a bug, how to reproduce? Please include a code sample if relevant.

When running apply, the cluster is built and in an active state but the node groups are never created, there are no ASG's or launch templates. 

It's stuck in a loop
module.eks-cluster.null_resource.wait_for_cluster[0]: Still creating...
module.eks-cluster.null_resource.wait_for_cluster[0]: Still creating...
module.eks-cluster.null_resource.wait_for_cluster[0]: Still creating...
module.eks-cluster.null_resource.wait_for_cluster[0]: Still creating...

Main.tf
module "eks-cluster" {
  source                          = "terraform-aws-modules/eks/aws"
  version                         = "9.0.0"
  cluster_name                   = "qa-eps-eks"
 subnets                             =  "${data.aws_subnet_ids.default.ids}"
  vpc_id                              =  "${var.vpc_id}"
  cluster_endpoint_private_access = true
  cluster_endpoint_public_access  = false
  cluster_security_group_id =  "${data.aws_security_group.default.id}"

  worker_groups = [
    {
      name = "qa-eks-workers"
      instance_type = "m4.large"
      key_name = "qa"
      asg_min_size      = 2
      asg_desired_size  = 2
      asg_max_size      = 5
      autoscaling_enabled = true
      tags = [{
        propagate_at_launch = true
        key         = "terraform"
        value       = "true"
      }]
    }
  ]
  tags = {
    environment = "qa"
    terraform   = "true"
  }
}

## What's the expected behavior?

The expected behavior is the cluster is built along with ASG and worker nodes.

## Environment details

* Affected module version:
* OS:
* Terraform version: 0.12.23

## Any other relevant info 

Module v9.0.0

js-timbirkett:
[...]

PR #791, opened 2020-03-13
improvement: Add timeout to default wait_for_cluster_cmd

dpiddockcmp:
# PR o'clock

## Description

The current default command for `wait_for_cluster` will happily sit forever in a failing state. Module previously had a timeout before the switch to kubernetes provider in 8.0.0.

Defaults to 5 minutes of sleeps. May be a lot longer depending on how wget hangs. 

Have tested logic on bash 3 (for those older Mac users), zsh, bash 5, and busybox (default in alpine)
```
$ docker run --rm -it alpine /bin/sh -c 'for i in `seq 1 60`; do false && exit 0 || true; sleep 0; done; echo TIMEOUT && exit 1' ; echo $?
TIMEOUT
1
$ docker run --rm -it alpine /bin/sh -c 'for i in `seq 1 60`; do true && exit 0 || true; sleep 0; done; echo TIMEOUT && exit 1' ; echo $?
0
```

Sample failure output (I lowered the seq count because ain't got all day):
```
module.eks.null_resource.wait_for_cluster[0]: Creating...
module.eks.null_resource.wait_for_cluster[0]: Provisioning with 'local-exec'...
module.eks.null_resource.wait_for_cluster[0] (local-exec): Executing: ["/bin/sh" "-c" "for i in `seq 1 2`; do wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null && exit 0 || true; sleep 5; done; echo TIMEOUT && exit 1"]
module.eks.null_resource.wait_for_cluster[0]: Still creating... [10s elapsed]
module.eks.null_resource.wait_for_cluster[0] (local-exec): TIMEOUT


Error: Error running command 'for i in `seq 1 2`; do wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null && exit 0 || true; sleep 5; done; echo TIMEOUT && exit 1': exit status 1. Output: TIMEOUT
```

I don't think there are any issues directly about this but it would have helped the user in #777 for example. Or at least wasted slightly less time watching paint dry 🙂 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -201,7 +201,7 @@ variable "cluster_delete_timeout" {
 variable "wait_for_cluster_cmd" {
   description = "Custom local-exec command to execute for determining if the eks cluster is healthy. Cluster endpoint will be available as an environment variable called ENDPOINT"
   type        = string
-  default     = "until wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null; do sleep 4; done"
+  default     = "for i in `seq 1 60`; do wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null && exit 0 || true; sleep 5; done; echo TIMEOUT && exit 1"
 }
 
 variable "cluster_create_security_group" {
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | subnets | A list of subnets to place the EKS cluster and workers within. | `list(string)` | n/a | yes |
 | tags | A map of tags to add to all resources. | `map(string)` | `{}` | no |
 | vpc\_id | VPC where the cluster and workers will be deployed. | `string` | n/a | yes |
-| wait\_for\_cluster\_cmd | Custom local-exec command to execute for determining if the eks cluster is healthy. Cluster endpoint will be available as an environment variable called ENDPOINT | `string` | `"until wget --no-check-certificate -O - -q $ENDPOINT/healthz \u003e/dev/null; do sleep 4; done"` | no |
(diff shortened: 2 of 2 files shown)


variables.tf
```diff
@@ -201,7 +201,7 @@ variable "cluster_delete_timeout" {
 variable "wait_for_cluster_cmd" {
   description = "Custom local-exec command to execute for determining if the eks cluster is healthy. Cluster endpoint will be available as an environment variable called ENDPOINT"
   type        = string
-  default     = "until wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null; do sleep 4; done"
+  default     = "for i in `seq 1 2`; do wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null && exit 0 || true; sleep 5; done; echo TIMEOUT && exit 1"
```
barryib: The sequence is not too short ?

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [[v10.?.?](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v10.0.0...HEAD)] - 2020-xx-xx
 
-- Write your awesome change here (GH-xxxx)
+- Add timeout to default `wait_for_cluster_cmd` (by @dpiddockcmp)
```
barryib: Can you please remove your modification the changelog.

variables.tf
dpiddockcmp: Ahaha I'm an idiot. I set that for testing 😁

dpiddockcmp: Fixed 🙂

CHANGELOG.md
dpiddockcmp: Done

PR #792, opened 2020-03-13
fix: Fix cluster security group ids in outputs

slimm609:
# PR o'clock

## Description
when an eks cluster is created, the cluster security group is automatically created by eks, you can only add an additional security group, the outputs were returning the incorrect security group id because it was only returning the "additional" security group and not the cluster security group. 

### Checklist

- [ ] CI tests are passing


outputs.tf
```diff
@@ -24,8 +29,8 @@ output "cluster_version" {
 }
 
 output "cluster_security_group_id" {
-  description = "Security group ID attached to the EKS cluster."
-  value       = local.cluster_security_group_id
+  description = "Cluster security group ID attached to the EKS cluster."
+  value       = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
```
dpiddockcmp: Can you test this output with a 1.13 cluster? The EKS controlled cluster security group is only added for kubenetes 1.14 and eks.3 platform. I'm wondering what Terraform sets the value to or whether it's an error.

Also, this pattern won't work with `create_eks = false` or possibly with a half destroyed state. See the `concat(aws[*].output, [""])[0]` patterned used in other outputs.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [[v10.?.?](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v10.0.0...HEAD)] - 2020-xx-xx
 
-- Write your awesome change here (GH-xxxx)
+- Fixed eks security group id outputs and additional security group id inputs (by @slimm609)
```
dpiddockcmp: This is a breaking change. You've changed the value of the `cluster_security_group_id` output.

outputs.tf
slimm609: The first issue exists today.    Renaming the variables here didn't change anything besides the output.   

I fixed the concat but haven't pushed it up yet because I am trying to work a solution for the 1.13 issue but now that I have tested,  the issue exists in the current release as well as this change.

slimm609: in a 1.13 cluster, the security_group_id is the primary field but still exists (either created or passed in), in a 1.14+ cluster, the security_group_id still either exists or is created but is just used as the additional field.  I added a check to only use attempt to create node groups with 1.14 or greater.    The eks.3 version doesn't matter because if you try to create a node group going forward, it will use the latest version (eks.8) and if you have an existing and try to modify it, it will force a recreate and pick up the latest.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [[v10.?.?](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v10.0.0...HEAD)] - 2020-xx-xx
 
-- Write your awesome change here (GH-xxxx)
+- **Breaking:** renamed eks security group id outputs and additional security group id inputs (by @slimm609)
```
barryib: Please remove your change from the changelog.

local.tf
```diff
@@ -8,10 +8,10 @@ locals {
     )
   ]
 
-  cluster_security_group_id = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
-  cluster_iam_role_name     = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
-  cluster_iam_role_arn      = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
-  worker_security_group_id  = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
+  additional_security_group_id = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.additional_security_group_id
```
barryib: Why rename this ?

variables.tf
```diff
@@ -19,8 +19,8 @@ variable "cluster_name" {
   type        = string
 }
 
-variable "cluster_security_group_id" {
-  description = "If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingress/egress to work with the workers"
+variable "additional_security_group_id" {
```
barryib: Why rename this variable ? I don't get it.

slimm609: Because they are 2 different things after 1.14 and you can not set the primary security group, only the additional security group id. The primary is created via eks automatically

slimm609: Because they are 2 different things after 1.14 and you can not set the primary security group, only the additional security group id. The primary is created via eks automatically

outputs.tf
```diff
@@ -24,8 +29,8 @@ output "cluster_version" {
 }
 
 output "cluster_security_group_id" {
-  description = "Security group ID attached to the EKS cluster."
-  value       = local.cluster_security_group_id
+  description = "Cluster security group ID attached to the EKS cluster."
+  value       = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : local.additional_security_group_id
```
dpiddockcmp: Please don't do this. It will make upgrade paths very scary.

slimm609: this will have no effect on future upgrades.  This problem only exists because node_groups are not backwards compatible.   As soon at 1.16 is released, the possibility to hit this edge case no longer exists and this can be removed.  This only exist because of the issue, which is also broken in the current release.

variables.tf
slimm609: https://www.terraform.io/docs/providers/aws/r/eks_cluster.html

vpc_config - Additional nested attributes:
cluster_security_group_id - The cluster security group that was created by Amazon EKS for the cluster.

You can not create this or even give an security group id for it, you can only get it from the output, so the additional_security_group_id is `security_group_ids` and what shows in the AWS console as additional security groups

CHANGELOG.md
slimm609: Why?   The module output is incorrect from what the terraform aws_eks_cluster is referencing and currently outputs.   Just 2 days ago, it was commented that it was a breaking change and to update it as so.

local.tf
slimm609: same as below.   The module is currently incorrect with the aws eks provider

outputs.tf
dpiddockcmp: This variable has nothing to do with node groups. You are changing the value of an output based on user supplied input. It is a dangerous pattern.

Nothing is broken in the current release. It just lacks output of the new `vpc_config.cluster_security_group_id` value.

Issue #687, opened 2020-01-15
Unable to define additonal EBS volumes for the launch templates


josh-stableprice:
# I have issues attaching an additional EBS volume to my launch templates for use in cluster wide storage.

I want to have a minimally sized root storage volume with a larger attached EBS volume, which is used via Rook/Ceph 

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
I attempt to specify additional volumes under worker_groups_launch_template as per the documentation for aws_launch_template resource, however I get told that this isn't possible

## If this is a bug, how to reproduce? Please include a code sample if relevant.
```
module "eks" {
  source = "terraform-aws-modules/eks/aws"
  version = "7.0.1"

  cluster_name = var.cluster_name
  subnets = var.private_subnet_ids

  tags = {
    Environment = var.environment
    GithubRepo = "terraform-aws-eks"
    GithubOrg = "terraform-aws-modules"
  }

  vpc_id = var.vpc_id

  worker_groups_launch_template = [
    {
      name = "spot-1"
      override_instance_types = var.instance_types
      spot_instance_pools = length(var.instance_types)
      asg_max_size = 10
      asg_desired_capacity = 5
      autoscaling_enabled = true
      kubelet_extra_args = "--node-labels=kubernetes.io/lifecycle=spot"
      public_ip = true
      eni_delete = true
      volumes = [{
        device_name = "/dev/sdb"
        volume_size = 2000
        encrypted = true
      }]
    },

    {
      name = "on-demand-1"
      instance_type = "t3.medium"
      asg_max_size = 10
      asg_min_size = 0
      asg_desired_capacity = 0
      autoscaling_enabled = true
      kubelet_extra_args = "--node-labels=spot=false"
      public_ip = true
      eni_delete = true
      volumes = [{
        device_name = "/dev/sdb"
        volume_size = 2000
        encrypted = true
      }]
    }
  ]
}
```

## What's the expected behavior?
I'm able to pass in a volume definition of the same or similar form as the definition on aws_launch_template resource and I end up with an ASG with EBS volumes created and attached.

## Environment details

* Affected module version: 7.0.1
* OS: Linux
* Terraform version: 0.12.19

josh-stableprice:
[...]

PR #800, opened 2020-03-16
feat: Add support for additional volumes in launch templates and launch configurations

jaimehrubiks:
# PR o'clock

## Description

Now the user can specify additional volumes for its instances. This is
specially useful when the cluster is going to be used for specific
applications such as storage.

Example 1 (Just required fields)

    additional_volumes = [{
        block_device_name = "/dev/sdb",
        volume_size = 200,
        delete_on_termination = false
    }]

Example 2 (Multiple additional volumes & All features)

    additional_volumes = [{
        block_device_name = "/dev/sdb",
        volume_size = 200,
        volume_type = "io1"
        iops = 100
        encrypted = true
        kms_key_id = "arn:aws:kms:us-east-1:89349:key/a4-ds-34"
        delete_on_termination = false
    },{
        block_device_name = "/dev/sdc",
        volume_size = 500,
        delete_on_termination = false
    }]

Relevant Issues:

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/687
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/572

Note: I only implemented this feature on launch template workers

### Checklist

- [ ] CI tests are passing


jaimehrubiks:
Why is the code format failing on the local.tf?

js-timbirkett:
Is there a valid use case for extra volumes mounted to worker nodes? Can the same be done with managed nodes?

My thinking is that Kubernetes nodes are autoscaled and would be recycled / scaled up and down by the Cluster Autoscaler. EBS volumes are generally destroyed when an instance is terminated.

I'd think that stateful workloads in need of block storage such as Kafka, MySQL or Ceph would use PersistentVolume's and PersistentVolumeClaims along with an appropriate CSI driver to create, mount and handle lifecycle of volumes?

I might be wrong and may have missed some valuable context somewhere. I'm also not an expert in that field so always appreciate these learning opportunities 😄

jaimehrubiks:
First, I am also learning, not an expert at all. For most stateful applications, as you mention, the CSI way to claim storage is the way to request storage. However, for storage itself, say, Ceph, I found that it might be easier to manage and control to manually specify or add devices, and have a list of specific nodes and their devices. In my case, I was using rook.io, and to make it similar to on-prem, this was the only way. Of course, I would love to see other people's opinion on this matter.

snstanton:
With the new AWS bottlerocket os, the AMI for the worker nodes automatically mounts a secondary volume with a default size of 20GB to act as the containerd cache for the node.  In order to change the size of this volume, I need the ability to override this default size to something larger.   I literally just wrote something very similar to this patch, so it would be great to get the merged in.

snstanton:
It looks like the lint issues are a missing white space at `workers_launch_template.tf:355` and too much white space before the comment at `local.tf:77`.  If you run `terraform fmt` on the files, it will fix the lint issues for you.

jaimehrubiks:
I messed up with my git a little bit

snstanton:
Yes, in fact I branched and added this PR and it works perfectly.  Here's what I have in my config:
```
      additional_volumes     = [{ block_device_name = "/dev/xvdb", volume_size = 50 }]
```

snstanton:
You can fix the Semantic Pull Request failure by changing the PR title to something like:

`feat: add support for additional volumes in launch templates`

workers_launch_template.tf
```diff
@@ -351,6 +351,43 @@ resource "aws_launch_template" "workers_launch_template" {
     }
   }
 
+  dynamic "block_device_mappings" {
+    for_each = lookup(var.worker_groups_launch_template[count.index], "additional_volumes", lookup(local.workers_group_defaults, "additional_volumes", []))
```
barryib:
```suggestion
    for_each = lookup(var.worker_groups_launch_template[count.index], "additional_volumes", local.workers_group_defaults["additional_volumes"])
```

for consistency and `local.workers_group_defaults["additional_volumes"]` is already set by default to `[]`

workers_launch_template.tf
```diff
@@ -351,6 +351,43 @@ resource "aws_launch_template" "workers_launch_template" {
...
+    content {
+      device_name = block_device_mappings.value.block_device_name
+
+      ebs {
```
barryib: Does it make sense to rename `local.additional_volumes` into `local. additional_ebs_volumes` ? It sounds that we are only dealing with EBS and not ephemeral storage.

@jaimehrubiks @js-timbirkett @max-rocket-internet any thoughts ?

jaimehrubiks: I don't really have an opinion in this matter, both alternatives look good for me.

jaimehrubiks:
Will take a look at launch configuration tomorrow, to see if it is similar.

jaimehrubiks:
All changes applied. I did not test this feature on working groups (launch configuration)

workers_launch_template.tf
jaimehrubiks: accepted as makes sense

barryib:
> All changes applied. I did not test this feature on working groups (launch configuration)

Nice @jaimehrubiks. Thanks for your contribution.

Issue #799, opened 2020-03-14
cluster_create_security_group is mapped to Additional security groups in EKS


pabrahamusa:
While creating the cluster I set the Cluster Control Plane security group by  assigning cluster_create_security_group = "sg-xxxxxxx" . However once the cluster is created this will be associated as "Additional security groups" in the EKS not  as primary "Cluster security group".  (The wording is as per EKS aws console Networking block) . Also if I try to  check the output of "cluster_create_security_group" it is still "Additional security groups".  

I have some VPC endpoints and I have to associate this "Cluster Security group" &  "Additional security groups" to that. However I cannot see any output for "Cluster Security group".

So is there anyway I can get, set or modify the primary "Cluster security group" ?

slimm609:
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/792

I already added a PR to fix the outputs and map the additional security group,  The cluster primary security group is automatically created and you can not point it to any other security group,  The only thing you can do it get the ID and modify it outside the module, which is what this PR provides.

this is from the outputs in the terraform eks cluster section.
```
cluster_security_group_id - The cluster security group that was created by Amazon EKS for the cluster.
```

once this PR gets merged into the module, you can add to it by doing something like this outside the module. 
```
resource "aws_security_group_rule" "eks_cluster_add_access" {
  type        = "ingress"
  from_port   = 111
  to_port     = 111
  protocol    = "tcp"
  cidr_blocks = ["1.2.3.4/32"]

  security_group_id = module.eks.cluster_security_group_id
}
```

dpiddockcmp:
The variable naming in the module around security groups is historical.

The primary "Cluster security group" created by the EKS platform is a recent addition they added as part of supporting managed node groups. The module hasn't been updated yet to output the ID number of this group.

Previously you had to create a security group and associate it when creating the cluster or nothing would work. Now the console has renamed those groups to "Additional Security Groups".

You cannot set the primary security group. It's controlled by the EKS platform.

The cluster security groups are for allowing communication between the cluster control plane and the kubernetes worker nodes. I'm not sure what you're doing with VPC endpoints so that they need to communicate with the cluster control plane but can you not achieve it via the "additional" security groups?

You can see more about the cluster's security groups in the [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html)

pabrahamusa:
[...]

PR #828, opened 2020-04-11
fix: Add `vpc_config.cluster_security_group` output as primary cluster security group id

slimm609:
# PR o'clock
Add cluster security group output as "primary cluster security group id".  

This is a short term fix to provide the cluster security group id until 1.16 is released and node groups can be refactored to move cluster security group id to additional security group id.   With 1.16 release, 1.13 will no longer be maintained which will cause the issues with backwards compatibility to go away because the old node groups will no longer be supported.  

## Description

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/799
and 
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/792

### Checklist

- [ ] CI tests are passing


barryib:
Thanks @slimm609 for opening a new suggestion for this issue. Holding this PR a bit so we can move forward on https://github.com/terraform-aws-modules/terraform-aws-eks/issues/799 discussion

barryib:
Awsome. Thanks @slimm609 and @dpiddockcmp.

PR #846, opened 2020-04-23
fix: Wrap local.configmap_roles.groups with tolist() to avoid panic

keznikl:
## Description

Wrap local.configmap_roles.groups with tolist() in order to avoid golang panic.

Without this change we get the following on Terraform v0.12.24:

```
panic: inconsistent list element types (cty.Object(map[string]cty.Type{"groups":cty.DynamicPseudoType, "rolearn":cty.String, "username":cty.String}) then cty.Object(map[string]cty.Type{"groups":cty.Tuple([]cty.Type{cty.String, cty.String}), "rolearn":cty.String, "username":cty.String}))
```

It is caused by the ternary operator ( X ? Y : Z) in conjunction with distinct();
tolist(...) resolves the type and thus the issue .

### Checklist

- [x] CI tests are passing


aws_auth.tf
```diff
@@ -48,13 +48,13 @@ locals {
     {
       rolearn  = role["worker_role_arn"]
       username = "system:node:{{EC2PrivateDNSName}}"
-      groups = concat(
+      groups = tolist(concat(
```
barryib, line 51: Maybe we should also add [`compact`](https://www.terraform.io/docs/configuration/functions/compact.html) function to remove the empty list before concat

keznikl: Isn't `contat([...], [])` enough to get rid of it?

barryib: 🤔 humm.. from what I'm reading in the example https://www.terraform.io/docs/configuration/functions/concat.html it sounds that concat doesn't remove empty elements from list, but compact does.

Maybe doc is outdated ?

dpiddockcmp: It won't remove empty string items, which is what the docs are showing. But concating an empty list won't add anything.

```
$ terraform console
> concat(["a"], [])
[
  "a",
]
>  
```

barryib: Right

Issue #843, opened 2020-04-21
Changing tags causes node groups to be replaced.


blongman-snapdocs:
# I have issues

## I'm submitting a...

* [X] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
When I change the values of tags, a plan shows that all node groups need to be replaced:

```
 (feature/DEVOPS-586/recreate-east-cluster-but-better) $ terraform plan -no-color | grep "must be replaced"
  # module.eks.local_file.kubeconfig[0] must be replaced
  # module.eks.module.node_groups.aws_eks_node_group.workers["jobs"] must be replaced
  # module.eks.module.node_groups.aws_eks_node_group.workers["web"] must be replaced
  # module.eks.module.node_groups.random_pet.node_groups["jobs"] must be replaced
  # module.eks.module.node_groups.random_pet.node_groups["web"] must be replaced
```

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Create a cluster with node groups, and then change a base level tag.

## What's the expected behavior?

Tags are updated in place.

## Are you able to fix this problem and submit a PR? Link here if you have already.

No.

## Environment details

* Affected module version: 11.0.0
* OS: OS-X 10.14.6
* Terraform version: 0.12.10

dpiddockcmp:
Yes, this is unfortunate behaviour with how everything was stitched together to ensure dependency ordering in the one-shot module.

From my understanding of how Terraform works, this is roughly the reason behind why things gets recreated:
- The `aws_eks_cluster` resource is now in an unknown state at refresh time due to pending changes
- One of the `null_data_source.node_groups` data source's inputs is now "unknown" and so it can't be refreshed at "refresh" time. The refresh is delayed until apply time. Its outputs are now "unknown". This data source is required to enforce dependency between the `eks_node_group` and other required resources
- `node_groups.random_pet.node_groups` has an unknown input, so is tagged for updating
- `node_groups.aws_eks_node_group.workers` thus get tagged for recreation as its name will be changed

I don't think there's much we can do design-wise in the module to work around this particular issue at this time. Terraform 0.13 is scheduled to get a bunch of dependency improvements around modules: https://github.com/hashicorp/terraform/issues/17101 . `depends_on` against variables would be the magic feature.

[...]

Issue #852, opened 2020-04-28
configmaps "aws-auth" already exists


moh-abk:
# I have issues

## I'm submitting a...

* [x] bug report

## What is the current behavior?

When deploying a cluster and using only managed node_groups I believe because they're managed, AWS creates the `aws-auth` automatically and joins them to the cluster.

This means that terraform throws the error `configmaps "aws-auth" already exists`. So the `kubernetes_config_map` should update and not throw an error saying the configmap already exists

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Deploy the cluster using `managed node_groups`.

## What's the expected behavior?

`aws-auth` config map should not already exist. `kubernetes_config_map` should apply/update-in-place `aws-auth`

## Are you able to fix this problem and submit a PR? Link here if you have already.

N/A

## Environment details

* Affected module version: v11.1.0
* OS: Linux
* Terraform version: 0.12.24

dpiddockcmp:
Which version of the module are you using?

This shouldn't be possible if you have `manage_aws_auth = true` (the default) since v8.0.0.

What is your full `module` block that produces the problem?

moh-abk:
`module version: v11.1.0`

Curious how `manage_aws_auth = true` is telling AWS not to automatically create the `aws-auth` config map?

You can see in the guide here - https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html

In the `Self-managed nodes` tab - AWS says you should `kubectl apply -f aws-auth-cm.yaml`
In the `Amazon EKS managed node groups` tab - there's no mention of it so it's automatically done

moh-abk:
To reproduce this;

- Create eks cluster with self-managed nodes
`kubectl --kubeconfig=kubeconfig_* -n kube-system get cm aws-auth -o yaml`

```
Error from server (NotFound): configmaps "aws-auth" not found
```

- Create eks cluster with no `self-managed nodes` only `managed node_groups` and `manage_aws_auth = false`
`kubectl --kubeconfig=kubeconfig_* -n kube-system get cm aws-auth -o yaml`

```
apiVersion: v1
data:
  mapRoles: |
    - groups:
      - system:bootstrappers
      - system:nodes
      rolearn: arn:aws:iam::<REDACTED>:role/<REDACTED>
      username: system:node:{{EC2PrivateDNSName}}
kind: ConfigMap
metadata:
  creationTimestamp: "2020-04-28T11:32:48Z"
[...]

PR #867, opened 2020-05-06
feat: Add a homemade `depends_on` for MNG submodule to ensure ordering of resource creation

barryib:
# PR o'clock

## Description

Hack to add a "homemade" depends_on as suggested https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2 to ensure ordering of ressource creation for managed node groups.

I don't really like the approach, but the hack doesn't add complexity (because, there are only 2 resources in the submodule in which we add an explicit `depends_on`) and seems to work as expected.

I tested it to resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/843

__Note:__ I'm opening this PR to start discussion. So feedbacks are welcomed.

### Questions

- Is this something we want to do ?
- @dpiddockcmp is your [comment](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L10) still apply ?

### Checklist

- [x] CI tests are passing

### DIFF
--- a/modules/node_groups/variables.tf
+++ b/modules/node_groups/variables.tf
@@ -34,3 +34,11 @@ variable "node_groups" {
   type        = any
   default     = {}
 }
+
+# Hack for a homemade `depends_on` https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2
+# Will be removed in Terraform 0.13 with the support of module's `depends_on` https://github.com/hashicorp/terraform/issues/10462
+variable "ng_depends_on" {
+  description = "List of references to other resources this submodule depends on"
+  type        = any
+  default     = null
+}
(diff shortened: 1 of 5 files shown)


barryib:
closing in favor of https://github.com/terraform-aws-modules/terraform-aws-eks/pull/868

modules/node_groups/variables.tf
```diff
@@ -34,3 +34,11 @@ variable "node_groups" {
   type        = any
   default     = {}
 }
+
+# Hack for a homemade `depends_on` https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2
+# Will be removed when Terraform will add support of module's `depends_on`https://github.com/hashicorp/terraform/issues/10462
```
dpiddockcmp, line 39: The module cannot depend on the aws_auth. Output from the module is used to generate the configmap's roles key.

barryib:
> Please could we reconsider this?
> 
> It is a more terraform native way of ensuring dependency between resources.
> 
> I really regret the null_resource hack as I didn't know that you could just depend on variables! We're seeing this horror being repeated in the PRs for moving workers to a module and fargate support.

Yes of course. I will rebase my branch.

modules/node_groups/variables.tf
barryib: fixed

barryib: Humm after reading some issues, I think your comment https://github.com/terraform-aws-modules/terraform-aws-eks/issues/852#issuecomment-621076125 still apply. Terraform won't go into dependency cycle because the [`aws_auth_roles` output from MNG](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/node_groups/outputs.tf#L6-L14) is built with locals.

PR #875, opened 2020-05-08
fix: Fix Launch Templates error with aws 2.61.0

### LINKED ISSUE #874: ASG Failed scale up after update aws_launch_template with aws provider 2.61.0 (currently latest version)
# I have an issue when I update aws_launch_template with [provider.aws=2.61.0](https://github.com/terraform-providers/terraform-provider-aws/releases/tag/v2.61.0). 

## I'm submitting a...

* [x] bug report
* [ ] feature request

The latest terraform provider rolled out a new improvement with in [MR](https://github.com/terraform-providers/terraform-provider-aws/pull/11655). It will set "partition_number" to 0 when you not handle it at [placement](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers_launch_template.tf#L297). This will cause ASG failed to launch instance with error 
`Launching a new EC2 instance. Status Reason: Could not launch Spot Instances. InvalidParameterValue - You cannot use PartitionNumber with a Placement Group that does not exist. Specify a valid Placement Group and try again. Launching EC2 instance failed.`

Use aws provider 2.61.0 (currently latest version), update any launch template and apply

ASG will failed to scale up

NA
## Environment details

* Affected module version:
* OS: mac
* Terraform version: 0.11.14

dpiddockcmp:
# PR o'clock

## Description

Updating Launch Templates with aws provider 2.61.0 generates an error when creating the ASG or attempting to launch new instances. This is caused by the aws provider adding support to the launch template for `placement.partition_number`. I think the root issue is go-aws-sdk defaulting this to 0 as it's an integer and the module setting the placement block when no placement group is in use. Having partition_number set but no placement group name is a validation error. For example when creating a new ASG:
`Error: Error creating AutoScaling Group: ValidationError: You must use a valid fully-formed launch template. You cannot use PartitionNumber with a Placement Group that does not exist. Specify a valid Placement Group and try again.`

Resolve the issue by not setting the placement block if no placement group name is passed in.

This will cause an update for all users using launch templates unless they are using placement groups.

Fixes #874 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -68,7 +68,7 @@ locals {
     suspended_processes           = ["AZRebalance"]             # A list of processes to suspend. i.e. ["AZRebalance", "HealthCheck", "ReplaceUnhealthy"]
     target_group_arns             = null                        # A list of Application LoadBalancer (ALB) target group ARNs to be associated to the autoscaling group
     enabled_metrics               = []                          # A list of metrics to be collected i.e. ["GroupMinSize", "GroupMaxSize", "GroupDesiredCapacity"]
-    placement_group               = ""                          # The name of the placement group into which to launch the instances, if any.
+    placement_group               = null                        # The name of the placement group into which to launch the instances, if any.
     service_linked_role_arn       = ""                          # Arn of custom service linked role that Auto Scaling group will use. Useful when you have encrypted EBS
     termination_policies          = []                          # A list of policies to decide how the instances in the auto scale group should be terminated.
     platform                      = "linux"                     # Platform of workers. either "linux" or "windows"
@@ -78,7 +78,7 @@ locals {
     root_kms_key_id                   = ""                                       # The KMS key to use when encrypting the root storage device
     launch_template_version           = "$Latest"                                # The lastest version of the launch template to use in the autoscaling group
     launch_template_placement_tenancy = "default"                                # The placement tenancy for instances
-    launch_template_placement_group   = ""                                       # The name of the placement group into which to launch the instances, if any.
+    launch_template_placement_group   = null                                     # The name of the placement group into which to launch the instances, if any.
     root_encrypted                    = false                                    # Whether the volume should be encrypted or not
     eni_delete                        = true                                     # Delete the Elastic Network Interface (ENI) on termination (if set to false you will have to manually delete before destroying)
     cpu_credits                       = "standard"                               # T2/T3 unlimited mode, can be 'standard' or 'unlimited'. Used 'standard' mode as default to avoid paying higher costs
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -294,17 +294,17 @@ resource "aws_launch_template" "workers_launch_template" {
     )
   }
 
-  placement {
-    tenancy = lookup(
-      var.worker_groups_launch_template[count.index],
-      "launch_template_placement_tenancy",
-      local.workers_group_defaults["launch_template_placement_tenancy"],
-    )
-    group_name = lookup(
-      var.worker_groups_launch_template[count.index],
-      "launch_template_placement_group",
-      local.workers_group_defaults["launch_template_placement_group"],
-    )
+  dynamic placement {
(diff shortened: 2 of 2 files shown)


riveryc:
NICE, this fixed my issue!

dpiddockcmp:
I've done a little more testing and this fixes launch templates for people who do not use placement groups.

However there's a bug in the provider or SDK. Placement groups specified directly in the launch template only work if the placement group is of type partition because `partition_number` is set. Except we don't expose the ability to set the partition number. I don't want to do that as it will require a bump to the broken provider version 2.61.0.

Non-partition placement groups can still be specified on the ASG via `placement_group` variable.

```
module "eks" {
  # ...
    {
      instance_type        = "m5.large"
      asg_desired_capacity = 1
      public_ip            = true
      placement_group      = aws_placement_group.this.id
    },
    {
      name                            = "worker-group-2"
      instance_type                   = "m5.large"
      asg_desired_capacity            = 1
      public_ip                       = false
      launch_template_placement_group = aws_placement_group.this.id
    },
    {
      name                 = "worker-group-3"
      asg_desired_capacity = 1
      public_ip            = false
    },
  ]
}
resource "aws_placement_group" "this" {
  name     = "testgroup"
  strategy = "cluster"
}
```
`worker-group-1` and `worker-group-3` are created without issue. `worker-group-2` fails with:
```
Error: Error creating AutoScaling Group: ValidationError: You must use a valid
fully-formed launch template. You cannot use PartitionNumber with a Placement
Group that does not use the 'partition' strategy. Placement Group 'testgroup'
uses 'cluster' strategy. Specify a different Placement Group and try again.
```

I guess users wanting to set a non-partition placement group on the launch template will have to pin to aws 2.60.0 until it's fixed?

barryib:
> I guess users wanting to set a non-partition placement group on the launch template will have to pin to aws 2.60.0 until it's fixed?

Can we just pin this in [versions.tf](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/versions.tf#L5) ? Something like `aws = ">= 2.52.0,<= 2.60.0"` or `aws = "~> 2.60.0"` until it's fixed upstream ?

dpiddockcmp:
We could but I don't feel that it is the correct solution in this situation. The remaining issue is a provider/sdk bug. By pinning to < 2.61 we stop users from being able to use future fixed versions. There are other fixes and features in the provider.

barryib:
Agree in general. But it's not a definitive pinning. I was thinking to pin the module version temporally, like we already did with the kubernetes provider.

Furthermore, it we should consider bumping the aws provider version. Maybe after https://github.com/terraform-providers/terraform-provider-aws/pull/13239 is merged ?

barryib:
@dpiddockcmp do you still want to merge this PR ?

kmgreen2:
Thanks!  Worked for me as well!

dpiddockcmp:
Kubernetes provider was slightly different as it broke default functionality used by probably the majority. At the time it was also the "expected behaviour" from the provider maintainer.

This PR fixes the issue for probably the majority of launch template users. There's a PR on the provider that fixes the real issue. Hopefully it gets released in the normal cycle on Thursday.

I think we should merge this and get 12.0.0 released.

Murty0:
I am experiencing the same issue with `aws provider version "~> 2.59"` and `eks module version "~> 10.0"`. 

We already have a cluster running in production which did not have a placement group, but is working fine. I am in the midst of creating a testing cluster, with the above mentioned provider versions, and again with no placement groups defined. But I am getting the following errors:

```
Error: Error creating AutoScaling Group: ValidationError: You must use a valid fully-formed launch template. You cannot use PartitionNumber with a Placement Group that does not exist. Specify a valid Placement Group and try again.
  status code: 400, request id: 8760f1ce-69ea-4457-83a3-37eee9e11804

  on .terraform/modules/eks_cluster/terraform-aws-eks-10.0.0/workers_launch_template.tf line 3, in resource "aws_autoscaling_group" "workers_launch_template":
   3: resource "aws_autoscaling_group" "workers_launch_template" {
```

I have already posted this here: https://github.com/terraform-providers/terraform-provider-aws/issues/13236

PR #880, opened 2020-05-14
fix: Ensure kubeconfig ends with \n

graywolf-at-work:
In order to be a text file according to POSIX, file needs to be composed of
text lines. Text line is defined as sequence of characters ending in \n. Sadly,
`~}` did strip everything including the \n, so the kubeconfing did not end
with a new line. Output empty string at the end to make sure of it.

# PR o'clock

## Description

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/templates/kubeconfig.tpl
+++ b/templates/kubeconfig.tpl
@@ -35,4 +35,4 @@ users:
         - name: ${k}
           value: ${v}
   %{~ endfor ~}
-%{ endif ~}
+%{ endif }


templates/kubeconfig.tpl
```diff
@@ -36,3 +36,4 @@ users:
           value: ${v}
   %{~ endfor ~}
 %{ endif ~}
+${""}
```
dpiddockcmp, line 39: It would be cleaner to just remove the ~ from the end of that last `endif` statement. That is what is removing the terminating whitespace.

graywolf-at-work: Done.

barryib:
@graywolf-at-work Thanks for your contribution. Can you please update your branch, so I can merge it ?

graywolf-at-work:
I've rebased on current master.

PR #883, opened 2020-05-15
fix: Revert removal of templates provider

### LINKED ISSUE #882: 12.0.0 broke userdata_template_file in worker_groups
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request

worker_groups key `userdata_template_file` used to contain the content of a template file. #854 accidentally changed this to being a filename

As reported here: https://github.com/terraform-aws-modules/terraform-aws-eks/commit/b183b972539fa9d39dc55c453d68a234fdb683b1#r39182495

Try to set worker_groups["userdata_template_file"] to the contents of a userdata file. It'll explode trying to find a file of that name

Input passed as raw data to be used as the userdata template.

Later

## Environment details

* Affected module version: 12.0.0
* OS: All
* Terraform version: 0.12.24

## Any other relevant info

My bad 😞

dpiddockcmp:
# PR o'clock

## Description

Revert removing the templates provider. It broke some workflows due to previously being able to pass in the raw contents of a template file for processing.

Unfortunately the Terraform built in `templatefile()` function requires use with a file that exists before running the plan. So we're stuck with using `template_file` data source and all the obnoxious random updates it causes for now.

Fixes #882 

### Checklist

- [x] CI tests are passing


mbarrien:
Ping? I'm still stuck pinning both this version and the Terraform AWS provider version because this regression is included in the same release as #875 which I need sooner than later. Trying to avoid forking this module just to be able to get #875 because I'm feeling the pain right now (needing some fixes in more recent AWS providers in unrelated areas). It'll be a LONG time before 0.13.0 final is out (it's unacceptable to force us to use a beta version) and just waiting months for the fix is not acceptable.

If #875 were in a separate version, I would've pinned on that, but since these 2 are tied together, I can't.

barryib:
@dpiddockcmp @mbarrien I'm going to test this. Sorry for the delay. A week off without internet.

PR #889, opened 2020-05-22
fix: Change the order of the worker tags

dwardu89:
# PR o'clock

## Description

When you spin up a terraform worker group, it will create everything, however when you run a subsequent `terraform plan` you get an ordering change of the tags. This will just make sure that the tag is created in the correct order.
The problem will give a plan output like this. This will fix it.

```
      ~ tags                      = [
            {
                "key"                 = "Name"
                "propagate_at_launch" = "true"
                "value"               = "k8s-dev-spot-2-eks_asg"
            },
          - {
              - "key"                 = "k8s.io/cluster/k8s-dev"
              - "propagate_at_launch" = "true"
              - "value"               = "owned"
            },
            {
                "key"                 = "kubernetes.io/cluster/k8s-dev"
                "propagate_at_launch" = "true"
                "value"               = "owned"
            },
          + {
              + "key"                 = "k8s.io/cluster/k8s-dev"
              + "propagate_at_launch" = "true"
              + "value"               = "owned"
            },
            {
                "key"                 = "stack-name"
                "propagate_at_launch" = "true"
                "value"               = "k8s-dev"
            },
        ]
```

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/workers.tf
+++ b/workers.tf
@@ -110,12 +110,12 @@ resource "aws_autoscaling_group" "workers" {
         "propagate_at_launch" = true
       },
       {
-        "key"                 = "kubernetes.io/cluster/${aws_eks_cluster.this[0].name}"
+        "key"                 = "k8s.io/cluster/${aws_eks_cluster.this[0].name}"
         "value"               = "owned"
         "propagate_at_launch" = true
       },
       {
-        "key"                 = "k8s.io/cluster/${aws_eks_cluster.this[0].name}"
+        "key"                 = "kubernetes.io/cluster/${aws_eks_cluster.this[0].name}"
         "value"               = "owned"
         "propagate_at_launch" = true
       },


dpiddockcmp:
This is a new change. I've just hit the same thing when making unrelated changes in an existing cluster.

Why have AWS or the provider stopped maintaining the order of the tags array? It's bad enough that we have to use this horrid data structure. The same thing happened to alternative names in ACM certificates and it's a nightmare to deal with.

dpiddockcmp:
I wonder if it's related to https://github.com/terraform-providers/terraform-provider-aws/pull/13360 that got released today in 2.63.0?
```
Terraform v0.12.24
+ provider.aws v2.63.0
+ provider.local v1.4.0
+ provider.null v2.1.2
+ provider.template v2.1.2
```

dayglojesus:
> I wonder if it's related to [terraform-providers/terraform-provider-aws#13360](https://github.com/terraform-providers/terraform-provider-aws/pull/13360) that got released today in 2.63.0?
> 
> ```
> Terraform v0.12.24
> + provider.aws v2.63.0
> + provider.local v1.4.0
> + provider.null v2.1.2
> + provider.template v2.1.2
> ```

It is 💯 related. We do regular audits of our TF envs. We have hundreds of `aws_autoscaling_group` resources in our environment and ALL of them are now producing **diffs** on ASG tags -- even after performing an `terraform apply` on any given stack.

#13360 broke the idempotency of the `aws_autoscaling_group` resource.

PS
I am not even sure what #13360 was supposed to fix, `propagate_at_launch` is working in the previous version of the provider and has always worked as far back I can remember. 🤷‍♂️

dpiddockcmp:
Created provider issue https://github.com/terraform-providers/terraform-provider-aws/issues/13469

Looking at the diff I don't understand how this has broken things. I guess it's deep terraform internals or a weird side affect of converting a bool to a string. Must break the state comparison system.

dayglojesus:
You're a ⭐  @dpiddockcmp -- thanks!

dwardu89:
@dpiddockcmp That is probably the problem, I think for now we would have to either pin the version of the AWS provider or fix this tag problem here. Unless AWS releases a quick fix soon.

dpiddockcmp:
The fix has been merged. It was released yesterday in aws provider 2.64.0.

I'm going to close this PR as unnecessary now 😄

PR #890, opened 2020-05-24
improvement: cluster_id output blocks on wait_for_cluster

dpiddockcmp:
# PR o'clock

## Description

Make the `cluster_id` module output block on the `null_resource.wait_for_cluster`. Users do not need to re-implement the null resource if they also want to create kubernetes resources in the same configuration.

This won't do anything for users with `manage_aws_auth = false`

From the rejected #884. 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -1,6 +1,9 @@
 output "cluster_id" {
   description = "The name/id of the EKS cluster."
   value       = element(concat(aws_eks_cluster.this.*.id, list("")), 0)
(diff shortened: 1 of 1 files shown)


ibratoev:
I was just looking how to fix that and found this PR. 👍

ibratoev:
Alternatively, `wait_for_cluster` can be exposed as output so that external resources have control over waiting.

PR #892, opened 2020-05-26
feat: Allow communication between pods on workers and pods using the primary cluster security group

itssimon:
# PR o'clock

## Description

Adds two security group rules allowing pods running on worker nodes to communicate with pods using the primary cluster security group (such as pods running on Fargate).

See https://github.com/terraform-aws-modules/terraform-aws-eks/pull/866#discussion_r428140829.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -8,10 +8,11 @@ locals {
     )
   ]
 
-  cluster_security_group_id = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
-  cluster_iam_role_name     = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
-  cluster_iam_role_arn      = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
-  worker_security_group_id  = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
+  cluster_security_group_id         = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
+  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : ""
+  cluster_iam_role_name             = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
+  cluster_iam_role_arn              = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
+  worker_security_group_id          = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
 
   default_iam_role_id = concat(aws_iam_role.workers.*.id, [""])[0]
   kubeconfig_name     = var.kubeconfig_name == "" ? "eks_${var.cluster_name}" : var.kubeconfig_name
--- a/local.tf
+++ b/local.tf
@@ -9,7 +9,7 @@ locals {
   ]
 
   cluster_security_group_id         = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
-  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : ""
+  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : null
   cluster_iam_role_name             = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
   cluster_iam_role_arn              = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
   worker_security_group_id          = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
--- a/workers.tf
+++ b/workers.tf
@@ -350,7 +350,7 @@ resource "aws_security_group_rule" "workers_ingress_cluster_primary" {
 }
 
 resource "aws_security_group_rule" "cluster_primary_ingress_workers" {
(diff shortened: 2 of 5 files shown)


dpiddockcmp:
This is a potentially breaking change as it may conflict with rules put in place by users already running mixed clusters. I think there should be a variable, default false, to enable creation of these rules. We can then remove it in a future major release.

itssimon:
Agree @dpiddockcmp and done.

workers.tf
```diff
@@ -338,6 +338,28 @@ resource "aws_security_group_rule" "workers_ingress_cluster_https" {
   type                     = "ingress"
 }
 
+resource "aws_security_group_rule" "workers_ingress_cluster_primary" {
+  count                    = var.worker_create_security_group && var.worker_create_cluster_primary_security_group_rules && local.cluster_primary_security_group_id != "" && var.create_eks ? 1 : 0
```
dpiddockcmp, line 342: These count arguments do not work on an empty state. `local.cluster_primary_security_group_id != ""` is unknown so count cannot be calculated. Replace it with `var.cluster_version >= 1.14`.

```
Error: Invalid count argument

  on ../../workers.tf line 353, in resource "aws_security_group_rule" "cluster_primary_ingress_workers":
 353:   count                    = var.worker_create_security_group && var.worker_create_cluster_primary_security_group_rules && local.cluster_primary_security_group_id != "" && var.create_eks ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
resources that the count depends on.
```

local.tf
```diff
@@ -8,10 +8,11 @@ locals {
     )
   ]
 
-  cluster_security_group_id = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
-  cluster_iam_role_name     = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
-  cluster_iam_role_arn      = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
-  worker_security_group_id  = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
+  cluster_security_group_id         = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
+  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : ""
```
dpiddockcmp, line 12: Else value of this should stay as `null`

workers.tf
dpiddockcmp: This line also needs updating

barryib:
```
Error: Provider produced inconsistent final plan

When expanding the plan for
module.eks.aws_security_group_rule.cluster_primary_ingress_workers[0] to
include new values learned so far during apply, provider
"registry.terraform.io/-/aws" produced an invalid new value for
.security_group_id: was cty.StringVal(""), but now
cty.StringVal("sg-0892893c8410071f4").

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

Means that `module.eks.aws_security_group_rule.cluster_primary_ingress_workers[0].security_group_id` was at `""` during the plan phase and became `sg-0892893c8410071f4` during the apply.

I think it's because of :

```hcl
cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : null
```
[`local.cluster_primary_security_group_id`](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/892/files#diff-0973ac2dc1229ea1662319bdc6954f08R12) is empty for all clusters created before [Kubernetes 1.14 and eks.3](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) during the plan phase. So during the apply phase, the cluster is updated and the primary security group is created by AWS. This is probably the inconsistency.

You can notice with a terraform plan, that there is no `security_group_id` for the `cluster_primary_ingress_workers` SG rule.

```
  # module.eks.aws_security_group_rule.cluster_primary_ingress_workers[0] will be created
  + resource "aws_security_group_rule" "cluster_primary_ingress_workers" {
      + description              = "Allow pods running on workers to send communication to cluster primary security group (e.g. Fargate pods)."
      + from_port                = 0
      + id                       = (known after apply)
      + protocol                 = "-1"
      + self                     = false
      + source_security_group_id = "sg-0d741dba799316523"
      + to_port                  = 65535
      + type                     = "ingress"
    }
```

So we have to ensure a correct ordering during resources creation to let Terraform wait for the SG creation or at least mark the `security_group_id` as unknown.

Here how I make it work (but I don't know if it brokes something else).

```hcl
locals {
  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? coalescelist(data.null_data_source.cluster[*].outputs["cluster_primary_security_group_id"], [""])[0] : null
}

data "null_data_source" "cluster" {
  count = var.create_eks ? 1 : 0

  inputs = {
    cluster_primary_security_group_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
  }
}
```
With the null_data_source, Terraform wait for the `cluster_security_group_id` attribute to be available.

You can see the difference with a plan (`security_group_id = (known after apply)`)

```
  # module.eks.aws_security_group_rule.cluster_primary_ingress_workers[0] will be created
  + resource "aws_security_group_rule" "cluster_primary_ingress_workers" {
      + description              = "Allow pods running on workers to send communication to cluster primary security group (e.g. Fargate pods)."
      + from_port                = 0
      + id                       = (known after apply)
      + protocol                 = "-1"
      + security_group_id        = (known after apply)
      + self                     = false
      + source_security_group_id = "sg-01ee228ae43275c29"
      + to_port                  = 65535
      + type                     = "ingress"
    }
```
This will also make the output `cluster_primary_security_group_id` available only when the SG (cluster_security_group_id) is created by AWS.

@dpiddockcmp @itssimon Is this work for you ? Do you see any downside or side effects of doing this ?

@dpiddockcmp I assigned this PR to you, so you can squash and merge it when it's ready ^^

dpiddockcmp:
I don't think the added complexity of using a `null_data_source` here is necessary. I also tried removing the `local` so that the security groups were directly referencing `aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id`. Same result when updating a 1.13 cluster but only for the `aws_security_group_rule.cluster_primary_ingress_workers` resource. The other applied correctly.

I guess don't enable this on a 1.13 cluster 😉

barryib:
> I don't think the added complexity of using a `null_data_source` here is necessary. I also tried removing the `local` so that the security groups were directly referencing `aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id`. Same result when updating a 1.13 cluster but only for the `aws_security_group_rule.cluster_primary_ingress_workers` resource. The other applied correctly.
> 
Ohh that was the first thing I've tried, but no luck. Maybe I didn't test it correctly ^^

PR #908, opened 2020-06-05
fix: Get `on_demand_allocation_strategy` from `local.workers_group_defaults` when deciding to use `mixed_instances_policy`

trnubo:
Get on_demand_allocation_strategy from local.workers_group_defaults when deciding to use mixed_instances_policy

# PR o'clock

## Description

This fix uses the default from `local.workers_group_defaults` instead of `null` (the workers_group_defaults_defaults is `null` anyway) when the dynamic blocks `mixed_instances_policy` and `launch_template` are being decided on.

We use many similar worker groups (worker group per AZ, on-demand and spot groups) via worker_groups_launch_template and in an attempt to reduce duplicated config between worker groups we use workers_group_defaults. However setting on_demand_allocation_strategy and override_instance_types in the defaults doesn't have any effect when on_demand_allocation_strategy isn't set in each worker group.

This fix allows `on_demand_allocation_strategy` to be set in `workers_group_defaults` which will enable the `mixed_instances_policy` block and use the default `override_instance_types` even when neither are defined in the worker group. I don't see the default `on_demand_allocation_strategy` used anywhere else.

### Checklist

- [ ] CI tests are passing


dpiddockcmp:
Should the `on_demand_allocation_strategy` block around line 100 do a lookup for local.workers_group_defaults["on_demand_allocation_strategy"]? Although at the moment the only valid value is `prioritized`

PR #937, opened 2020-06-29
docs: add example for node draining

karlderkaefer:
# PR o'clock

## Description

Im not sure if this fits into the repo. I did some research on the topic and want to share a possible solution. Im not sad if you reject this but still happy about any feedback.
There is an existing issue for that topic here https://github.com/terraform-aws-modules/terraform-aws-eks/issues/462
If you run a production cluster with this module an important feature is missing.
How to update eks worker kubernetes version or worker group variables without downtime?

One way to solve this is to create a lifecycle hook which will drain the node before shutting down instances.

### Checklist

- [ ] CI tests are passing


max-rocket-internet:
Yes, I think this PR is a bit much for this repo 😅  If we have this python code in here then we will also likely get PRs and issues for it, which I don't think we would be keen for.

@karlderkaefer you seem to have it nicely packaged together so why don't you create a standalone repo for this?

karlderkaefer:
Sry for late response. Thanks for response. I totally agree with your points, I close this PR. The reference repo can be found here although it's not 100% finished yet https://github.com/karlderkaefer/eks-node-drainer but at least it gives some ideas.

PR #938, opened 2020-06-30
feat: :lock: IDMSv2 metadata restrictions

danquack:
# Metadata Options for Launch Template

## Description
Adding metadata options to the launch template options, which will allow for IDMS v2 restrictions: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

### Checklist

- [x] CI tests are passing


danquack:
@barryib @brandonjbjelland @bryantbiggs It wasnt apparent in the contributing guidelines who would review this, but I'm wondering if someone could take a look

Issue #945, opened 2020-07-10
Double Name tags on asg workers


michaelPotter:
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

Terraform tries to add two Name tags to autoscaling group workers on _every_ apply.

It looks like the module applies a custom Name tag to autoscaling group workers in addition to the tags passed into the module. Terraform tries to apply two Name tags to the worker resources, and fails silently. During the next plan phase, TF will try to apply double Name tags again.

This clutters up the plan output since it tries to add on every apply.

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Pass a custom Name tag to the module

## What's the expected behavior?

I think the module should apply only the internal version of the tag (or replace the internal Name tag with something else.)

## Are you able to fix this problem and submit a PR? Link here if you have already.

Pull Request: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/946

## Environment details

* Affected module version:
* OS:
* Terraform version:

## Any other relevant info

Here's an example of the output I see on EVERY apply

```
  # module.kubernetes.module.eks.aws_autoscaling_group.workers[0] will be updated in-place
  ~ resource "aws_autoscaling_group" "workers" {
        arn                       = "arn:aws:autoscaling:us-west-2:xxxx:autoScalingGroup:173f87c5-5555-4b83-b90b-c3f04e9cb21f:autoScalingGroupName/group120200526193303054200000002"
        default_cooldown          = 300
        desired_capacity          = 1
        enabled_metrics           = []
        ...
      ~ tags                      = [
            ...
          + {
              + "key"                 = "Name"
              + "propagate_at_launch" = "true"
              + "value"               = "project-dev"
            },
            {
                "key"                 = "Name"
                "propagate_at_launch" = "true"
                "value"               = "project-dev-group1-eks_asg"
            },
        ]
        ...
    }
```

acarsercan:
I'm seen the same problem

dpiddockcmp:
Fixed in #946

PR #946, opened 2020-07-10
fix: Strip user Name tag from asg_tags (#945)

michaelPotter:
# PR o'clock
See issue #945 

## Description

I added a call to strip out any user given Name tags from `locals.asg_tags` so only the internal Name tags will be applied to asg workers. It looks like this var is only used for two resources, and both of them also try to apply their own Name tags, so there shouldn't be any fallout from dropping it.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -1,6 +1,6 @@
 locals {
   asg_tags = [
-    for item in keys(var.tags) :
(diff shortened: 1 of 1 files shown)


michaelPotter:
Aww zoot

PR #963, opened 2020-07-30
fix: #961 - Allow running cloud-init boothook commands

### LINKED ISSUE #961: Support for boothook user-data script
## I'm submitting a...

* [ X ] feature request

I can currently set a pre_userdata that will be used in this template to generate a user-data bash script:

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/templates/userdata.sh.tpl

But I need to set a [boothook script](https://cloudinit.readthedocs.io/en/latest/topics/format.html#cloud-boothook) to change settings before docker and kubelet are started for the first time, in addition to the bash script to bootstrap the instance.

Specifically, I'd like to format and mount an nvme volume, and use it for /var/lib/docker on a c5d instance to avoid [soft lockup errors](https://github.com/awslabs/amazon-eks-ami/issues/454) on high disk I/O workloads.

Maybe a non-empty boothook variable would point to a [multipart mime](https://cloudinit.readthedocs.io/en/latest/topics/format.html#helper-script-to-generate-mime-messages) template file here: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/data.tf#L64

Please let me know if this would work and I'll make a PR.

## Environment details

* Affected module version:
* OS: AL2
* Terraform version:

raonitimo:
#  Description

Fixes #961 by changing user-data template into a [Multipart MIME format](https://cloudinit.readthedocs.io/en/latest/topics/format.html#mime-multi-part-archive) that includes a cloud-init [boothook section](https://cloudinit.readthedocs.io/en/latest/topics/format.html#cloud-boothook) to allow users running commands early in the boot process.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -72,6 +72,11 @@ data "template_file" "userdata" {
     cluster_name        = coalescelist(aws_eks_cluster.this[*].name, [""])[0]
     endpoint            = coalescelist(aws_eks_cluster.this[*].endpoint, [""])[0]
     cluster_auth_base64 = coalescelist(aws_eks_cluster.this[*].certificate_authority[0].data, [""])[0]
+    boothook = lookup(
+      var.worker_groups[count.index],
+      "boothook",
+      local.workers_group_defaults["boothook"],
+    )
     pre_userdata = lookup(
       var.worker_groups[count.index],
       "pre_userdata",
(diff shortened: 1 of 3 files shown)


raonitimo:
This is actually unnecessary as we can use userdata_template_file and replace the whole thing.

Issue #974, opened 2020-08-12
Module Issues w/ Terraform v0.13.0


cazter:
# After upgrading to Terraform v0.13.0, the module errors out with refresh/plan/apply.
- Upgraded to tf v0.13.0
- Ran `tf init`
- Backend stored in S3 bucket failed, requiring reconfigure. Ran `tf init -reconfigure`
- Running tf refresh/plan/apply fails with error from module (details below).

## I'm submitting a...
* [x] bug report

## What is the current behavior?
On tf refresh/plan/apply, I get the following error for each worker:

>         Error: Invalid resource instance data in state
> 
>           on ../../../modules/k8s/workers.tf line 390:
>          390: resource "aws_iam_instance_profile" "workers" {
> 
>         Instance module.eks.aws_iam_instance_profile.workers[4] data could not be
>         decoded from the state: unsupported attribute "roles".

From tfstate:

>       "module": "module.eks",
>       "mode": "managed",
>       "type": "aws_iam_instance_profile",
>       "name": "workers",
>       "each": "list",
>       "provider": "provider.aws",
>       "instances": [
>         {
>           "index_key": 0,
>           "schema_version": 0,
>           "attributes": {
>             "arn": "arn:aws:iam::xxxxxxxx:instance-profile/eks-cxxxxxxxx",
>             "create_date": "2019-07-13T16:40:52Z",
>             "id": "eks-cxxxxxxxx",
>             "name": "eks-cxxxxxxxx",
>             "name_prefix": "eks-c",
>             "path": "/",
>             "role": "eks-cxxxxxxxx",
>             "roles": [
>               "eks-cxxxxxxxx"
>             ],
>             "unique_id": "xxxxxxxx"
>           },
>           "private": "xxxxx",
>           "dependencies": [
>             "aws_security_group.eks-node-group-mgmt1",
>             "module.eks.aws_cloudwatch_log_group.this",
>             "module.eks.aws_eks_cluster.this",
>             "module.eks.aws_iam_role.cluster",
>             "module.eks.aws_iam_role.workers",
>             "module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSClusterPolicy",
>             "module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSServicePolicy",
>             "module.eks.aws_security_group.cluster",
>             "module.vpc.aws_vpc.eks"
>           ]
>         },

dpiddockcmp:
[...]

PR #976, opened 2020-08-14
fix: More compatibility fixes for Terraform v0.13 and aws v3

dpiddockcmp:
# PR o'clock

## Description

Terraform v0.13 and aws v3 support time!

- The update to the vpc module was, strictly speaking, unnecessary but it adds the `terraform` block with supported versions.
- Update for iam module was very necessary to support new versions
- Workaround for "Provider produced inconsistent final plan" when creating ASGs at the same time as the cluster. See  https://github.com/terraform-providers/terraform-provider-aws/issues/14085 for full details.

When upgrading an existing TF 0.12/aws 2 cluster to TF 0.13/aws 3 you currently have to manually edit the state file to delete removed attributes. Should be fixed in v0.13.1 https://github.com/hashicorp/terraform/pull/25779 . For the basic example I had to remove the following attributes:
- data `aws_availability_zones`: `blacklisted_names` and `blackedlisted_zone_ids`
- `aws_iam_instance_profile`: `roles` (#974)

You also need to mess around with `replace-provider` a lot:
```sh
terraform state replace-provider -- -/aws hashicorp/aws
terraform state replace-provider -- -/kubernetes hashicorp/kubernetes
terraform state replace-provider -- -/local hashicorp/local
terraform state replace-provider -- -/null hashicorp/null
terraform state replace-provider -- -/random hashicorp/random
terraform state replace-provider -- -/template hashicorp/template
```

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/workers.tf
+++ b/workers.tf
@@ -102,31 +102,38 @@ resource "aws_autoscaling_group" "workers" {
     }
   }
 
-  tags = concat(
-    [
-      {
-        "key"                 = "Name"
-        "value"               = "${aws_eks_cluster.this[0].name}-${lookup(var.worker_groups[count.index], "name", count.index)}-eks_asg"
-        "propagate_at_launch" = true
-      },
-      {
-        "key"                 = "kubernetes.io/cluster/${aws_eks_cluster.this[0].name}"
-        "value"               = "owned"
-        "propagate_at_launch" = true
-      },
-      {
-        "key"                 = "k8s.io/cluster/${aws_eks_cluster.this[0].name}"
-        "value"               = "owned"
-        "propagate_at_launch" = true
-      },
-    ],
-    local.asg_tags,
-    lookup(
-      var.worker_groups[count.index],
-      "tags",
-      local.workers_group_defaults["tags"]
+  dynamic "tag" {
+    for_each = concat(
+      [
+        {
+          "key"                 = "Name"
+          "value"               = "${aws_eks_cluster.this[0].name}-${lookup(var.worker_groups[count.index], "name", count.index)}-eks_asg"
+          "propagate_at_launch" = true
+        },
+        {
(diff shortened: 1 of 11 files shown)


workers.tf
```diff
@@ -102,31 +102,38 @@ resource "aws_autoscaling_group" "workers" {
...
-        "value"               = "owned"
-        "propagate_at_launch" = true
-      },
-    ],
-    local.asg_tags,
-    lookup(
-      var.worker_groups[count.index],
-      "tags",
-      local.workers_group_defaults["tags"]
+  dynamic "tag" {
```
barryib, line 105: I don't test it yet, but is there any reason to change `tags` into `dynamic "tag"` ?

barryib: Never mind. I read https://github.com/terraform-providers/terraform-provider-aws/issues/14085

daroga0002:
@dpiddockcmp  looks Terraform 13.1 was released so should solve mentioned bug

https://github.com/hashicorp/terraform/releases/tag/v0.13.1

dpiddockcmp:
@barryib are we good to merge this?

barryib:
Please let me test this tonight with TF 0.13.1

barryib (approved):

barryib:
@dpiddockcmp LGTM. Just tested an upgrade TF 0.12 to TF 0.13.1 without error. Shouldn't we change the Terraform required version to avoid TF 0.13.0 (`"~> 0.12.9, >= 0.13.1"`) ?

dpiddockcmp:
That version constraint can't be met. It can't be `0.12.*` and greater than `0.13.0`. Could use `>= 0.12.9, != 0.13.0`

barryib:
> That version constraint can't be met. It can't be `0.12.*` and greater than `0.13.0`. Could use `>= 0.12.9, != 0.13.0`

Ok. LGTM.

barryib (approved):

PR #977, opened 2020-08-15
fix: Remove conditionals from IAM policy attachments and documents

gswallow:
# fix: Remove conditionals for IAM policy attachments

## Description

The IAM policy attachments and documents are applied tolocal.cluster_iam_role_name, which is either the role created by this module or a role supplied by the module's user.  The conditional is the local variable; in either case, they should still be applied?

### Checklist

- [x] CI tests are passing
- [] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -122,13 +122,11 @@ resource "aws_iam_role" "cluster" {
 }
 
 resource "aws_iam_role_policy_attachment" "cluster_AmazonEKSClusterPolicy" {
-  count      = var.manage_cluster_iam_resources && var.create_eks ? 1 : 0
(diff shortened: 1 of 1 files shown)


dpiddockcmp:
The original justification for `var.cluster_iam_role_name` was from restrictive environments where the user is not able to create/modify IAM resources. This change would break in their environment.

Due to the logic these attachments only happen to the iam role created by the module, even with the `local` indirection.

gswallow:
Thanks.  I'll close this PR.

PR #981, opened 2020-08-18
fix: Create `cluster_private_access` security group rules when it should

### LINKED ISSUE #942: Private security rules not created when they should
# I have issues

## I'm submitting a...

* [X] bug report
* [ ] feature request
* [X] kudos, thank you, warm fuzzy

Thanks for your work! This module is awesome!

I'm having the next problem. When creating an EKS with both public and private endpoints, the security group rules are not added.

With this code:
```hcl

module "eks" {
  cluster_version = var.eks_version
  version         = "v12.1.0"
  subnets = concat(
    data.terraform_remote_state.vpc.outputs.private_subnets_ids,
    data.terraform_remote_state.vpc.outputs.public_subnets_ids,
  )
  tags                                  = local.tags
  vpc_id                                = data.terraform_remote_state.vpc.outputs.vpc_id
  worker_groups                         = local.worker_groups
  manage_aws_auth                       = true
  map_users                             = local.k8s_map_users
  cluster_create_security_group         = true
  cluster_endpoint_public_access        = true
  cluster_endpoint_public_access_cidrs  = local.endpoint_public_access_cidr
  cluster_endpoint_private_access       = true
  cluster_endpoint_private_access_cidrs = ["10.70.0.0/16"]
  cluster_enabled_log_types             = ["api", "audit", "authenticator", "controllerManager", "scheduler"]
}
```

I should be able to reach from the network `10.70.0.0/16` the k8s api. But I can't, since [this rule](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/9a0e548dcc9606ac79522363f26aacfc01647ff3/cluster.tf#L49) is not created. I know that's not created because I examine the output of terraform state:

```bash
terraform state list | grep cluster_private
```

I should be able to reach the api within a private subnet added to `cluster_endpoint_private_access_cidrs`.

I can't, since I don't know the implications of the change. But taking a look to the code, in [this line](cluster_endpoint_private_access_cidrs) you can see the next:

```hcl
resource "aws_security_group_rule" "cluster_private_access" {
  count       = var.create_eks && var.cluster_endpoint_private_access && var.cluster_endpoint_public_access == false ? 1 : 0
  type        = "ingress"
  from_port   = 443
  to_port     = 443
  protocol    = "tcp"
  cidr_blocks = var.cluster_endpoint_private_access_cidrs

  security_group_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
}
```

I think it should be changed to:

```hcl
resource "aws_security_group_rule" "cluster_private_access" {
  count       = var.create_eks && var.cluster_endpoint_private_access 
  type        = "ingress"
  from_port   = 443
  to_port     = 443
  protocol    = "tcp"
  cidr_blocks = var.cluster_endpoint_private_access_cidrs

  security_group_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
}
```

So the ternary operator should go out, since if you have a private endpoint, you want to modify it's SG, whether you have a public endpoint or not. If that makes sense, I can make a PR without problem, it's a small change.

## Environment details

* Affected module version: v12.1.0
* OS: Arch Linux
* Terraform version: v0.12.28

### LINKED ISSUE #980: Add cluster_create_security_group condition in private access security group
# I have issues

I can't avoid the creation of the security group named: `aws_security_group_rule.cluster_private_access` when I deploy an EKS cluster with `var.cluster_endpoint_private_access && var.cluster_endpoint_public_access == false`

## I'm submitting a...

* [X] bug report

In `12.1.0` I could avoid the creation of this security group, thanks to the `var.manage_aws_auth` variable.
In `12.2.0` , while solving the issue https://github.com/terraform-aws-modules/terraform-aws-eks/issues/914 , the PR https://github.com/terraform-aws-modules/terraform-aws-eks/pull/915 made impossible to disable the creation of this security group

If you try to upgrade from `12.1.0` to `12.2.0` , and if you have limited permissions (lack of VPC permissions) to your AWS account, you'll hit this error while trying to upgrade:

```
module.eks.aws_security_group_rule.cluster_private_access[0]: Creating...

Error: Error authorizing security group rule type ingress: UnauthorizedOperation: You are not authorized to perform this operation. Encoded authorization failure message: TXgpzCB6noNIFS4     [.....]  WVUtmemnuvt2UFhHAU
        status code: 403, request id: 5f92d5   [.....]    566

```

Allow users to skip the creation of this security group.

Yes: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/981

## Environment details

* Affected module version: `12.2.0`
* Terraform version: `v0.13.0`

## Any other relevant info

Discussion started here: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/915#issuecomment-675535037

anthonydahanne:
# PR o'clock

## Description

Add `cluster_create_security_group` condition to create the security group rule named `cluster_private_access`

Resolves #980
Resolves #942

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,13 +163,14 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | attach\_worker\_cni\_policy | Whether to attach the Amazon managed `AmazonEKS_CNI_Policy` IAM policy to the default worker IAM role. WARNING: If set `false` the permissions must be assigned to the `aws-node` DaemonSet pods via another method or nodes will not be able to join the cluster. | `bool` | `true` | no |
+| cluster\_create\_endpoint\_private\_access\_sg\_rule | Whether to create security group rules for the access to the Amazon EKS private API server endpoint. | `bool` | `false` | no |
 | cluster\_create\_security\_group | Whether to create a security group for the cluster or attach the cluster to `cluster_security_group_id`. | `bool` | `true` | no |
 | cluster\_create\_timeout | Timeout value when creating the EKS cluster. | `string` | `"30m"` | no |
 | cluster\_delete\_timeout | Timeout value when deleting the EKS cluster. | `string` | `"15m"` | no |
 | cluster\_enabled\_log\_types | A list of the desired control plane logging to enable. For more information, see Amazon EKS Control Plane Logging documentation (https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html) | `list(string)` | `[]` | no |
 | cluster\_encryption\_config | Configuration block with encryption configuration for the cluster. See examples/secrets\_encryption/main.tf for example format | <pre>list(object({<br>    provider_key_arn = string<br>    resources        = list(string)<br>  }))</pre> | `[]` | no |
 | cluster\_endpoint\_private\_access | Indicates whether or not the Amazon EKS private API server endpoint is enabled. | `bool` | `false` | no |
-| cluster\_endpoint\_private\_access\_cidrs | List of CIDR blocks which can access the Amazon EKS private API server endpoint, when public access is disabled | `list(string)` | <pre>[<br>  "0.0.0.0/0"<br>]</pre> | no |
+| cluster\_endpoint\_private\_access\_cidrs | List of CIDR blocks which can access the Amazon EKS private API server endpoint. | `list(string)` | `null` | no |
 | cluster\_endpoint\_public\_access | Indicates whether or not the Amazon EKS public API server endpoint is enabled. | `bool` | `true` | no |
 | cluster\_endpoint\_public\_access\_cidrs | List of CIDR blocks which can access the Amazon EKS public API server endpoint. | `list(string)` | <pre>[<br>  "0.0.0.0/0"<br>]</pre> | no |
 | cluster\_iam\_role\_name | IAM role name for the cluster. Only applicable if manage\_cluster\_iam\_resources is set to false. | `string` | `""` | no |
--- a/cluster.tf
+++ b/cluster.tf
@@ -48,7 +48,7 @@ resource "aws_eks_cluster" "this" {
 }
 
 resource "aws_security_group_rule" "cluster_private_access" {
-  count       = var.create_eks && var.cluster_endpoint_private_access && var.cluster_endpoint_public_access == false ? 1 : 0
+  count       = var.create_eks && var.cluster_create_endpoint_private_access_sg_rule && var.cluster_endpoint_private_access ? 1 : 0
   type        = "ingress"
   from_port   = 443
   to_port     = 443
--- a/variables.tf
+++ b/variables.tf
@@ -245,10 +245,16 @@ variable "iam_path" {
   default     = "/"
 }
 
+variable "cluster_create_endpoint_private_access_sg_rule" {
+  description = "Whether to create security group rules for the access to the Amazon EKS private API server endpoint."
+  type        = bool
+  default     = false
+}
+
 variable "cluster_endpoint_private_access_cidrs" {
-  description = "List of CIDR blocks which can access the Amazon EKS private API server endpoint, when public access is disabled"
+  description = "List of CIDR blocks which can access the Amazon EKS private API server endpoint."
   type        = list(string)
-  default     = ["0.0.0.0/0"]
+  default     = null
 }
 
 variable "cluster_endpoint_private_access" {


anthonydahanne:
as a FYI, I tried it out in my environment and it worked fine!

barryib:
Great. Can you please address this https://github.com/terraform-aws-modules/terraform-aws-eks/issues/942 in your PR also ?

anthonydahanne:
@barryib @alexppg : I've added this commit https://github.com/terraform-aws-modules/terraform-aws-eks/pull/981/commits/66a661a9141cdb1cc90aab6c37c8bb28143a1d73 , let me know if that solves #942

alexppg:
Nice, it does seem to fix it. I may be able to try it tomorrow. I'll let you know.

alexppg:
It does seem to create it in the plan. I can't apply it to test it right know, but when it's merged to master I'll do it eventually and let you know if there's a problem. Thanks!

barryib (approved):
LGTM.

Let's wait for @dpiddockcmp feedback.

dpiddockcmp:
I hate this damn rule 😆 It keeps generating issues. Plus it breaks the principle of least access by default. We sometimes create a rule that allows access from all IPs that have internal access to the VPC.

I don't see how this relates to `var.cluster_create_security_group` as it does not interact with that security group at all. This is a rule for the security group created by the EKS platform. Eventually we should re-work the cluster security group to be totally optional creation now that EKS creates one for us. That change would conflict with this rule.

Personally I would prefer a unique flag that explicitly enables this rule with a note that `var.cluster_endpoint_private_access_cidrs` probably needs modifying. Maybe we should add them to one of the examples?

barryib:
>I don't see how this relates to var.cluster_create_security_group as it does not interact with that security group at all. This is a rule for the security group created by the EKS platform.

I my mind, the `var.cluster_create_security_group` was a flag to allow the creation of all SG related to the EKS cluster (not only for a single SG) as is mentioned in the variable description **_Whether to create a security group for the cluster or attach the cluster to `cluster_security_group_id`_**. That's why I proposed that flag.

> Eventually we should re-work the cluster security group to be totally optional creation now that EKS creates one for us. That change would conflict with this rule.

Agree. I think it's time to rework the security group creation. This was partially discussed in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/799 and https://github.com/terraform-aws-modules/terraform-aws-eks/pull/792.

> Personally I would prefer a unique flag that explicitly enables this rule with a note that var.cluster_endpoint_private_access_cidrs probably needs modifying. Maybe we should add them to one of the examples?

Taking the decision on different flag sounds good to me. But instead of adding new variable, we can:
- change de default value `cluster_endpoint_private_access_cidrs` to `null`
- and add the SG only when someone define explicitly a list of private access CIDR. 

```hcl
variable "cluster_endpoint_private_access_cidrs" {
  description = "List of CIDR blocks which can access the Amazon EKS private API server endpoint, when public access is disabled"
  type        = list(string)
  default     = null
}

resource "aws_security_group_rule" "cluster_private_access" {
  count       = var.create_eks && var.cluster_endpoint_private_access_cidrs != null && var.cluster_endpoint_private_access ? 1 : 0
  type        = "ingress"
  from_port   = 443
  to_port     = 443
  protocol    = "tcp"
  cidr_blocks = var.cluster_endpoint_private_access_cidrs

  security_group_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
}
```
This will force everyone who want to have a private access to set explicitly their allowed subnets.

dpiddockcmp:
OK, yes that's a better solution rather than another flag. Although it does raise the risk of the good old ""count" value depends on resource attributes that cannot be determined until apply" error.

barryib:
@anthonydahanne if the proposed fix, works for you, can you please add that change to your PR. Thanks again for your time.

otodorov (approved):
I have the same issue and this fix looks good to me

anthonydahanne:
hello there! 
thanks @barryib and @dpiddockcmp for your inputs!
I'm considering the latest suggestion https://github.com/terraform-aws-modules/terraform-aws-eks/pull/981#issuecomment-681111718 and I'm wondering why the null check on `cluster_endpoint_private_access_cidrs` rather than a boolean check on `cluster_endpoint_private_access` ?
It's definitely not a big issue to me, even if I'm currently using the default `cluster_endpoint_private_access_cidrs` (not setting it) I could set it from now on (could probably make my deployment more precise)

barryib:
@anthonydahanne 

> I'm considering the latest suggestion #981 (comment) and I'm wondering why the null check on cluster_endpoint_private_access_cidrs rather than a boolean check on cluster_endpoint_private_access ?

`cluster_endpoint_private_access` is a variable to enable the [endpoint private access](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html). But it doesn't mean you want to create a SG or a SG rule. Some peoples want to manage them by themselves.

I think it's better to add another switch (`var.cluster_create_endpoint_private_access_sg_rule`) for this rule as @dpiddockcmp suggested https://github.com/terraform-aws-modules/terraform-aws-eks/pull/981#issuecomment-681665686 and change the default value of `var.cluster_endpoint_private_access_cidrs` to `null`.

```hcl
variable "cluster_create_endpoint_private_access_sg_rule" {
  description = "Whether to create security group rules for the access to the Amazon EKS private API server endpoint."
  type        = bool
  default     = false
}

variable "cluster_endpoint_private_access_cidrs" {
  description = "List of CIDR blocks which can access the Amazon EKS private API server endpoint."
  type        = list(string)
  default     = null
}

resource "aws_security_group_rule" "cluster_private_access" {
  count       = var.create_eks && var.cluster_create_endpoint_private_access_sg_rule && var.cluster_endpoint_private_access ? 1 : 0
  type        = "ingress"
  from_port   = 443
  to_port     = 443
  protocol    = "tcp"
  cidr_blocks = var.cluster_endpoint_private_access_cidrs

  security_group_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
}
```

anthonydahanne:
@barryib - to me, either way is OK.
I believe it could be a breaking change to users that the new default for `cluster_endpoint_private_access_cidrs` is `null` instead of `[0.0.0.0/0]` but after all the PR is already tagged as breaking; so that may be the time to change this default.
Updating the PR - merci !

anthonydahanne:
@barryib @dpiddockcmp @otodorov - ready for review!

barryib (changes requested):
Thanks.

Checks are failing because you don't rebuild docs. Please follow https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation to update docs with your new variables.

otodorov (approved):

otodorov:
> I believe it could be a breaking change to users that the new default for `cluster_endpoint_private_access_cidrs` is `null` instead of `[0.0.0.0/0]` but after all the PR is already tagged as breaking; so that may be the time to change this default.

Actually, every time when someone wants to upgrade something (no matter what) it is mandatory to check for breaking changes in the new version. Also, this will be mention in the Changelog.

anthonydahanne:
now it should be good! happy reviewing / merging !

anthonydahanne:
@barryib Allô! the PR seems to indicate there are changes requested from you, do you want me to update anything? Thanks

barryib (approved):

barryib:
Just merged.

Thank you @anthonydahanne for your contribution.

PR #1030, opened 2020-10-01
improvement: Make the `cpu_credits` optional for workers launch template

gillg:
# PR o'clock

## Description

If you choose non bustable families in your template (c5 for example...) and you no remove `credit_specification` section, ASG can't start any instance.
Now you can set module `cpu_credits` option to null to made it fonctionnal.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -309,13 +309,9 @@ resource "aws_launch_template" "workers_launch_template" {
       var.worker_groups_launch_template[count.index],
       "cpu_credits",
       local.workers_group_defaults["cpu_credits"]
-    ) != null ? [{}] : []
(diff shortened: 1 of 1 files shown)


barryib:
Thanks @gillg for opening this PR. I was wondering if setting the `cpu_credit` to null doesn't work without your change ?

gillg:
Hi @barryib I tried it but it's an invalid value for aws provider API.
So the solution is to remove it

barryib (changes requested):

barryib:
> Hi @barryib I tried it but it's an invalid value for aws provider API.
> So the solution is to remove it

Ok. Good to know. I just added one comment for the coding style. Just resolves that and we're good to go ^^.

barryib:
Please rebase your branch to fix the CI.

gillg:
Everything is ok !

barryib (approved):
Thanks @gillg for your contribution.

PR #1039, opened 2020-10-06
fix: Use customer managed policy instead of inline policy for `cluster_elb_sl_role_creation`

### LINKED ISSUE #1035: Avoid inline policy for cluster_elb_sl_role_creation?
## I'm submitting a...

* [X] bug report
* [X] feature request

The policy attached to the `cluster_elb_sl_role_creation` role ([here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/31ad394dbc61390dc46643b571249a2b670e9caa/cluster.tf#L161-L166) is an inline policy, which fails to apply in organizations that disable inline policies in an SCP, which is at least _somewhat_ common.

## What is the expected behavior? 

This policy should be a managed policy, like all the other policies this module creates.

Able, yes -- but wanted to get your input before doing so.

## Environment details

* Affected module version: v12.2.0
* OS: Mac OS 10.14.6
* Terraform version: 0.13.4

wlonkly:
## Description

Uses a customer-managed policy, not an inline policy, for the `cluster_elb_sl_role_creation` policy. It is common for "enterprise" AWS users to disallow inline policies with an SCP rule for auditing-related reasons, and this accomplishes the same thing. 

Verified `iam:CreatePolicy` is already listed in `docs/iam-permissions.md`. 

Fixes #1035. 

### Checklist

- [x] CI tests are passing

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -158,9 +158,15 @@ data "aws_iam_policy_document" "cluster_elb_sl_role_creation" {
   }
 }
 
-resource "aws_iam_role_policy" "cluster_elb_sl_role_creation" {
+resource "aws_iam_policy" "cluster_elb_sl_role_creation" {
   count       = var.manage_cluster_iam_resources && var.create_eks ? 1 : 0
   name_prefix = "${var.cluster_name}-elb-sl-role-creation"
-  role        = local.cluster_iam_role_name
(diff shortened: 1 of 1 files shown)


barryib (approved):

barryib:
Thanks @wlonkly for your contribution.

PR #1042, opened 2020-10-08
fix: jsonencode-instead-of-yamlencode

sortigoza:
# PR o'clock

## Description

Not a big issue but when generating a plan on mac vs linux `yamlencode` generates different output. This function is still experimental and may change in the future, for a consistent output we could use `jsonencode` until `yamlencode` is ready.

linux:

image

mac:

image

ref. https://www.terraform.io/docs/configuration/functions/yamlencode.html

Terraform v0.12.26

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -71,13 +71,13 @@ resource "kubernetes_config_map" "aws_auth" {
   }
 
   data = {
-    mapRoles = yamlencode(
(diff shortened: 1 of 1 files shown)


barryib:
Fair point. I was wondering if this introduce any breaking change ?

PR #1067, opened 2020-10-26
feat: Add EKS Fargate support

barryib:
# PR o'clock

## Description

This PR is taken from https://github.com/terraform-aws-modules/terraform-aws-eks/pull/866 and add suggested review.

### Checklist

- [x] CI tests are passing


barryib:
@itssimon I updated your PR and fixed suggested reviews. I've also add example. Can you review this please.

**Community:** We need help to test this. I would like to merge this during the week.

barryib:
@sc250024 sounds like you're also working on fargate support. Can you review this please ?

modules/fargate/variables.tf
```diff
@@ -0,0 +1,56 @@
+variable "cluster_name" {
```
antonbabenko, line 1: When the cluster is disabled (create_eks = false) all other variables should not be required. Give the default values to all of them, and add a simple module block for the disabled in the example.

modules/fargate/locals.tf
```diff
@@ -0,0 +1,10 @@
+locals {
+  create_eks              = var.create_eks && length(var.fargate_profiles) > 0
```
antonbabenko, line 2: I think it is better to introduce another variable that should be known in advance (eg, type boolean) to be able to use it in `count`. Something like `create_eks_fargate = true`.

The way it is written now `fargate_profiles` can't contain references to not-yet-created resources and modules inside of the value.

antonbabenko (commented):
Looks pretty good.

modules/fargate/fargate.tf
```diff
@@ -1,19 +1,19 @@
...
 resource "aws_iam_role_policy_attachment" "eks_fargate_pod" {
-  count      = local.create_eks && var.create_fargate_pod_execution_role ? 1 : 0
+  count      = var.create_eks && var.create_fargate_pod_execution_role ? 1 : 0
   policy_arn = "${var.iam_policy_arn_prefix}/AmazonEKSFargatePodExecutionRolePolicy"
   role       = aws_iam_role.eks_fargate_pod[0].name
 }
 
 resource "aws_eks_fargate_profile" "this" {
-  for_each               = local.create_eks ? local.fargate_profiles_expanded : {}
+  for_each               = var.create_eks ? local.fargate_profiles_expanded : {}
```
antonbabenko, line 16: This `for_each` will fail if `local.fargate_profiles_expanded` contains not-yet-known values (as you've just replaced for `local.create_eks`). I am not sure what is the best solution for this here.

barryib: `var.fargate_profiles` and `local.fargate_profiles_expanded` are maps and to work with `for_each`, their keys must be known during plan. If keys are known, I think the Terraform will wait correctly for `aws_eks_fargate_profile` creation if there is a not yet known values inside `local.fargate_profiles_expanded`.

So `length(var.fargate_profiles)` will be known because keys must be known.

modules/fargate/variables.tf
barryib: I added some default to the submodule even their are all coming from the root module. So if create_eks is false, all variable in this submodule will be set correctly.

This fargate submodule is not intended to be used alone.

barryib:
Thanks again @itssimon for all your work on this feature.

PR #1075, opened 2020-10-29
Update faq.md

daduskacpokus:
# PR o'clock

## Description

Overriding the ``null_resource.wait_for_cluster[0]`` to applying necessary steps based on AWS recommendations for windows worker-group joining to the cluster

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -134,24 +134,12 @@ Windows worker nodes requires additional cluster role (eks:kube-proxy-windows). 
 
 Amazon EKS clusters must contain one or more Linux worker nodes to run core system pods that only run on Linux, such as coredns and the VPC resource controller.
 
-1. Build AWS EKS cluster with the next workers configuration (default Linux):
+1. Build AWS EKS cluster with the next workers configuration (for local-exec provider a kubectl, openssl and jq are required):
 
 ```
-worker_groups = [
-    {
-      name                          = "worker-group-linux"
-      instance_type                 = "m5.large"
-      platform                      = "linux"
-      asg_desired_capacity          = 2
-    },
-  ]
-```
+wait_for_cluster_interpreter         = ["/bin/sh", "-c"]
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -134,24 +134,12 @@ Windows worker nodes requires additional cluster role (eks:kube-proxy-windows).
...
...            = "for i in `seq 1 60`; do if [ \"$(curl -k -s $ENDPOINT/healthz 2>/dev/null)\" = \"ok\" ]; then export KUBECONFIG=kubeconfig_${var.cluster_name} && kubectl apply -f https://amazon-eks.s3.us-west-2.amazonaws.com/manifests/${var.region}/vpc-resource-controller/latest/vpc-resource-controller.yaml && curl -o webhook-create-signed-cert.sh https://amazon-eks.s3.us-west-2.amazonaws.com/manifests/${var.region}/vpc-admission-webhook/latest/webhook-create-signed-cert.sh && curl -o webhook-patch-ca-bundle.sh https://amazon-eks.s3.us-west-2.amazonaws.com/manifests/${var.region}/vpc-admission-webhook/latest/webhook-patch-ca-bundle.sh && curl -o vpc-admission-webhook-deployment.yaml https://amazon-eks.s3.us-west-2.amazonaws.com/manifests/${var.region}/vpc-admission-webhook/latest/vpc-admission-webhook-deployment.yaml && chmod +x webhook-create-signed-cert.sh webhook-patch-ca-bundle.sh && ./webhook-create-signed-cert.sh && kubectl get secret -n kube-system vpc-admission-webhook-certs && cat ./vpc-admission-webhook-deployment.yaml | ./webhook-patch-ca-bundle.sh > vpc-admission-webhook.yaml && kubectl apply -f vpc-admission-webhook.yaml && exit 0;fi;sleep 5;done;echo TIMEOUT && exit 1"
```
barryib, line 141: It can't be so complicated right ? Is there another (simple) way to do this ?

daduskacpokus: As far as you can see by yourself there is only one [way](https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html) and this one is implemented in this commit

barryib: To me the purpose of this variable is to pull regularly the EKS API to ensure it's ready. Any kind of modification on the cluster, nodes, configmap are done through differents terraform ressources.

If you want to do this, add another null_ressource which depends on the wait_for_cluster null ressource.

barryib (changes requested):
There is a design issue here. You can't use the wait_for_cluster command to do what you want to do.

docs/faq.md
daduskacpokus: Its just an example how the user could deploy the eks-cluster with **windows worker-group** througth this module without any changes of module's code cos 9 of 10 terrraform's users won't write code by themself

barryib: If you use this module then you write are already writing a terraform module.

My point here, is that the `wait_for_cluster_cmd ` purpose is for waiting for EKS endpoint only (simple !). If you want to do more, there is probably a better place to do that.

My suggestion is to add something like this:

```hcl
resource "null_resource" "eks_windows_boostrap" {
  provisioner "local-exec" {
    # Do whatever you want here to bootstrap your windows EKS cluster
  }
  depends_on = [module.eks.cluster_id]
}
```

Furthermore, In AWS docs you link, it says 
> Amazon EKS clusters must contain one or more Linux nodes to run core system pods that only run on Linux, such as coredns and the VPC resource controller.

So you'll need a linux worker node at least for coredns and vpc ressource controller. So are you sure we don't need them ? Because you removed those parts from the FAQ.

daduskacpokus: Agree with you, but in this case code that you're wrote is wrong it should be depend on `module.eks.wait_for_cluster`

barryib: I can write mistakes. I'm just giving some guidance. As I said: "I suggest something like this or like that". Don't take my code snippet as the only tested and validated truth.

With that said, you can't use `module.eks.wait_for_cluster` because that output doesn't exist. The module  `module.eks.cluster_id` exists and depends already on wait_for_cluster null resource.

barryib (commented):
See my comments

docs/faq.md
daduskacpokus: @barryib copy that

barryib:
Closing this PR. Feel free to re-open it when you started working on it again.

PR #1076, opened 2020-10-29
improvement: Added asg tag override option

kernelold:
# PR o'clock

## Description

Added asg tag override option to improve availability to exclude worker instances from AWS backup.
So in case you need to backup only PV's you can set extra tags for all resources but EC2 instances to exclude them.

### Checklist

- ✅  CI tests are passing
- ✅  README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | permissions\_boundary | If provided, all IAM roles will be created with this permissions boundary attached. | `string` | `null` | no |
 | subnets | A list of subnets to place the EKS cluster and workers within. | `list(string)` | n/a | yes |
 | tags | A map of tags to add to all resources. | `map(string)` | `{}` | no |
+| tags\_override\_asg | A map of tags to override in autoscaling group resources. | `map(string)` | `{}` | no |
 | vpc\_id | VPC where the cluster and workers will be deployed. | `string` | n/a | yes |
 | wait\_for\_cluster\_cmd | Custom local-exec command to execute for determining if the eks cluster is healthy. Cluster endpoint will be available as an environment variable called ENDPOINT | `string` | `"for i in `seq 1 60`; do if `command -v wget > /dev/null`; then wget --no-check-certificate -O - -q $ENDPOINT/healthz >/dev/null && exit 0 || true; else curl -k -s $ENDPOINT/healthz >/dev/null && exit 0 || true;fi; sleep 5; done; echo TIMEOUT && exit 1"` | no |
 | wait\_for\_cluster\_interpreter | Custom local-exec command line interpreter for the command to determining if the eks cluster is healthy. | `list(string)` | <pre>[<br>  "/bin/sh",<br>  "-c"<br>]</pre> | no |
--- a/local.tf
+++ b/local.tf
@@ -1,5 +1,5 @@
 locals {
-  asg_tags = [
+  asg_tags = var.tags_override_asg == {} ? var.tags_override_asg : [
     for item in keys(var.tags) :
     map(
       "key", item,
--- a/variables.tf
+++ b/variables.tf
@@ -90,6 +90,12 @@ variable "tags" {
   default     = {}
 }
 
+variable "tags_override_asg" {
+  description = "A map of tags to override in autoscaling group resources."
+  type        = map(string)
+  default     = {}
+}
+
 variable "vpc_id" {
   description = "VPC where the cluster and workers will be deployed."
   type        = string


barryib:
Humm... Can't you achieve this by using [workers groups tags](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v13.0.0/local.tf#L39) ?

```hcl
module "eks" {
  # insert the 6 required variables here

  # Will add these tags to all worker groups
  workers_group_defaults = {
    tags = [
      {
        "key"                 = "foo"
        "propagate_at_launch" = "true"
        "value"               = "bar"
      },
    ]
  }

  # Overwrite tags at for a specific worker group
    {
      asg_desired_capacity = 1
      tags = [
        {
          "key"                 = "foo"
          "propagate_at_launch" = "false"
          "value"               = "fred"
        },
      ]
    }
  ]
}
```

kernelold:
> Humm... Can't you achieve this by using [workers groups tags](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v13.0.0/local.tf#L39) ?
> 
> ```terraform
> module "eks" {
>   source = "terraform-aws-modules/eks/aws"
>   # insert the 6 required variables here
> 
>   # Will add these tags to all worker groups
>   workers_group_defaults = {
>     tags = [
>       {
>         "key"                 = "foo"
>         "propagate_at_launch" = "true"
>         "value"               = "bar"
>       },
>     ]
>   }
> 
>   # Overwrite tags for a specific worker group
>   worker_groups_launch_template = [
>     {
>       name                 = "worker-group-1"
>       instance_type        = "t2.medium"
>       asg_desired_capacity = 1
>       tags = [
>         {
>           "key"                 = "foo"
>           "propagate_at_launch" = "false"
>           "value"               = "fred"
>         },
>       ]
>     }
>   ]
> }
> ```

I tried but merge is here: 
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers_launch_template.tf#L205
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers.tf#L129
 so it still will have also all the tags from `tag` var

Idea is to have some tags (a=1 b=2 c=3 backup=true) on EKS resources but in the same time have a different set of tags on EC2 nodes from ASG (a=1 b=2 c=3).

barryib:
To be honest, there is already a lot of way to define tag in this module so I'm afraid to add a new one.

But before considering adding a new one, let's try to figure it out if we can solve your problem without adding new change.

------

> so it still will have also all the tags from tag var and I didn't found any var to tag only cluster resources (PV's is my goal).

If I understood correctly your issue, you want tag your EBS volumes created by your containers with a volume claim. Right ? I don't think that EKS propagate tags to the created EBS disk or volumes. But you can do this by using the [EBS CSI Driver](https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi.html). I don't know if you're already using it (probably yes) or not, but you can define `--extra-tags` for you volumes.

> Added asg tag override option to improve availability to exclude worker instances from AWS backup.

In your backup script, can't you filter all EKS ressources except those with a specific tag ? I mean, can't you add a specific tag on your worker nodes and tell your script to not backup ressource with that particular tag ?

kernelold:
@barryib Problem is AWS Backup service is not my script 😄  and it doesn't support exclude tags(for some reason) or some other option to **not** backup any resource by specific tag. That's why `tags`  var in worker not an option.

barryib:
@kernelold Ok that invalidate my 2nd point. But does the first one still apply ?

kernelold:
@barryib I'm afraid not, we are using gp2 storage class now, so I haven't found any extra tags option for it. 😞 
I understand your concerns about extra option but tbh I don't see any alternatives for now.

barryib:
So, if you are not using the EBS CSI driver, then I think you can't propagate tags to your PVC volumes (volumes attached to your containers). Actually this module propagate tags only to EBS attached to worker nodes. FWIW, those attached volumes don't have your application data.

----
With that said, I think instead of adding new variable, we could probably improve the ASG tag processing by allowing tag overwriting at worker group level (worker groups tags suppressed `var.tags` for ASG). We can do that during by example during the `local.asg_tags` generation and skip all tags present in  `local.asg_tags` from `var.tags` we can `workers_group_defaults.tags` or worker group tags. This will allow to have tag `foo=bar` on all resources and `foo=fred` on ASG.

kernelold:
ok will try that way

huddy:
I'm not 100% sure but I just made a PR that might fix this as per @barryib suggestion: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1092

barryib:
@kernelold FWIW I merged https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1091 and released it in [v13.2.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v13.2.0). You can have a look, it'll probably help you achieve what you're trying to do.

PR #1046, opened 2020-10-12
fix: kubelet-extra-args double quote bug

pbnj:
# PR o'clock

## Description

Revert the change made in #474 which breaks --kubelet-extra-args taints and node labels (as described in #814 and #1004).

Closes #1004

### Checklist

- [x] CI tests are passing
- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

Issue #1004, opened 2020-09-09
Register nodes with multiple taints


cthiebault:
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

Using `worker_groups` I want to assign multiple taints to nodes.

It works when I assign a single taint:
```
kubelet_extra_args = "--node-labels=env-test=true --register-with-taints=env=prod:NoSchedule"
```

I tried multiple ways to assign several taints without success:
```
kubelet_extra_args = "--register-with-taints=env=prod:NoSchedule --register-with-taints=env=staging:NoSchedule" // env=staging:NoSchedule only

kubelet_extra_args = "--register-with-taints=env=prod:NoSchedule,env=staging:NoSchedule" // not even registered in EKS cluster

kubelet_extra_args = "--register-with-taints='env=prod:NoSchedule,env=staging:NoSchedule'" // no taints

kubelet_extra_args = "--register-with-taints='env=prod:NoSchedule','env=staging:NoSchedule'"  // no taints

kubelet_extra_args = "--register-with-taints 'env=prod:NoSchedule, env=staging:NoSchedule'" // no taint
```

## What's the expected behavior?

I expect to have 2 taints:
- env=prod:NoSchedule
- env=staging:NoSchedule

## Environment details

* Affected module version: 12.2.0
* Kubernetes version: 1.17
* Terraform version: 
```
Terraform v0.12.29
+ provider.aws v2.70.0
+ provider.kubernetes v1.12.0
```

anarsen:
Do the arguments make their way into the EC2 user data?

cthiebault:
It seems so:

```
# Bootstrap and join the cluster
/etc/eks/bootstrap.sh --b64-cluster-ca 'xxx' --apiserver-endpoint 'https://xxx.yl4.eu-central-1.eks.amazonaws.com'  --kubelet-extra-args "--node-labels=env-test=true --register-with-taints=env=prod:NoSchedule,env=staging:NoSchedule" 'default'
```

PG2000:
@cthiebault can you post an minimal example of your terraform where you define your worker_groups? Then i will test it

cthiebault:
Here is my code:

```
locals {
[...]

PR #1082, opened 2020-11-03
fix: Put KubeletExtraArgs in double quotes for Windows

yveslaroche:
# PR o'clock

## Description

Put KubeletExtraArgs in double quotes for Windows, it is already double quoted in the Bash script. This allows for variable substitution within the string e.g.
```
kubelet_extra_args = "node.kubernetes.io/lifecycle=$(Invoke-WebRequest -Uri http://169.254.169.254/latest/meta-data/instance-life-cycle | Select-Object -ExpandProperty Content)"
```

### Checklist

- [x] CI tests are passing

### DIFF
--- a/templates/userdata_windows.tpl
+++ b/templates/userdata_windows.tpl
@@ -4,7 +4,7 @@ ${pre_userdata}
 [string]$EKSBinDir = "$env:ProgramFiles\Amazon\EKS"
 [string]$EKSBootstrapScriptName = 'Start-EKSBootstrap.ps1'
 [string]$EKSBootstrapScriptFile = "$EKSBinDir\$EKSBootstrapScriptName"
-& $EKSBootstrapScriptFile -EKSClusterName ${cluster_name} -KubeletExtraArgs '${kubelet_extra_args}' 3>&1 4>&1 5>&1 6>&1
+& $EKSBootstrapScriptFile -EKSClusterName ${cluster_name} -KubeletExtraArgs "${kubelet_extra_args}" 3>&1 4>&1 5>&1 6>&1
 $LastError = if ($?) { 0 } else { $Error[0].Exception.HResult }
 
 ${additional_userdata}


bogdanalov-sw (approved):
LGTM

barryib:
Is this different from https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1046 ? Because the double quote in linux introduce a bug for multiple taints support.

See also https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1004

yveslaroche:
I have successfully tested applying multiple labels and taints for both Linux and Windows nodes using double quotes in the bootstrap scripts with the examples below.

```
# Linux
kubelet_extra_args = "--node-labels=node.kubernetes.io/role=worker,node.kubernetes.io/lifecycle=$(curl -s http://169.254.169.254/latest/meta-data/instance-life-cycle) --register-with-taints=somekey=somevalue:NoSchedule,anotherkey=anothervalue:NoSchedule"

# Windows
kubelet_extra_args = "--node-labels=node.kubernetes.io/role=windows,node.kubernetes.io/lifecycle=$(Invoke-WebRequest -Uri http://169.254.169.254/latest/meta-data/instance-life-cycle | Select-Object -ExpandProperty Content) --register-with-taints=os=windows:NoSchedule,anotherkey=anothervalue:NoSchedule"
```

In addition, I tested the taints applied in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1004. In this case the node won't register, but if I change the name of one of the keys to something else, it works. Hence, I believe the problem is applying two taints with the same key and effect, but different values.

barryib:
By reading this https://github.com/awslabs/amazon-eks-ami/issues/179 it sounds like we should use `''` (single quotes) to surround `kubelet_extra_args` and `""` (double quotes) to surround arguments.

Something like: `'--register-with-taints="os=windows:NoSchedule,anotherkey=anothervalue:NoSchedule"'`

I haven’t test it yet, but I'm going to dig a bit.

> In addition, I tested the taints applied in #1004. In this case the node won't register, but if I change the name of one of the keys to something else, it works. Hence, I believe the problem is applying two taints with the same key and effect, but different values.

I think you can normally provide multiple taints with the same key (by example `key1=value1:NoSchedule` and `key1=value1:NoExecute`) https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

yveslaroche:
@barryib 
> I think you can normally provide multiple taints with the same key (by example key1=value1:NoSchedule and key1=value1:NoExecute)
Yes, but that's a different effect that's set which is allowed. Setting the same key and effect, but with a different value multiple times is not allowed. If you apply it manually you'd get:
```
error: node ip-xx-xx-xx-xx.us-west-2.compute.internal already has env taint(s) with same effect(s)

barryib:
> @barryib
> 
> > I think you can normally provide multiple taints with the same key (by example key1=value1:NoSchedule and key1=value1:NoExecute)
> > Yes, but that's a different effect that's set which is allowed. Setting the same key and effect, but with a different value multiple times is not allowed. If you apply it manually you'd get:
> 
> ```
> error: node ip-xx-xx-xx-xx.us-west-2.compute.internal already has env taint(s) with same effect(s)
> ```

Sorry for the typo. I was thinking about `key1=value1:NoSchedule` and `key1=value2:NoSchedule`. I'll make some tests also.

barryib:
@yveslaroche just made some tests and I can confirm your point.

sebas-w:
Are we able to reopen this? What additional items were needed to get this merged?

barryib:
I don't use windows at all, so the point is how this different from https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1082#issuecomment-721205636 ? There is already a PR to put back single quote for linux workers https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1046.

I need people to test this.

antonbabenko:
Thanks @yveslaroche !

[v17.8.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.8.0) has been just released.

Issue #1091, opened 2020-11-06
Override ASG Tag propagation 


huddy:
# I have issues
Firstly thanks very much for maintaining this module, we've received a ton of value from this code. 

Currently when passing a list of tags into the module they get applied to both The ASG and instances/workers due to `propagate_at_launch` defaulting to true. 

I have a particular tag that I want applied to all resources including the ASG but except the worker nodes. 

A similar PR/Issue has been raised here: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1076 but was rejected with a good suggestion on how to solve this, which I plan to raise a PR for. 

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [x] kudos, thank you, warm fuzzy

## What is the current behavior?

Currently all tags passed in via `var.tags` are propagated to ASG and worker nodes. 

## What's the expected behavior?

Based on the discussion in #1076 it seems there's a desire to solve this by allowing launch templates in the variable `worker_groups_launch_template` to override `var.tags`.  This would happen for ASGs by using `local.asg_tags`, so that during processing we use `var.workers_group_defaults.tags`, thus overriding the `var.tag`.

## Are you able to fix this problem and submit a PR? Link here if you have already.

https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1092

## Environment details

* Affected module version: latest/master
* OS: N/A
* Terraform version: N/A

## Any other relevant info

It would really help although obviously not expected if this would be merged in and tagged fairly soon as this would allow us to ship some functionality we require and I know others are looking forward to this as well!

barryib:
Closing this, since #1092 is now merged.

barryib:
[v13.2.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v13.2.0) is now released.

PR #1092, opened 2020-11-06
feat: Tags passed into worker groups override tags from `var.tags` for Autoscaling groups

huddy:
# PR o'clock

## Description

This PR brings an enhancement that allows tags that are passed into `var.worker_groups_launch_template` or `var.worker_groups` to override tags passed in via `var.tags` for ASGs only. 

This relates to #1091 

The main use case for this is where a tag is required everywhere apart from worker nodes. In this case we can pass the tag in via `var.tags`, but then pass it in with the worker template and set `propagate_at_launch` to `false`. 

In order to ensure that the override only happens where it needs to I had to move the asg_tags for expression from `locals.tf` to be in the context of each worker template.

This way we don't incorrectly override tags where we haven't actually overwritten them. 

I wasn't sure if you would class this as an improvement or feature, but I went with improvement, happy to change it to a "feature".

### Checklist

- [x] CI tests are passing


workers.tf
```diff
@@ -126,7 +126,15 @@ resource "aws_autoscaling_group" "workers" {
...
         },
       ],
-      local.asg_tags,
+      [
+        for item in keys(var.tags) :
+        map(
+          "key", item,
+          "value", element(values(var.tags), index(keys(var.tags), item)),
+          "propagate_at_launch", "true"
+        )
```
barryib, line 135: It's probably because you take it from the `local.asg_tags`, but can't we simplify this writing ?

```hcl
for k,v in var.tags :
map(
  "key", k,
  "value", v,
  "propagate_at_launch", "true"
)
```
Same point in worker_groups_launch_template ASG.

huddy: Hey @barryib good suggestion. I had thought about that but wanted to keep my first PR fairly as simple as possible. I've made these changes. 

Billy

huddy:
I've just merged latest master too

barryib (approved):

barryib:
Thanks @huddy for your contribution. I'll make a new release as soon as [fargate PR](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1067) is merged.

huddy:
Thanks for the very quick turnaround @barryib and once again thanks to you and other maintainers for awesome work.

PR #1095, opened 2020-11-10
improvement: Tags passed into worker groups should also be excluded from Launch Template tag specification

huddy:
# PR o'clock

## Description

I recently raised #1092 which allows us to overide tags on the asg and set `propagate_at_launch` to `false` to prevent instances from getting the tag. 

This works well for launch configuration, but for launch templates we also currently pass in `tag_specification` for instances which then negates the point of #1092 

This PR therefore adds a feature to exclude any tags on the tag specification for instances on the launch templates, that are passed in via worker groups config. 

This means we can control if a tag makes it to instance from the ASG or not without the tag specification interfering and the ASG tags become the source of truth for propagation only if they are passed in with worker groups. 

### Checklist

- [x] CI tests are passing


workers_launch_template.tf
```diff
@@ -456,7 +456,10 @@ resource "aws_launch_template" "workers_launch_template" {
           count.index,
         )}-eks_asg"
       },
-      var.tags,
+      { for tag_key, tag_value in var.tags :
+        tag_key => tag_value
+        if ! contains([for tag in lookup(var.worker_groups_launch_template[count.index], "tags", local.workers_group_defaults["tags"]) : tag["key"]], tag_key)
```
barryib, line 461: Can you please exclude the tag `Name` also, like you already did for ASG ?

barryib (changes requested):

workers_launch_template.tf
huddy: Thanks for the review, Yep, I've added this now :)

barryib (approved):
Thanks @huddy.

barryib:
[v13.2.1](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v13.2.1) is now released

PR #1096, opened 2020-11-10
fix: Don’t add empty Roles ARN in aws-auth configmap, specifically when no Fargate profiles are specified

huddy:
# PR o'clock

## Description
The new fargate code seems to have caused a bug in authmap.

Currently even when no fargate profiles are specified the fargate module output is used to create an authmap entry with an empty ARN. 

The plan shows:
```
                - "groups":
              +   - "system:bootstrappers"
              +   - "system:nodes"
              +   - "system:node-proxier"
              +   "rolearn": ""
              +   "username": "system:node:{{SessionName}}"
              + - "groups":
```
Apply errors:

```
Error: Failed to update Config Map: Unauthorized
  on .terraform/modules/eks_cluster/aws_auth.tf line 65, in resource "kubernetes_config_map" "aws_auth":
  65: resource "kubernetes_config_map" "aws_auth" {
```

This seems to be because "rolearn" is empty.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -59,7 +59,6 @@ locals {
         role["platform"] == "fargate" ? ["system:node-proxier"] : [],
       ))
     }
-    if role["worker_role_arn"] != ""
   ]
 }
 
--- a/aws_auth.tf
+++ b/aws_auth.tf
@@ -59,6 +59,7 @@ locals {
         role["platform"] == "fargate" ? ["system:node-proxier"] : [],
       ))
     }
+    if role["worker_role_arn"] != ""
   ]
 }
 
--- a/modules/fargate/outputs.tf
+++ b/modules/fargate/outputs.tf
@@ -24,6 +24,6 @@ output "aws_auth_roles" {
     for i in range(1) : {
       worker_role_arn = local.pod_execution_role_arn
       platform        = "fargate"
-    } if length(var.fargate_profiles) > 0
+    } if local.create_eks
   ]
 }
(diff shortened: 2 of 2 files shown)


aws_auth.tf
```diff
@@ -59,6 +59,7 @@ locals {
         role["platform"] == "fargate" ? ["system:node-proxier"] : [],
       ))
     }
+    if role["worker_role_arn"] != ""
```
barryib, line 62: I think we should instead return an empty list for `module.fargate.aws_auth_roles` https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/fargate/outputs.tf#L21-L27.

Return [] if `local.create_eks`. Something like this by example:

```hcl
output "aws_auth_roles" {
  description = "Roles for use in aws-auth ConfigMap"
  value = [for i in range(1) : {
    worker_role_arn = local.pod_execution_role_arn
    platform        = "fargate"
    } if local.create_eks
  ]
}
```

huddy: Yeah this might more elegant :) I've made this change with one exception and that's using `if length(var.fargate_profiles) > 0` instead of `local.create_eks` since it seems that `local.create_eks` would always be true and `var.fargate_profiles` seems to be control point for that submodule.

huddy: Yes this seems more elegant actually, so I'll go with this! I'm just testing the change now.

huddy: All working as I would expect!

barryib: great.

barryib (approved):

barryib:
Thanks @huddy for your fix. Nice catch.

barryib:
[v13.2.1](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v13.2.1) is now released

PR #1116, opened 2020-11-23
feat: Added input for worker nodes ssh key

dtherhtun:
# PR o'clock

## Description

Just add an input variable for worker node ssh `key_name`.  
for example usage: `key_name        = module.key_pair.this_key_pair_key_name`
### Checklist

- [x] CI tests are passing


barryib:
I don't get it. You can already define a key_name for all your worker groups (`var.workers_group_defaults.key_name`) or define one per worker groups (`var.worker_groups.key_name`). Why do you need an extra variable ?

dtherhtun:
Because I didn't know these 2  variables `workers_group_defaults.key_name`, `worker_groups.key_name` before you commented and I can't find input variables for ssh key on this https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest?tab=inputs

dtherhtun:
I added an additional `key_name` variable with a default null value and it is nothing change on `local.key_name` but I will display on modules input tags and I will help some people like me.

barryib:
Closing this since, it's already implemented.

PR #1119, opened 2020-11-24
feat: Ignore changes to SG name_prefix and description to support migration paths to module

jaygorrell:
# PR o'clock

## Description

When migrating existing resources to this module with `terraform state mv`, TF will want to re-create the Security Group (and thus, the cluster) due to the `name_prefix` and `description` field changing unless they happened to match. These are safe enough fields to ignore changes on because they will always force re-creation anyway, which is never wanted.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -88,6 +88,9 @@ resource "aws_security_group" "cluster" {
       "Name" = "${var.cluster_name}-eks_cluster_sg"
     },
   )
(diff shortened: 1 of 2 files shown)


barryib:
I don't think we should ignore name or name_prefix change. Maybe we can add the ability to specify a SG name instead ?

Issue #1105, opened 2020-11-18
Feat/bug: workers ASGs that are more resilient to unnecessary change


mariadb-JeffBachtel:
# I have issues

## I'm submitting a...

* [X] bug report
* [X] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [X] kudos, thank you, warm fuzzy

## What is the current behavior?

If I mutate the array I pass in for `worker_groups` at all, all of my ASGs are regenerated and it's kinda explosive

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Pass in a list for `worker_groups`, append an entry on a subsequent run, note the resources to destroy, try not to cry, cry anyway.

## What's the expected behavior?

By preference, only ASGs that didn't exist before would be added, or if existing ASGs were changed would they be replaced.

## Are you able to fix this problem and submit a PR? Link here if you have already.

Not really but I had an idea that instead of a count loop (https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers.tf#L4 ) we could use a for_each loop, maybe. A map would be better for this than a list, so perhaps a different ASG fed by a new module param (`worker_groups_map`?) would be needed for backwards compatibility. Or maybe you could generate a map from a list I'm no scientist.

## Environment details

* Affected module version: `12.2.0` but the code looks the same in `13.2.1`
* OS: Linux
* Terraform version: `0.13.4`

## Any other relevant info

Thanks for this module, I've been using it I think actual years plural now, and it's always super useful and composable (in fact, for this ticket I could easily create my own `workers.tf` and have it work with your existing module easy peasy)

pre:
This issue of ”resource needs to be recreated” seems to be a deeper one. 

It is also a pain with managed node groups. You can’t modify a launch template without recreating the whole node group, which is just plain wrong: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1109

mariadb-JeffBachtel:
I put in a PR that internally creates a map based on the list of workers from the caller (so no API change to module) and uses `for_each`. That approach could probably be used elsewhere resources are created based on a `count`

mariadb-JeffBachtel:
unstale?

james-callahan:
I would love to see this issue fixed.

Is this module still being maintained? (there hasn't been a commit since January)

james-callahan:
.

[...]

Issue #1101, opened 2020-11-14
How to handle state migrations as code during module upgrades ?


barryib:
# I have issues

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## Discussion

I'm opening this issue to start discussion about migrations and state manipulations during upgrades, for https://github.com/terraform-aws-modules/terraform-aws-eks/pull/858 and for the next migrations ?

Today there is no an easy way to upgrade. As this module is getting more complex, we took the decision to split the module into small submodules. We discussed it in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635 and https://github.com/terraform-aws-modules/terraform-aws-eks/issues/774.

We now support Fargate and Managed Node Groups as submodules. The next step will be :
- Move worker groups as submodules and drop Launch Configuration support https://github.com/terraform-aws-modules/terraform-aws-eks/pull/858
- Move cluster and aws-auth into submodules

Those moves will add a lot of breaking changes and will break all clusters and workloads, so your production environement.

To avoid that, we're actually looking for a way to ease that migration.

One of the solution I've in mind is to use https://github.com/minamijoyo/tfmigrate and create a submodule to generate a `tfmigrate.hcl` file to help users in their migrations.

The workflow should be:
1. `terraform apply` before upgrade to ensure that your code is in sync with your states
2. backup your states
3. Run the `terraform apply` with `var.eks_tfmigrate=true` to generate the `tfmigrate.hcl`
4. Run `tfmigrate plan`
5. Run `tfmigrate apply`
6. Upgrade your module version and run  `terraform plan` to see if there are some change

- What do you think about this workflow ?
- Do you have another idea ?
- Should we provide this module only for https://github.com/terraform-aws-modules/terraform-aws-eks/pull/858 or should we continue to support if ? If yes, for what kind of change ?

cc @terraform-aws-modules/triage-supporters @grzegorzlisowski @sc250024 @js-timbirkett

## Additional infos
- https://github.com/minamijoyo/tfmigrate
- https://github.com/hashicorp/terraform/issues/19354

daroga0002:
In general I like your plan. Questions/comments are:

[...]

Issue #774, opened 2020-03-10
Newly refactored module 


js-timbirkett:
Hello 👋 

Firstly, this module is great and has a lot to deal with as standing up an EKS cluster and nodes / workers is not a trivial task.

There have been issues regarding orchestrating the creation of the control plane and the workers / nodes [#519], issues questioning complexity or coupling within the module [#635], questions with regards to LC / LT worker_groups and whether they should be merged or simplified [#563], and issues around numeric indexes and modifying arrays in general.

I had a go at creating a new, refactored version of this module and my attempt is here:  
https://github.com/devopsmakers/terraform-aws-eks

It breaks things out into sub-modules that can be used by both the parent module (to maintain the simple single module gets you a working cluster) and directly by the user of the module for more complex situations (Custom CNI).

It also allows a map of maps for worker groups and the code has been simplified a bit (reduction of the death by lookups in the current worker related code).

I made a few tough decisions that might stimulate some debate and possibly some outrage. These are highlighted in the repos README.

I'm not saying it's perfect (yet), but I think it would make a good base for a fairly major refactor of this module and I'd be glad to assist in any way I can. Feel free to use any ideas or the codebase itself to further improve this module.

Thanks!

barryib:
Hello @js-timbirkett thanks for opening this discussion and for your work. It's a good start.

Can you please open a PR so we can iterate and give feedbacks ?

For the design decisions:
- To me we can split worker groups in separate submodule like we already did for managed node groups.
- As for dropping LC support, I don't have a strong opinion on this. I personally don't use LC at all, but I don't know if there are lot of people out there which still ely on this. See https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635#issuecomment-571076487

@max-rocket-internet @dpiddockcmp @antonbabenko please advices (mostly for the LC removal)

antonbabenko:
+1 for dropping LC. I think we all agreed in some previous issues that dropping support for LC is the way to go.

I wonder, are there any downsides of using LT at all comparing to LC?

max-rocket-internet:
> I wonder, are there any downsides of using LT at all comparing to LC?

No downsides to LT that I know of, LC is just older and has less features.

max-rocket-internet:
> I had a go at creating a new, refactored version of this module

Sounds great but we need a PRs to review and a path forward for current users.

I look forward to seeing some PRs 😃

js-timbirkett:
[...]

PR #1132, opened 2020-12-02
feat: Use a Map internally for workers data structure

mariadb-JeffBachtel:
# PR o'clock

## Description

This changeset morphs the List structure passed in for workers into a Map, so that it churns less with changes (additions/removals/etc)

This addresses https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1105 somewhat (not for all worker types)

### Checklist

- [?] CI tests are passing
- [👎 ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation - (no changes to variables or outputs intended)


pre:
I'm in favor of not using an Array ever with Terraform resources, but .. 

Wouldn't this be a breaking change - big time ?

Maybe a softer approach for changing an array structure (dangerous terraform) to a map (good terraform), would be to create a separate configuration key altogether. Ie leave the current `worker_groups` array definition as is and introduce a better one with eg `autoscaling_groups` (as worker group in this project seems to mean an asg). 

Some keys already are implemented using the map format, like `node_groups`. 

(Array is bad because you can't remove an element in the middle of the list without recreating all the resources in that list. Map is good because you can remove elements independently.)

mariadb-JeffBachtel:
It's not a breaking change because the array -> map change is done entirely internally, so the input/output interface to the module remains exactly the same. I'm seeing unexpected churn due to launch configuration changes, but I can't verify why.

mariadb-JeffBachtel:
I see I have a lint failure, but I'm not able to repeat it locally. What version of terraform is being run for that step?

mmingorance-dh:
I'd love to see this PR merged. Is there any ETA when this could be released in a new module version?

mariadb-JeffBachtel:
@antonbabenko would you mind approving this for workflow?

barryib:
@mariadb-JeffBachtel thanks for working on this. But since we wanted to split this module into submodule, we want to split this and move from `count` to `for_each` in the same time. Probably by using for_each to loop over sub-module.

So this work, is good, but it doesn't go into that direction. I'll update following issues to share a design direction and start a new branch as preview branch.

- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/774
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1101

mariadb-JeffBachtel:
I see, thanks Thierno. I don't want to litter the module's PR list, so I'll close for now. Cheers :)

PR #1129, opened 2020-12-02
feat: support for capacity_type and list of instance_types in node_groups

jonathan-mothership:
# PR o'clock

## Description
* Adds support for specifying the node group `capacity_type` to leverage the new AWS provider feature for spot instances in managed node groups. 
* Allows for specifying a list of instance types in the `node_group` map, defaulting to a list with the instance type in `workers_group_defaults`.

These [features](https://github.com/hashicorp/terraform-provider-aws/pull/16510) are available in the [latest v3 release of the AWS provider](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v3.19.0), so the `required_providers` version for `aws` was pushed to `3.19.0`.

With these changes, I was able to switch my existing node groups to using spot with a custom launch configuration so we can pass our `user_data`! 

### Checklist

- [x] CI tests are passing
- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #997, opened 2020-08-30
feat: LaunchTemplate support for managed node-groups

philicious:
# PR o'clock

## Description

fixes #979

Just recently on 17th August, AWS released LaunchTemplate support for managed node-groups. https://aws.amazon.com/blogs/containers/introducing-launch-template-and-custom-ami-support-in-amazon-eks-managed-node-groups/
Furthermore the `aws` provider also supports it since `3.3.0` https://github.com/terraform-providers/terraform-provider-aws/pull/14639

This module didn't support it yet, only LTs for self-managed worker groups.

As the module is quite complex already, I only added support for providing the id of LT you create yourself and then supply the Id.
The existing `workers_launch_template.tf` couldn't have been easily reused imho as its related to also creating ASGs and other resources for self-managed node-groups. Also at least the `iam_instance_profile` should NOT be supplied for LTs being used for managed node-groups as I noticed. AWS API will error then.
So instead of adding another LT manifest and variables and wiring that all together, I prefered to take care of the LT myself and copied the userdata template 

I now create a LT like
```
data "template_file" "launch_template_userdata" {
  template = file("${path.module}/templates/userdata.sh.tpl")

  vars = {
    cluster_name        = var.cluster_name
    endpoint            = module.eks.cluster_endpoint
    cluster_auth_base64 = module.eks.cluster_certificate_authority_data

    bootstrap_extra_args = ""
    kubelet_extra_args = ""
    }
}

// this is mostly the default LT that AWS would create if you dont specify your own
resource "aws_launch_template" "default" {
  name_prefix     = "${var.cluster_name}-"
  description     = "Default Launch-Template for clusters"
  update_default_version = true

  block_device_mappings {
    device_name = "/dev/xvda"

    ebs {
      volume_size           = 100
      volume_type           = "gp2"
      delete_on_termination = true
    }
  }

  ebs_optimized = true // some instance types dont support it, so check when changing type

  image_id      = "ami-00341e507eb458a09" //TODO use our custom AMI
  instance_type = var.instance_type

  monitoring {
    enabled = true
  }

  network_interfaces {
    associate_public_ip_address = false
    delete_on_termination       = true
    security_groups             = [module.eks.worker_security_group_id]
  }

  user_data = base64encode(
    data.template_file.launch_template_userdata.rendered,
  )

  lifecycle {
    create_before_destroy = true
  }  
}
```

and pass it to the eks module
```
module "eks" {
  source          = "../../../../../terraform-aws-eks/"
[...]

PR #1138, opened 2020-12-07
feat: Create launch template for Managed Node Groups

ArchiFleKs:
Signed-off-by: Kevin Lefevre <lefevre.kevin@gmail.com>

# PR o'clock

## Description

Enable the creation of a default launch template if needed to use with managed node pool. This enable the use of `kubelet_extra_args` and to add taint quickly without having to manage a separate launch template Terraform config.

It implements this logic: https://github.com/aws/containers-roadmap/issues/864

I think the launchtemplate default might need some trimming, tell me what you think.

### Checklist

- [X] CI tests are passing


cabrinha:
Did you try creating a cluster using this TF code? I'm getting the following error when using one or more node_groups:

```

  on ../../../modules/terraform-aws-eks/modules/node_groups/launchtemplate.tf line 2, in data "template_file" "workers_userdata":
   2:   for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] }

The "for_each" value depends on resource attributes that cannot be determined
resources that the for_each depends on.
```

modules/node_groups/launchtemplate.tf
```diff
@@ -0,0 +1,94 @@
+data "template_file" "workers_userdata" {
+  for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] }
+  template = file("${path.module}/templates/userdata.sh.tpl")
+
+  vars = {
+    bootstrap_extra_args = each.value["bootstrap_extra_args"]
```
cabrinha, line 6: This variable is not used in the template. I had to remove it in order for a plan to succeed.

ArchiFleKs:
@cabrinha thanks for the review, actually I used Terragrunt + Terraform but I was able to get a cluster running, I havent tested without disk_size (I tested with 50 also) so I think we need to set a default because there is none with launch template.

ArchiFleKs:
@cabrinha also the plan is passing with the `examples/managed_node_groups` and:

```
  node_groups = {
    example = {
      desired_capacity       = 1
      max_capacity           = 10
      min_capacity           = 1
      create_launch_template = true
      kubelet_extra_args     = "--node-labels=role=nginx,group=nginx"

      instance_type = "m5.large"
      k8s_labels = {
        Environment = "test"
        GithubRepo  = "terraform-aws-eks"
        GithubOrg   = "terraform-aws-modules"
      }
      additional_tags = {
        ExtraTag = "example"
      }
    }
  }
```

modules/node_groups/launchtemplate.tf
ArchiFleKs: I remove it

cabrinha:
> @cabrinha thanks for the review, actually I used Terragrunt + Terraform but I was able to get a cluster running, I havent tested without disk_size (I tested with 50 also) so I think we need to set a default because there is none with launch template.
> 
> If we set the `for_each` to `for_each = local.node_groups_expanded` it will create a launch template for every node group

Could we add a simple `create = true/false` on each group? I guess thats what the `create_launch_template` flag would do.

> About the `name` I don think managed node group is setting the name tag on instances as even my "classic" managed node pool don't have one.

I wish there was a way to add the Name tag to the instance spun up.

Also, it'd be nice to add [capacity_type](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#capacity_type) too, since AWS now supports Managed Node Groups with Spot Instances.

modules/node_groups/node_groups.tf
```diff
@@ -14,8 +14,8 @@ resource "aws_eks_node_group" "workers" {
   }
 
   ami_type        = lookup(each.value, "ami_type", null)
-  disk_size       = lookup(each.value, "disk_size", null)
-  instance_types  = each.value["launch_template_id"] != null ? [] : [each.value["instance_type"]]
+  disk_size       = each.value["launch_template_id"] != null || each.value["create_launch_template"] ? null : lookup(each.value, "disk_size", null)
+  instance_types  = each.value["launch_template_id"] != null || each.value["create_launch_template"] ? [] : [each.value["instance_type"]]
```
cabrinha, line 18: It'd be cool to add `capacity_type   = lookup(each.value, "capacity_type", "ON_DEMAND")` right above this, so users could choose to use `"SPOT"`

ArchiFleKs:
@cabrinha Alright for the spot and the name. 

But I have to admit I really do not understand the part about about the `create = true/false` because that is what I'm trying to with ` for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] }` which is same syntax as [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/node_groups/locals.tf#L17)

I'm using the latest Terraform 0.13.5, I have ont tried with 0.12 I will test on my end.

mjwilkerson-strateos (approved):

cabrinha:
> @cabrinha Alright for the spot and the name.
> 
> But I have to admit I really do not understand the part about about the `create = true/false` because that is what I'm trying to with ` for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] }` which is same syntax as [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/node_groups/locals.tf#L17)
> 
> I'm using the latest Terraform 0.13.5, I have ont tried with 0.12 I will test on my end.

Supposedly this issue is more likely to happen on a blank tfstate.

modules/node_groups/node_groups.tf
ArchiFleKs: If think this is implemented by https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1129

cabrinha:
Seems these three PRs are all related: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1161 https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1129

binnythomas-1989:
Guys I have a question related to remote access, you won't be able  to mention remote access on node groups as per aws docs

```
Per our documentation[1] When using a launch template, if any of the following parameters are specified in the node group configuration, your create or update request will fail. Specify these in your launch template:

- Instance type
- Disk size
- Remote access configuration
- EC2 SSH key
```

I dont see a handle of that on your code launch_template.tf. would be great if you can clarify.

ArchiFleKs:
> Guys I have a question related to remote access, you won't be able to mention remote access on node groups as per aws docs
> 
> ```
> Per our documentation[1] When using a launch template, if any of the following parameters are specified in the node group configuration, your create or update request will fail. Specify these in your launch template:
> 
> - Instance type
> - Disk size
> - Remote access configuration
> - EC2 SSH key
> ```
> 
> I dont see a handle of that on your code launch_template.tf. would be great if you can clarify.

You mean to perform validation on the input ?

binnythomas-1989:
> > Guys I have a question related to remote access, you won't be able to mention remote access on node groups as per aws docs
> > ```
> > Per our documentation[1] When using a launch template, if any of the following parameters are specified in the node group configuration, your create or update request will fail. Specify these in your launch template:
> > 
> > - Instance type
> > - Disk size
> > - Remote access configuration
> > - EC2 SSH key
> > ```
> > 
> > 
> > I dont see a handle of that on your code launch_template.tf. would be great if you can clarify.
> 
> You mean to perform validation on the input ?

What I mean is if you add a remote access as an input with a launch template, you would end up with the below error.

```
Error: error creating EKS Node Group (nodegroup-test:x86_64-driving-louse): InvalidParameterException: Remote access configuration cannot be specified with a launch template.
{
```

ArchiFleKs:
@binnythomas-1989 I think I understand, we need to prevent setting remote_access here : https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/node_groups/node_groups.tf#L21 if launchtemplate is used

binnythomas-1989:
you are correct @ArchiFleKs

binnythomas-1989:
Guys adding the Kubelet argument is great. I just tested it. Im kinda sorry with amazon their EKS is kinda really annoying.

So let me explain what I have figured.
You can add node-taints  something like below.

```
kubelet_extra_args = "--node-labels=eks.amazonaws.com/nodegroup=company-net --register-with-taints=network=company:NoSchedule"

```

The problem is once you add the taints with a NoSchedule for example, the node won't join the cluster. You would have  issues with the coredns pod running since its a deployment. It won't be able to schedule.

So letting people to use this for taints is a bad idea on EKS.

ArchiFleKs:
@binnythomas-1989 it is true for coreDNS, aws-node tolerate every taint, you should always have a "default"  pool or a "criticalAddonsOnly" pool if you want.

I agree this is kind of a poweruser feature, node join also failed if you are using `kubernetes.io` forbidden label. But letting user the ability to have "reserved" node pool, for GPU etc is a must have feature in my opinion while it is not natively supported by the EKS API whereas label are.

ArchiFleKs:
There is an example right now on how to use a launch template but it is not straightforward, this PR goal is to enable a simple configuration while still allowing to use a custom launch template if needed or a basic node pool.

`eksctl` is using a default launch template also to enable taint on node groups

binnythomas-1989:
> @binnythomas-1989 it is true for coreDNS, aws-node tolerate every taint, you should always have a "default" pool or a "criticalAddonsOnly" pool if you want.
> 
> I agree this is kind of a poweruser feature, node join also failed if you are using `kubernetes.io` forbidden label. But letting user the ability to have "reserved" node pool, for GPU etc is a must have feature in my opinion while it is not natively supported by the EKS API whereas label are.


The kube-proxy and aws-node is a DaemonSet so that's okay, its just the core-dns is an issue since its a deployment. Anyways it could be useful if we add a condition on the README. i was trying to adopt your solution on my local terraform module I use. Im planning to handle this using terraform using kubectl provider for now. :-) I dont use eksctl

ArchiFleKs:
@binnythomas-1989 this : https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R22 should take care of remote access, could you try ? If not I'll try to test today

binnythomas-1989:
> @binnythomas-1989 this : https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R22 should take care of remote access, could you try ? If not I'll try to test today

This won't work again, Because now you need to handle the Remote access on Launch template  and it would go like this

```
  key_name = each.value.ec2_ssh_key_pair
```

So that bit is being handled on the launch_template.tf

ArchiFleKs:
> > @binnythomas-1989 this : https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R22 should take care of remote access, could you try ? If not I'll try to test today
> 
> This won't work again, Because now you need to handle the Remote access on Launch template and it would go like this
> 
> ```
>   key_name = each.value.ec2_ssh_key_pair
> ```
> 
> So that bit is being handled on the launch_template.tf

I added the key_name lookup

bobbywatson3:
This is exactly what we're looking for. Thank you for the work done on this so far!

barryib (commented):
I just realized that I didn't submit my review. It was in pending state.

stevehipwell:
@ArchiFleKs do you have a rough timeline for this work?

Also, would it be possible to add support for `pre_userdata` and `additional_userdata`? We need this for custom certificates, docker credentials and custom labels from AWS metadata.

modules/node_groups/launchtemplate.tf
```diff
@@ -0,0 +1,99 @@
...
+    enabled = true
+  }
+
+  network_interfaces {
+    associate_public_ip_address = false
+    delete_on_termination       = true
+  }
+
+  # if you want to use a custom AMI
+  # image_id      = var.ami_id
```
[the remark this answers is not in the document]
ArchiFleKs: I think because if using a custom image, this does not apply: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/launch_templates_with_managed_node_groups/launchtemplate.tf#L18 I think this lead to different behavior between EKS ami and other AMI, but I have not tested with custom AMI

ArchiFleKs:
@stevehipwell working on @barryib comments :)

modules/node_groups/launchtemplate.tf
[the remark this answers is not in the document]
ArchiFleKs: I think we should stay simple, because we still allow user to use a custom launch template if needed, or maybe this can be added in another PR as I'm not really sure on how to handle cloud-init with custom AMI

ArchiFleKs:
@stevehipwell I'm not sure it will work with `pre_userdata` and `additional_userdata` because eks launch template merge cloud init when using default AMI, so I'm not sure how to enforce that `pre_userdata`  is executed before the EKS cluster is joined and additional_userdata after, like specified here : https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/templates/userdata.sh.tpl

stevehipwell:
@ArchiFleKs the requirements we have is that we can run a script before the node is added, without this there is no way to use secondary subnets on a node group that spans subnets where there is more than one subnet per AZ (the best practice pattern). We also need this to install a custom root cert for the docker daemon.

ArchiFleKs:
I think it is doable because EKS merge happens last, the user data injected in this PR takes place before the EKS default one.

Do you mind explaining this to me :  without this there is no way to use secondary subnets on a node group that spans subnets where there is more than one subnet per AZ (the best practice pattern). ?

ArchiFleKs:
@stevehipwell this is what happen when using default AMI:

```
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="//"

--//
Content-Transfer-Encoding: 7bit
Content-Type: text/x-shellscript
Mime-Version: 1.0

#!/bin/bash -e

# Allow user supplied pre userdata code


sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" "' /etc/eks/bootstrap.sh

# Allow user supplied userdata code


--//
Content-Type: text/x-shellscript; charset="us-ascii"
#!/bin/bash
set -ex
B64_CLUSTER_CA=LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUN5RENDQWJDZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFWTVJNd0VRWURWUVFERXdwcmRXSmwKY201bGRHVnpNQjRYRFRJeE1ESXdNakUyTXpJeU0xb1hEVE14TURFek1URTJNekl5TTFvd0ZURVRNQkVHQTFVRQpBeE1LYTNWaVpYSnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBTkhXCk5vZjgxekorcGIxdEswMXRWVExSNEd0NDBDbkw5TU5vV0hSWGc3WndNVFkzcHVQMm05TlkvSXJ2bEZ2dDNNUVcKejUrb0FRdU8rcHA2RUFQOEZFK0JGaUVSVXpMZTYvbXFscGg2S2hmOEsyQU45QUN2RUYvMWlYNlQvWFlDdlRrRQp5MmhYSk1CUnVGSVF6dGVSaDEwRTFBZG5UWDdxNUY5RlhIY2VzR285TGlPbmRNMVpQRGpPS2lnZ0hMK2xheG4wCnN0bDlxeGZrYWZpMHNzb0ZCcUM3eGU1SGt2OVowYTYvRmxWeVNXazFQQXFCWDZOTlUvc0RjNTA3bXN0OEVMc0oKSU9naWFTcGZLaXVnekZNaTlTS3NQbjRQcm94UDEwRlErOGpSdTZZdm9tQmswMHFnU2NFTGxadng0bG1CVGloSgpCdDdFTlUxMzdvSXdhY1pCUUNFQ0F3RUFBYU1qTUNFd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0RRWUpLb1pJaHZjTkFRRUxCUUFEZ2dFQkFJam1ZWmthV3NuQ1lSNkpQUGw1WmVpcGkzYkYKREpBUzgvM2E4UFVnL3BsWTFVYlhCalU3b0FDb21UUzd2Y2hPUFU5aFNXdC9jNit5RnF5a0FwakMyRjFuSHg4WQpaQUg5NDFWYUNzRyt3VmE3MTJlcFRPTSt1TWxNSENFYVlMVTRKOXEvaUd1aVZtM2NPOGhmMTFoNjVGd3NuekE0CmdqQ0YxUC9Sdi9acnFSSk9XZmJaRE00MzlwajVqQzNYRVAyK1FXVlIzR2tzbW1NcDVISm9NZW5JaDBSTFhnK1oKTVRVNXFsdW0xTWZDdXRNVjkzNGJFQ21BRERJSm4rZVdHSERwRi9QOThnR1RyRU1QclhiUXZMblpwZHBNYldjNQp5LzZldkNtYXozMzllSlUwWkRaM1M0R2YvbEpBUTBZcFZoQkRlS2hXVHEwSXJYb2NWWHU5MDN0OXU5TT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
API_SERVER_URL=https://A62655F81AE9347A761BB172E28A633F.sk1.eu-west-1.eks.amazonaws.com
K8S_CLUSTER_DNS_IP=172.20.0.10
/etc/eks/bootstrap.sh pio-thanos --kubelet-extra-args '--node-labels=eks.amazonaws.com/sourceLaunchTemplateVersion=1,GithubRepo=terraform-aws-eks,eks.amazonaws.com/nodegroup-image=ami-066fad1ae541d1cf9,Environment=test,eks.amazonaws.com/capacityType=ON_DEMAND,eks.amazonaws.com/nodegroup=pio-thanos-default-eu-west-1a-expert-bass,eks.amazonaws.com/sourceLaunchTemplateId=lt-079cbc5cf74ace131,GithubOrg=terraform-aws-modules' --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $API_SERVER_URL --dns-cluster-ip $K8S_CLUSTER_DNS_IP


--//--

```

ArchiFleKs:
@stevehipwell  I removed the additional user data as the are basicly the same as `pre` in this case. I just pushed my latest work and rebased against master 

@barryib I implemented the changes you requested

stevehipwell:
@ArchiFleKs it would be useful to have a `pre_userdata` argument for a managed node group if possible, is that what you've done above?

RE secondary subnets; if you have a network with public and private subnets there are no node annotations to differentiate between these if you are using a secondary subnet per primary subnet. Without an annotation it's you can't map the VPC config to a node. Our solution is to add a node label via the `kubelet_extra_args` that uses a variable we declare in the `pre_userdata`, userdata code below.

```shell
bootstrap_token="$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")"
bootstrap_mac="$(curl -s -H "X-aws-ec2-metadata-token: ${bootstrap_token}" "http://169.254.169.254/latest/meta-data/mac")"
bootstrap_subnet="$(curl -s -H "X-aws-ec2-metadata-token: ${bootstrap_token}" "http://169.254.169.254/latest/meta-data/network/interfaces/macs/${bootstrap_mac}/subnet-id")"
```

ArchiFleKs:
@stevehipwell you should be able to do that with the `pre_userdata` as implemented above

stevehipwell:
Thanks @ArchiFleKs, that looks exactly like what we need.

modules/node_groups/launchtemplate.tf
```diff
@@ -0,0 +1,99 @@
+data "template_file" "workers_userdata" {
```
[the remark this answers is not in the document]
ArchiFleKs: done

modules/node_groups/launchtemplate.tf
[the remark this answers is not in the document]
ArchiFleKs: This is not implemented for now

modules/node_groups/launchtemplate.tf
```diff
@@ -0,0 +1,102 @@
...
+  instance_type = each.value.instance_type
+
+  monitoring {
+    enabled = lookup(each.value, "enable_monitoring", null)
+  }
+
+  network_interfaces {
+    associate_public_ip_address = lookup(each.value, "public_ip", null)
+    delete_on_termination       = lookup(each.value, "eni_delete", null)
+  }
```
philicious, line 50: I think you miss `security_groups             = [module.eks.worker_security_group_id]` here

philicious, line 50: I think you miss security_groups = [module.eks.worker_security_group_id] here

modules/node_groups/launchtemplate.tf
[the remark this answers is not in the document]
philicious: as you partially copied the LT from the examples that got added in my #997 , I might be able to help here:

so I am using LT **with** a custom AMI and it works just fine. 
however there is indeed subtle but important differences between using an LT w/ or w/o a custom AMI. In the old PR, someone else described it quite well, see https://github.com/terraform-aws-modules/terraform-aws-eks/pull/997#issuecomment-705286151

He also mentions a then required fix to the MIME boundary when using cloudinit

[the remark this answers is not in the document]
ArchiFleKs: @philicious do you mean that this part should be added manually when using custom AMI:

```
set -ex
B64_CLUSTER_CA=LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUN5RENDQWJDZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFWTVJNd0VRWURWUVFERXdwcmRXSmwKY201bGRHVnpNQjRYRFRJeE1ESXdNakUyTXpJeU0xb1hEVE14TURFek1URTJNekl5TTFvd0ZURVRNQkVHQTFVRQpBeE1LYTNWaVpYSnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBTkhXCk5vZjgxekorcGIxdEswMXRWVExSNEd0NDBDbkw5TU5vV0hSWGc3WndNVFkzcHVQMm05TlkvSXJ2bEZ2dDNNUVcKejUrb0FRdU8rcHA2RUFQOEZFK0JGaUVSVXpMZTYvbXFscGg2S2hmOEsyQU45QUN2RUYvMWlYNlQvWFlDdlRrRQp5MmhYSk1CUnVGSVF6dGVSaDEwRTFBZG5UWDdxNUY5RlhIY2VzR285TGlPbmRNMVpQRGpPS2lnZ0hMK2xheG4wCnN0bDlxeGZrYWZpMHNzb0ZCcUM3eGU1SGt2OVowYTYvRmxWeVNXazFQQXFCWDZOTlUvc0RjNTA3bXN0OEVMc0oKSU9naWFTcGZLaXVnekZNaTlTS3NQbjRQcm94UDEwRlErOGpSdTZZdm9tQmswMHFnU2NFTGxadng0bG1CVGloSgpCdDdFTlUxMzdvSXdhY1pCUUNFQ0F3RUFBYU1qTUNFd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0RRWUpLb1pJaHZjTkFRRUxCUUFEZ2dFQkFJam1ZWmthV3NuQ1lSNkpQUGw1WmVpcGkzYkYKREpBUzgvM2E4UFVnL3BsWTFVYlhCalU3b0FDb21UUzd2Y2hPUFU5aFNXdC9jNit5RnF5a0FwakMyRjFuSHg4WQpaQUg5NDFWYUNzRyt3VmE3MTJlcFRPTSt1TWxNSENFYVlMVTRKOXEvaUd1aVZtM2NPOGhmMTFoNjVGd3NuekE0CmdqQ0YxUC9Sdi9acnFSSk9XZmJaRE00MzlwajVqQzNYRVAyK1FXVlIzR2tzbW1NcDVISm9NZW5JaDBSTFhnK1oKTVRVNXFsdW0xTWZDdXRNVjkzNGJFQ21BRERJSm4rZVdHSERwRi9QOThnR1RyRU1QclhiUXZMblpwZHBNYldjNQp5LzZldkNtYXozMzllSlUwWkRaM1M0R2YvbEpBUTBZcFZoQkRlS2hXVHEwSXJYb2NWWHU5MDN0OXU5TT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
API_SERVER_URL=https://A62655F81AE9347A761BB172E28A633F.sk1.eu-west-1.eks.amazonaws.com
K8S_CLUSTER_DNS_IP=172.20.0.10
/etc/eks/bootstrap.sh pio-thanos --kubelet-extra-args '--node-labels=eks.amazonaws.com/sourceLaunchTemplateVersion=1,GithubRepo=terraform-aws-eks,eks.amazonaws.com/nodegroup-image=ami-066fad1ae541d1cf9,Environment=test,eks.amazonaws.com/capacityType=ON_DEMAND,eks.amazonaws.com/nodegroup=pio-thanos-default-eu-west-1a-expert-bass,eks.amazonaws.com/sourceLaunchTemplateId=lt-079cbc5cf74ace131,GithubOrg=terraform-aws-modules' --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $API_SERVER_URL --dns-cluster-ip $K8S_CLUSTER_DNS_IP
```

About the boudaries i think this is fixed and I implemented it [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-88d020257d5b3a657e10667cc97a9f3b0fa430b32bcd72d894d262d9c5351f3aR5)

I'm just not sure how to handle the custom AMI, how do you do it on your end ?  From what I understand you need to pass the bootstrap command on your own when using custom AMI, because it does not get merge

ArchiFleKs:
@philicious I implemented the same security group logic as worker nodes, actually, the more we add to this PR the more I think we should make using a default launch template the "default" because we can have almost the same feature partity as worker_group_launch_template but with the managed rolling update

modules/node_groups/launchtemplate.tf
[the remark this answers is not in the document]
philicious: you could be right about the boundaries. I remember that someone wanted to update the cloudinit to support custom boundaries. so ye, probably and if this PR produces a running EKS, then its proven.

and ye, with a custom AMI, you have to supply the entire userdata yourself as no merging happens with the default one from EKS.
so I in the end use it like I added in the examples. 

I would have to have a closer look and do tests on how to add custom AMI support and satisfy these differences.

it would for sure be great if the module could also handle custom AMI if it already got LT generation added.

philicious:
> @philicious I implemented the same security group logic as worker nodes

ok great if it works w/o ! I only noticed you didnt add the SG like in my example and I for sure had to add it to get it working in #997 , but you might have done things a bit different that obsoletes adding the SG 🤷‍♂️

stevehipwell:
@barryib do you know when this might be able to be merged?

ck3mp3r:
These missing features are exactly what is stopping us from using node-groups at the moment.
@barryib is there a way to expedite the merge?
Kudos to @ArchiFleKs !

philicious:
@ck3mp3r just to note, this is already possible since my #997 with supplying LT manually. See examples

modules/node_groups/launchtemplate.tf
```diff
@@ -0,0 +1,111 @@
...
+    device_name = "/dev/xvda"
+
+    ebs {
+      volume_size           = lookup(each.value, "disk_size", null)
+      volume_type           = lookup(each.value, "disk_type", null)
+      delete_on_termination = true
+    }
+  }
+
+  instance_type = each.value.instance_type
```
tayritenour, line 41: When I attempt to build with a simple `node_groups` such as: 
```
node_groups = {
    general_pods = {
      instance_types = ["m4.large", "t3a.large"]
      desired_capacity = 1
      min_capacity = 1
      max_capacity = 5
      capacity_type = "SPOT"
      create_launch_template = true
      kubelet_extra_args = "--node-labels=lifecycle=Ec2Spot --register-with-taints=Ec2Spot=true:NoSchedule",

      tags = {
        "k8s.io/cluster-autoscaler/node-template/label/lifecycle" = "Ec2Spot"
        "k8s.io/cluster-autoscaler/node-template/taint/Ec2Spot" = "true:NoSchedule"
        "k8s.io/cluster-autoscaler/node-template/resources/cpu" = "2"
        "k8s.io/cluster-autoscaler/node-template/resources/memory": "8Gi"
      }
    }
}
```
It fails by not being able to get the instance_types:
```
Error: Unsupported attribute

  on ../../../../../terraform-aws-eks/modules/node_groups/launchtemplate.tf line 41, in resource "aws_launch_template" "workers":
  41:   instance_type = each.value.instance_type
    |----------------
    | each.value is object with 20 attributes

This object does not have an attribute named "instance_type".
```
Is there something I'm misunderstanding about how this should be initialized?

ck3mp3r:
@philicious yup, I've seen that in the examples and docs. I'd still prefer something similar to the way worker nodes are currently implemented. I think the way @ArchiFleKs implemented it is a great feature to have.

philicious:
@ck3mp3r yes it is. Only tried to unblock you in case you didn't know. 

So, I for myself vaguely plan on extending this feature in another PR with ability to provide encryption KMS key and custom AMI. Then it's fully-fledged

stevehipwell:
The sooner this can get merged the better, even if it's only MVP.

modules/node_groups/launchtemplate.tf
tayritenour: The issue is that it appears as though we cannot use the `instance_types` attribute as defined. Only when I specifically say `instance_type` and use only one will it work.

ArchiFleKs: Not sure I have updated to 14.0 so that might be the issue

ArchiFleKs: Should be fixed with latest commit

ArchiFleKs:
Any update on this ?

ck3mp3r:
Yes, would be great to understand what the strategy is regarding proper support of node-groups...

riccardofreixo:
This would be massively useful to us, even if just at an MVP state.

stevehipwell:
Does anyone know when this is likely to be progressed? We're being blocked waiting for it as we don't want to reinvent the wheel.

burnertoday:
@philicious , @cabrinha , @tayritenour , @mjwilkerson-strateos , @binnythomas-1989 do u guys have a plan to review this PR. There's a lot of us waiting for it?

surajnarwade:
I've got a couple of questions on this PR:

* how do you pass the list of different instance types? for example [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R18):  this only takes one value technically

* also how this works when `launch_template` has `instance_type` [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-88d020257d5b3a657e10667cc97a9f3b0fa430b32bcd72d894d262d9c5351f3aR41) & `aws_eks_node_group` also has `instance_types` [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R18) ? AFAIK terraform will throw an error if you provide instance_type at the both the places

ArchiFleKs:
@surajnarwade from what I understand launch template only supports `instance_type` so I take the first element of the `instance_types` inputs [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-88d020257d5b3a657e10667cc97a9f3b0fa430b32bcd72d894d262d9c5351f3aR41). I'm not sure how to use the mixed instance type with spot as it is at the ASG level and not LT.

surajnarwade:
@ArchiFleKs yeah that's right but then you are also passing instance_types in node_groups [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R18) how does that work ?

ArchiFleKs:
@surajnarwade not sure what line I'm supposed to see in your link. The instance_types is passed to NG [here](https://github.com/ArchiFleKs/terraform-aws-eks/blob/feat/ng-extra-args/modules/node_groups/node_groups.tf#L18) only if no launch template configuration is defined (either the launch template ID which was supported before or with the `create_launch_template` for the default launch template. If either one is used the node group instance_types will be empty

surajnarwade:
@ArchiFleKs, thanks for the clarification :)  does that mean we can't specify `instance_types` in NG if the launch template is used?

my use case is moreover I want to use launch template for NG but my NG config looks like this,

```
  node_groups = {
    example = {
      desired_capacity = 1
      max_capacity     = 10
      min_capacity     = 1

      instance_types = ["m5.large", "t2.medium"] 
      capacity_type  = "SPOT"
      k8s_labels = {
        Environment = "test"
        GithubRepo  = "terraform-aws-eks"
        GithubOrg   = "terraform-aws-modules"
      }
      additional_tags = {
        ExtraTag = "example"
      }
    }
  }
```

will that work ?

ArchiFleKs:
@surajnarwade from what I gather [here](https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html)

> If you deploy a node group using a launch template, then you can specify zero or one Instance type under Launch template contents in a launch template or you can specify 0-20 instance types for Instance types on the Set compute and scaling configuration page in the console, or using other tools that use the Amazon EKS API. If you specify an instance type in a launch template, and use that launch template to deploy your node group, then you can't specify any instance types in the console or using other tools that use the Amazon EKS API. If you don't specify an instance type in a launch template, in the console, or using other tools that use the Amazon EKS API, then the t3.medium instance type is used, by default. If your node group is using the Spot capacity type, then we recommend specifying multiple instance types using the console. For more information, see Managed node group capacity types.

 I think maybe we should pass no instance type at all in the launchtemplate so that we can let EKS manage the intance type

surajnarwade:
@ArchiFleKs agree we should take this into consideration and set  `instance_type` to null if instance_types is provided or something like that, wdyt ?

ArchiFleKs:
> @ArchiFleKs agree we should take this into consideration and set `instance_type` to null if instance_types is provided or something like that, wdyt ?

I think we should always pass `null` and letting EKS do the work with `instances_types`

cabrinha:
> @philicious , @cabrinha , @tayritenour , @mjwilkerson-strateos , @binnythomas-1989 do u guys have a plan to review this PR. There's a lot of us waiting for it?

I'm not a reviewer nor owner. I'd love to see this merged too :D

philicious:
> > @ArchiFleKs agree we should take this into consideration and set `instance_type` to null if instance_types is provided or something like that, wdyt ?
> 
> I think we should always pass `null` and letting EKS do the work with `instances_types` it is supposed to work with only one instances also. I can update the PR if you can test it that would be great, I cannot really test at the moment

I would say:
- set `instance_types` on the MNG in case no LT is supplied (at the moment it only takes the first element `[each.value["instance_type"]]`)
- introduce a flag that lets you choose where to set the instance_type in case of using an LT: either in LT or MNG
- use first element of `instance_types` in the LT as it only supports one. set it to null if the flag is set

wdyt @ArchiFleKs @surajnarwade

cabrinha:
> > > @ArchiFleKs agree we should take this into consideration and set `instance_type` to null if instance_types is provided or something like that, wdyt ?
> > 
> > 
> > I think we should always pass `null` and letting EKS do the work with `instances_types` it is supposed to work with only one instances also. I can update the PR if you can test it that would be great, I cannot really test at the moment
> 
> I would say:
> 
> * set `instance_types` on the MNG in case no LT is supplied (at the moment it only takes the first element `[each.value["instance_type"]]`)
> * introduce a flag that lets you choose where to set the instance_type in case of using an LT: either in LT or MNG
> * use first element of `instance_types` in the LT as it only supports one. set it to null if the flag is set
> 
> wdyt @ArchiFleKs @surajnarwade

Let's go.

ArchiFleKs:
Sounds good to me. I'll update the PR this weekend

philicious:
Nice 👍🏼 
I can help testing then if required

ArchiFleKs:
@philicious @cabrinha : I have tested with the following config:

```
  node_groups = {
    "default-${local.aws_region}a" = {
      desired_capacity = 1
      max_capacity     = 3
      min_capacity     = 1
      instance_types   = ["t3a.medium","t3.medium"]
      create_launch_template = true
      subnets          = [dependency.vpc.outputs.private_subnets[0]]
      kubelet_extra_args = "--node-labels=role=monitoring --register-with-taints=dedicated=monitoring:NoSchedule"
      disk_size        = 50
    }

    "default-${local.aws_region}b" = {
      desired_capacity = 1
      max_capacity     = 3
      min_capacity     = 1
      instance_types    = ["t3a.medium", "m5.large"]
      create_launch_template = true
      set_instance_types_on_lt = true
      subnets          = [dependency.vpc.outputs.private_subnets[1]]
      disk_size        = 50
    }

    "default-${local.aws_region}c" = {
      desired_capacity = 1
      max_capacity     = 3
      min_capacity     = 1
      instance_type    = "t3a.medium"
      subnets          = [dependency.vpc.outputs.private_subnets[2]]
      instance_types   = ["t3a.medium","t3.medium"]
      disk_size        = 50
      kubelet_extra_args = "--node-labels=role=spot --register-with-taints=dedicated=spot:NoSchedule"
      create_launch_template = true
      capacity_type = "SPOT"
    }
  }
```

cabrinha:
> @philicious @cabrinha : I have tested with the following config:

Tested and confirmed working for me. Thanks! 😎

philicious:
awesome @ArchiFleKs ! to me sounds as there is no more testing required from my side then !

I am also looking forward to a new release ! 

I would also base on your work and extend it with customAMI and root volume encryption features for the LT (just as I use it already with supplying an external LT)

philicious:
fixes #1211  
closes #1221

cabrinha:
@barryib can we get a review?

sbarbat:
Couldn't we add the `bootstrap_extra_args` parameter as well? 🙏

modules/node_groups/templates/userdata.sh.tpl
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+
+# Allow user supplied pre userdata code
+${pre_userdata}
+
+sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" ${kubelet_extra_args}"' /etc/eks/bootstrap.sh
```
philicious, line 6: btw @ArchiFleKs I am trying to understand why your bootstrap command is missing the cluster-name argument which should be a problem according to https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L266-L269

see for reference https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/templates/userdata.sh.tpl#L7 or https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/launch_templates_with_managed_node_groups/templates/userdata.sh.tpl#L10

and also misses ${bootstrap_extra_args} that @sbarbat mentioned, which would be a good addition so we can supply any of the other optional options

ArchiFleKs: The issue is with the auto merge of official AWS usersata and the custom supplied. So there is no possibility to add bootstrap args before.

ArchiFleKs:
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138#issuecomment-771736034

philicious:
from the docs:
>Note that user data is used by EKS to insert the EKS bootstrap invocation for your managed nodes. EKS will automatically merge this in for you, unless a custom AMI is specified. In that case, you’ll need to add that in. More on this point below.

so that would mean that we dont have to / cannot supply the bootstrap command at all when not using a custom AMI. i.e. I would expect your line of
>sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" ${kubelet_extra_args}"' /etc/eks/bootstrap.sh
 to have no effect at all as the real bootstrap arg is merged in by EKS (and yours missing the mandatory cluster-name)

ArchiFleKs:
I'm just modifying with sed the existing file before the bootstrap command before it is called by EKS AMI. I'm not calling the bootstrap command. 

If we specific the bootstrap command manually with NO custom AMI, it will break. And if we do not specify it with custom AMI it will break also from what I understand. But I have not tried. It is working as it is with official AMI. You can check also the discussion here https://github.com/aws/containers-roadmap/issues/864

I guess we could set a custom user data in case we supply a AMI which is non default. 

Or maybe have a look at how they do it in eksctl

philicious:
@ArchiFleKs sorry yes you are right, you are modifying file via sed.
I was confused as you can also supply these args as env when calling the file, which looks similar. all good 👍 

in a follow-up PR, I will add custom-AMI + userdata/bootstrap handling which then also supports `bootstrap_extra_args` to specify any like 

lets not take another turn here and get it merged as it is

cabrinha:
@barryib can we get a review?

LarsBroeders:
Could we please get some progress on this? We've held back on doing work because of this PR since this seemed so close to being merged.

cabrinha:
@barryib can we get some more reviewers / approvers on this repo?

cabrinha:
@bryantbiggs @antonbabenko @brandonjbjelland can we get some more eyes on this PR please?

antonbabenko:
Unfortunately, I can't help with reviewing this PR since I am now entirely into the subject and this module.

Let's hear what @barryib decides when he has time for this.

cabrinha:
> Unfortunately, I can't help with reviewing this PR since I am now entirely into the subject and this module.
> 
> Let's hear what @barryib decides when he has time for this.

@barryib hasn't been active on this PR since January.

antonbabenko:
@cabrinha We just need to wait and see when he has time for this PR. We all have other things to do except open-source. 

Let's raise this issue in a week or so if @barryib does not reply.

barryib:
> @cabrinha We just need to wait and see when he has time for this PR. We all have other things to do except open-source. 
> 
> Let's raise this issue in a week or so if @barryib does not reply.

I'll merge this for a release during the week. After that I think I'll focus on code refactoring on a `next` branch. During that period, I'll probably stop introducing new features untill the refactoring is done. I'll write more about design on a specific thread.

cabrinha:
@ArchiFleKs looks like we got some conflicts here. Can you fix these up please?

ArchiFleKs:
> > @cabrinha We just need to wait and see when he has time for this PR. We all have other things to do except open-source.
> > Let's raise this issue in a week or so if @barryib does not reply.
> 
> I'll merge this for a release during the week. After that I think I'll focus on code refactoring on a `next` branch. During that period, I'll probably stop introducing new features untill the refactoring is done. I'll write more about design on a specific thread.

Is there any way to assist on this refactoring ? I assume splitting worker type in modules is part of it ? as well as switching to for_each loop

ArchiFleKs:
> @ArchiFleKs looks like we got some conflicts here. Can you fix these up please?

@cabrinha I"m not sure I'll have the time to retest after the merge this week end but I'll try when I can. It would be great if you could or someone else

cabrinha:
> > @ArchiFleKs looks like we got some conflicts here. Can you fix these up please?
> 
> @cabrinha I"m not sure I'll have the time to retest after the merge this week end but I'll try when I can. It would be great if you could or someone else

Sure, I can retest this at any time and post my configs here.

ArchiFleKs:
Alright conflict should be fixed now

cabrinha:
> > @ArchiFleKs looks like we got some conflicts here. Can you fix these up please?
> 
> @cabrinha I"m not sure I'll have the time to retest after the merge this week end but I'll try when I can. It would be great if you could or someone else

Just tested this config:
```terraform
  node_groups = {
    managed = {
      desired_capacity = 1
      max_capacity     = 5
      min_capacity     = 1

      instance_types = [
        "c3.2xlarge",
        "c4.xlarge",
        "c4.2xlarge",
      ]
      capacity_type  = "SPOT"
      root_volume_type = "gp2"
      root_volume_size = 10
      kubelet_extra_args = "--node-labels=node.kubernetes.io/lifecycle=spot,role=worker,node.kubernetes.io/exclude-from-external-load-balancers --register-with-taints=dedicated=managed:NoSchedule"
      k8s_labels = {
        Environment = "test"
        GithubRepo  = "terraform-aws-eks"
        GithubOrg   = "terraform-aws-modules"
      }

      additional_tags = {
        CustomTag = "EKS example"
      }
    }
  }
```

Seems to be working well

devy294:
When will this be merged?

cabrinha:
> When will this be merged?

Great question.

barryib (approved):

stevehipwell:
@barryib it looks like the changes in this PR didn't make it into v15.0.0 or v15.1.0, do you have a plan for when they are going to be released?

cabrinha:
> @barryib it looks like the changes in this PR didn't make it into v15.0.0 or v15.1.0, do you have a plan for when they are going to be released?

Aren't the changes here? https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v15.1.0...master

https://github.com/terraform-aws-modules/terraform-aws-eks/commit/2e1651df86bd315000738cf901a4cc0586be1af3

archoversight:
> > @barryib it looks like the changes in this PR didn't make it into v15.0.0 or v15.1.0, do you have a plan for when they are going to be released?
> 
> Aren't the changes here? [v15.1.0...master](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v15.1.0...master) [2e1651d](https://github.com/terraform-aws-modules/terraform-aws-eks/commit/2e1651df86bd315000738cf901a4cc0586be1af3)

That first link is showing the diff between v15.1.0 and master. Your second link shows that it is only on `master` and not yet in any tags.

martin308:
I'm not seeing this work with `kubelet_extra_args`. When this is used the userdata that is provided to the launch template only includes the additions provided to `kubelet_extra_args` and none of the other required parameters to bootstrap.

Example:

```
retriever = {
       instance_types    = ["c5.xlarge"]
       desired_capacity  = 1
       max_capacity      = 1
       min_capacity      = 1
       disk_size         = 20
       kubelet_extra_args = "--node-labels=test=extra_args"
       create_launch_template = true
      }
```

Produces the following userdata for the launch configuration:

```
Content-Type: multipart/mixed; boundary="//"
MIME-Version: 1.0

--//
Content-Transfer-Encoding: 7bit
Content-Type: text/x-shellscript
Mime-Version: 1.0

#!/bin/bash -e

# Allow user supplied pre userdata code


sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" --node-labels=test=extra_args"' /etc/eks/bootstrap.sh

--//--
```

Which is missing all of the other required parameters provided by the EKS AMI which is vital for registering the nodes in the cluster.

Am I missing something or does this not work?

stevehipwell:
@martin308 this hasn't been released yet, so unless you're using `ref=master` it won't do anything.

jfoechsler:
> ```
> 
> Which is missing all of the other required parameters provided by the EKS AMI which is vital for registering the nodes in the cluster.
> 
> Am I missing something or does this not work?

@martin308 I'm pretty sure what you are missing/forgetting, is this: https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#launch-template-user-data about using custom user data snippets while still using official AMI. 

You would need full user data in case of custom AMI, but in that case you would also have created the full regular launch template.

martin308:
> @martin308 this hasn't been released yet, so unless you're using `ref=master` it won't do anything.

yup, pulling down the merge commit by ref 👍 

`source           = "github.com/terraform-aws-modules/terraform-aws-eks?ref=2e1651df86bd315000738cf901a4cc0586be1af3"`

> > ```
> > 
> > Which is missing all of the other required parameters provided by the EKS AMI which is vital for registering the nodes in the cluster.
> > 
> > Am I missing something or does this not work?
> > ```
> 
> 
> @martin308 I'm pretty sure what you are missing/forgetting, is this: https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#launch-template-user-data about using custom user data snippets while still using official AMI.
> 
> 
> You would need full user data in case of custom AMI, but in that case you would also have created the full regular launch template.

I'm not using a custom AMI as per my example. I guess I'm just confused as how to use the `kubelet_extra_args` feature added in this PR. 

Are you saying that it is expected that my example above would not work? If so is there an example of how to make use of the `kubelet_extra_args` feature with the official AMI?

jfoechsler:
> Are you saying that it is expected that my example above would not work? If so is there an example of how to make use of the `kubelet_extra_args` feature with the official AMI?

No I'm saying the opposite :) My understanding is your example with the resulting LT should work and be able to join cluster (due to the merging of user data done outside control of this terraform module). I'm interested to hear if your testing confirm that.

ArchiFleKs:
I just tested with `master` and the following configuration:

```
  node_groups = {
    "default-${local.aws_region}a" = {
      ami_type = "AL2_ARM_64"
      desired_capacity = 1
      max_capacity     = 3
      min_capacity     = 1
      instance_types    = ["t4g.large"]
      subnets          = [dependency.vpc.outputs.private_subnets[0]]
      disk_size        = 20
    }

    "default-${local.aws_region}b" = {
      ami_type = "AL2_ARM_64"
      desired_capacity = 1
      max_capacity     = 3
      min_capacity     = 1
      instance_types    = ["t4g.large"]
      subnets          = [dependency.vpc.outputs.private_subnets[1]]
      disk_size        = 20
    }

    "default-${local.aws_region}c" = {
      ami_type = "AL2_ARM_64"
      create_launch_template        = true
      desired_capacity = 1
      max_capacity     = 3
      min_capacity     = 1
      instance_types    = ["t4g.large"]
      subnets          = [dependency.vpc.outputs.private_subnets[2]]
      kubelet_extra_args = "--node-labels=role=private --register-with-taints=dedicated=private:NoSchedule"
      disk_size        = 20
    }
  }
```

It is working as expected

ArchiFleKs:
You have to use the `create_launch_template` flag which is not by default or else the `kubelet_extra_args`  are not passed to anything.

stevehipwell:
@ArchiFleKs I've been looking at further customizing the manage node group bootstrap process and I'd be interested if you tried setting the `KUBELET_EXTRA_ARGS` environment variable instead of using `sed`?

ipleten:
> @ArchiFleKs I've been looking at further customizing the manage node group bootstrap process and I'd be interested if you tried setting the `KUBELET_EXTRA_ARGS` environment variable instead of using `sed`?

I tried and seems cloud-init  doesn't preserve exported variables between its parts.

stevehipwell:
@ipleten it was a leading question, I've done this exact thing for some of the other env variables by persisting the export. I'll probably open a PR to change this as it's more resilient to AMI changes than the sed solution.

stevehipwell:
@ipleten it looks like I already did, https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1433.

Issue #1056, opened 2020-10-18
supports configurable Kubernetes service IP address range


yanivpaz:
## Feature request
support configurable Kubernetes service IP address range

## What is the current behavior?
[configurable Kubernetes service IP address range](https://aws.amazon.com/about-aws/whats-new/2020/10/amazon-eks-supports-configurable-kubernetes-service-ip-address-range/)  is not supported 

## What is the expected behavior?
Support this feature ( for reference  CFN - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-cluster-kubernetesnetworkconfig.html ) 

## Are you able to fix this problem and submit a PR? Link here if you have already.
No

barryib:
This feature is not available yet in the aws provider, but there is an ongoing issue/PR https://github.com/terraform-providers/terraform-provider-aws/issues/15505.

etoews:
For visibility, the PR [Enhancement/eks adding service ipv4 range](https://github.com/hashicorp/terraform-provider-aws/pull/15518) has merge and been released in [v3.19.0](https://github.com/hashicorp/terraform-provider-aws/blob/v3.19.0/CHANGELOG.md) of the provider.

PR #1139, opened 2020-12-08
feat: Add support for `service_ipv4_cidr` for the EKS cluster

nasuku:
# PR o'clock

## Description

This change adds support for service_ipv4_cidr that has been added to terraform-provider-aws last week.
( please see https://github.com/hashicorp/terraform-provider-aws/issues/15505 )

This allows the user to specify the ipv4 CIDR for the k8s services (resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1056 )

### Checklist

- [x] CI tests are passing

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -22,6 +22,10 @@ resource "aws_eks_cluster" "this" {
     public_access_cidrs     = var.cluster_endpoint_public_access_cidrs
   }
 
+  kubernetes_network_config {
+    service_ipv4_cidr = var.cluster_service_ipv4_cidr
+  }
+
   timeouts {
     create = var.cluster_create_timeout
     delete = var.cluster_delete_timeout
--- a/variables.tf
+++ b/variables.tf
@@ -369,3 +369,9 @@ variable "fargate_pod_execution_role_name" {
   type        = string
   default     = null
 }
+
+variable "cluster_service_ipv4_cidr" {
+  description = "service ipv4 cidr for the kubernetes cluster"
+  type        = string
+  default     = null
+}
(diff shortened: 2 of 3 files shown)


nasuku:
The linter failures (which are happening on top of tree) are fixed in a separate pr #1140 .

goodspellar:
Is there a rough time estimate of when this might get merged? Looks like a rather benign change, but it's blocking our prod EKS deployment :)

snvnandan:
@nasuku we forked you repo to use it. With the new Service IPv4 CIDR, we ran into `--dns-cluster-ip` issue. We were getting 172.20.0.10 as the cluster DNS on the pods, which is not the subnet we are using. This also needs to be looked into on [bootstrap.sh](https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh) in amazon-eks-ami. 

As a workaround we specified this `--dns-cluster-ip <ip>` in `bootstap_extra_args` referenced in [local.tf](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/local.tf)

nasuku:
@im-nonedone  thanks for mentioning it. Yes. I too was passsing the dns-cluster-ip in bootstrap_extra_args.
I felt that this was more like a bug/enhancement in bootstrap.sh than handling in terraform.
if no fix comes from aws, then may be, a separate PR which accepts dns_cluster_ip as another variable would be a better option.

cluster.tf
```diff
@@ -22,6 +22,10 @@ resource "aws_eks_cluster" "this" {
     public_access_cidrs     = var.cluster_endpoint_public_access_cidrs
   }
 
+  kubernetes_network_config {
+    service_ipv4_cidr = var.service_ipv4_cidr != "" ? var.service_ipv4_cidr : null
```
barryib, line 26: wouldn't be simpler if we  define `var.service_ipv4_cidr` default to `null` and then just write:

```hcl
kubernetes_network_config {
    service_ipv4_cidr = var.service_ipv4_cidr
}
```

variables.tf
```diff
@@ -367,3 +367,9 @@ variable "fargate_pod_execution_role_name" {
   type        = string
   default     = null
 }
+
+variable "service_ipv4_cidr" {
```
barryib, line 371: Can you please rename this into `cluster_service_ipv4_cidr` for namming consistency.

variables.tf
```diff
@@ -367,3 +367,9 @@ variable "fargate_pod_execution_role_name" {
...
+  description = "service ipv4 cidr for the kubernetes cluster"
+  type        = string
+  default     = ""
```
barryib, line 374: default to `null` as suggested upper ?

nasuku:
thanks for the review @barryib . Made changes per your suggestion and rebased to  master branch.

barryib (approved):

Issue #1107, opened 2020-11-18
[Feature Request] Spot for Managed Node Groups


psoares:
# I have issues

The documentation shows an example of how to create a spot fleet using a workergroup launch template.

Is it possible to do the same as a nodeGroup instead ?

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
The documentation shows how to create a spot fleet using launch templates.
```hcl
worker_groups_launch_template = [
    {
      name                    = "spot-1"
      override_instance_types = ["m5.large", "m5a.large", "m5d.large", "m5ad.large"]
      spot_instance_pools     = 4
      asg_max_size            = 5
      asg_desired_capacity    = 5
      kubelet_extra_args      = "--node-labels=node.kubernetes.io/lifecycle=spot"
      public_ip               = true
    },
  ]
```

## What's the expected behavior?
I'd like to have a nodegroup that manages a spot fleet instead, so it appears in the EKS console or when I do `aws eks list-nodegroups`

## Environment details

* Affected module version: 13.1.0
* OS: Linux
* Terraform version:  0.13.5

barryib:
This is not yet available in EKS. AWS is working on it. From https://github.com/aws/containers-roadmap/issues/583, it sounds like the feature is coming soon.

lifelofranco:
Updating that spot for managed node groups has just been released! Would be cool if this gets supported.

Here's the blog post: https://aws.amazon.com/blogs/containers/amazon-eks-now-supports-provisioning-and-managing-ec2-spot-instances-in-managed-node-groups/

psoares:
I've tweaked the module locally and think I could propose a change.
It would involve the following:
- Add a "capacity_type" attribute to the nodegroup map. It can take 2 values "ON_DEMAND" (default) or "SPOT" (see https://aws.amazon.com/blogs/containers/amazon-eks-now-supports-provisioning-and-managing-ec2-spot-instances-in-managed-node-groups/ and https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#capacity_type)
[...]

PR #1161, opened 2020-12-24
feat: Add support for spot instances node groups.

psoares:
# PR o'clock

## Description

This is to implement feature request #1107.
It allows creating nodegroups with spot instances, as described in https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-eks-support-ec2-spot-instances-managed-node-groups/

example usage:
```
module "eks" {
  // [...]
  node_groups = {
    spot-inst-managed = {
      name             = "spot-managed-${var.cluster_name}"
      desired_capacity = 2
      max_capacity     = 20
      min_capacity     = 2
      instance_type    = ["m5.large", "m5a.large", "m5d.large", "m5ad.large"]
      capacity_type    = "SPOT"
    }
```

### Checklist

- [x] CI tests are passing

### DIFF
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -15,7 +15,7 @@ resource "aws_eks_node_group" "workers" {
 
   ami_type        = lookup(each.value, "ami_type", null)
   disk_size       = lookup(each.value, "disk_size", null)
-  instance_types  = try(each.value["launch_template_id"] != "", false) ? [] : flatten([each.value["instance_type"]])
+  instance_types  = each.value["launch_template_id"] != "" ? [] : flatten([each.value["instance_type"]])
   release_version = lookup(each.value, "ami_release_version", null)
   capacity_type   = lookup(each.value, "capacity_type", "ON_DEMAND")
 
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -15,7 +15,7 @@ resource "aws_eks_node_group" "workers" {
 
   ami_type        = lookup(each.value, "ami_type", null)
   disk_size       = lookup(each.value, "disk_size", null)
-  instance_types  = each.value["launch_template_id"] != null ? [] : each.value["instance_types"] != null ? each.value["instance_types"] : [each.value["instance_type"]]
+  instance_types  = try(each.value["launch_template_id"] != "", false) ? [] : flatten([each.value["instance_type"]])
   release_version = lookup(each.value, "ami_release_version", null)
   capacity_type   = lookup(each.value, "capacity_type", "ON_DEMAND")
 
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -15,8 +15,9 @@ resource "aws_eks_node_group" "workers" {
 
   ami_type        = lookup(each.value, "ami_type", null)
   disk_size       = lookup(each.value, "disk_size", null)
-  instance_types  = each.value["launch_template_id"] != null ? [] : [each.value["instance_type"]]
+  instance_types  = each.value["launch_template_id"] != null ? [] : each.value["instance_types"] != null ? each.value["instance_types"] : [each.value["instance_type"]]
   release_version = lookup(each.value, "ami_release_version", null)
+  capacity_type   = lookup(each.value, "capacity_type", "ON_DEMAND")
 
   dynamic "remote_access" {
     for_each = each.value["key_name"] != "" ? [{
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -15,7 +15,7 @@ resource "aws_eks_node_group" "workers" {
 
   ami_type        = lookup(each.value, "ami_type", null)
   disk_size       = lookup(each.value, "disk_size", null)
-  instance_types  = each.value["launch_template_id"] != "" ? [] : flatten([each.value["instance_type"]])
+  instance_types  = each.value["launch_template_id"] != null ? [] : flatten([each.value["instance_type"]])
   release_version = lookup(each.value, "ami_release_version", null)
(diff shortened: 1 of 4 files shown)


zmingxie:
Could you also update the example here with the new `capacity_type` option? 
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/launch_templates_with_managed_node_groups/main.tf

This will be easier for future references

brannondorsey:
Managed spot instance node pools add support for mixed instance types. Would it be possible to make the `instance_type` accept a list of instance types in addition to a string?

psoares:
@brannondorsey I'm updating the documentation to make it obvious.

brannondorsey:
Thanks for that docs change 😸

We're currently using this module and have had to make our managed spot instance node groups out of band without Terraform. I appreciate this small PR which will make our lives much easier. 🙏

brannondorsey:
Any idea when this might land in a new release? We're looking to use this functionality this week and are contemplating if we should just vendor this branch into our code base or wait for a new release of the module with these changes included.

psoares:
What's the process to get a PR reviewed here?

worldofgeese:
> What's the process to get a PR reviewed here?

Looks like @barryib has write access. Pinging him for you.

psoares:
> > What's the process to get a PR reviewed here?
> 
> Looks like @barryib has write access. Pinging him for you.
Merci @barryib!

worldofgeese:
Maybe if I try @dpiddockcmp also

duboisf:
Hi @max-rocket-internet! Would it be possible to get a review for this PR pretty please?

sumitg892:
When will this PR get merged?

psoares:
> When will this PR get merged?

This is my first PR... I have no experience on how long it usually takes to get a review.

huiser (approved):

sumitg892:
@psoares, have you please tested your fix that it works?

modules/node_groups/node_groups.tf
```diff
@@ -15,8 +15,9 @@ resource "aws_eks_node_group" "workers" {
 
   ami_type        = lookup(each.value, "ami_type", null)
   disk_size       = lookup(each.value, "disk_size", null)
-  instance_types  = each.value["launch_template_id"] != null ? [] : [each.value["instance_type"]]
+  instance_types  = each.value["launch_template_id"] != "" ? [] : flatten([each.value["instance_type"]])
```
duboisf, line 18: What was the reason that the check was changed from verifying if the `launch_template_id` is `""` instead of `null`? I forked and tried this PR and the resulting `instance_types` was a `t3.large`. I didn't have a `launch_template_id` anywhere (`workers_group_defaults`, `node_group_defaults` or `node_groups`).

Changing the predicate back to `each.value["launch_template_id"] != null` resulted in the `instance_types` being picked up from the `instance_type` set in `var.node_groups`.

psoares: Thanks @duboisf. I don't recall exactly how, but at least on terraform 0.14, null doesn't work for me, whereas "" does.
What's your version of Terraform? Should I be testing in 0.12 instead?

psoares: It might be a side effect of https://github.com/terraform-aws-modules/terraform-aws-eks/commit/571da60aea10034cef20c435c2b49933ab5ede47

psoares: Ok, I understand now. It's indeed related to the PR I mention above. I was doing my tests by editing files directly in my .terraform folder and forgot to update the value of launch_template_id to null in local.tf. I reverted that value back to null and pushed a new commit. Please check now.

duboisf (approved):
Tested it :+1:

modules/node_groups/node_groups.tf
duboisf: Tested it, it works :+1:

psoares:
> Tested it 👍

Thank you 😉
Any idea why the merge is still blocked?

duboisf:
Yes we need one of the maintainers to approve the PR

mknapcok (approved):

antimack:
also interested in this, any estimations?

darkhelmet:
How will this work with a launch template and wanting to pick multiple instance types? Do I create a launch template per instance type and then a node group per launch template?

Basically, I have a launch template, and I want a spot instance node group of "any of these instance types" (which it looks like your example supports, if you don't specify a launch template).

Maybe that's a separate thing? I can leave the instance type off the launch template, but then the node group module ignores the instance type I specify on the group itself. Should that be the case?

psoares:
> How will this work with a launch template and wanting to pick multiple instance types? Do I create a launch template per instance type and then a node group per launch template?
> 
> Basically, I have a launch template, and I want a spot instance node group of "any of these instance types" (which it looks like your example supports, if you don't specify a launch template).
> 
> Maybe that's a separate thing? I can leave the instance type off the launch template, but then the node group module ignores the instance type I specify on the group itself. Should that be the case?
> 
> Edit: Looking at the AWS help when creating a new node group I see `When using a launch template do not specify instanceTypes, diskSize or remoteAccess in the node group configuration. These properties must be configured within the launch template itself.` which suggests it's doing things properly.

I've been scratching my head on this one for a few days. I even opened #1192. I'll let you know if I find a way. So far I've been able to do a launch template, but not override it's parameters with a launch configuration to handle multiple instance types...

barryib:
I just merged https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1129 and it adds the SPOT support for managed node groups.

@psoares can we close this PR as it duplicates https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1129 ?

psoares:
Closing since the feature was implemented and merged through #1129.

psoares:
@barryib, should we redo this PR to fix #1211 ?

barryib:
> @barryib, should we reopen this PR to fix #1211 ?

Nope. Thanks and sorry for the delay.

PR #1169, opened 2020-12-31
docs: Change instance type from `t2` to `t3` in examples

oscr:
# PR o'clock

## Description

The t2 instance type isn't available in newer regions such as eu-north-1. Therefore the examples using t2 will fail in this region.

Available instance types can be verified with: 
`aws ec2 describe-instance-type-offerings --location-type "region" --region eu-north-1 | jq '.InstanceTypeOfferings[].InstanceType | select(startswith("t"))'`

### Checklist

- [x] CI tests are passing


barryib (approved):
Thanks @oscr for your contribution.

Issue #1188, opened 2021-01-20
Remove modules/node_groups/random.tf completely - it only causes issues


pre:
# I have issues

The `random_pet` resource is a nice helper for starting up a new environment quickly. 

However, in a live cluster it will only cause maintenance pain with bad side-effects. 

See https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1109#issuecomment-763450622

## I'm submitting a...

* [X] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

The `random_pet` for node_groups is only used for dynamic node_group name. 

However, the mere existence of modules/node_groups/random.tf causes a side-effect. The random_pet will create an unrelated resource `module.eks.module.node_groups.random_pet.node_groups["test-workers"]` which is bound to the node groups. 

As a result, changing any attribute in the Launch Template bound to a Managed Node Group will cause the whole Node Group being recreated. The correct behaviour would be to update the Launch Template in-place and then update Managed Node Group to use the new Launch Template version (which can be a separate action).

## If this is a bug, how to reproduce? Please include a code sample if relevant.

See issue https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1109 for details.

## What's the expected behavior?

Updating a Launch Template must not recreate a Managed Node Group.

Only the Launch Template version should be bound to a Managed Node Group and the Managed Node Group must be updated in-place (not recreated).

## Environment details

* Affected module version: v13.2.1
* OS: Mac OS
* Terraform version: 0.13.x - 0.14.4.

## Any other relevant info

See issue https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1109 for details.

tjsampson:
Yes, we are running into some issues as well with this piece of the code. I have submitted a ticket here: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1194

treksler:
i am also having issues. my use case is to specify 3 subnets for the eks cluster, but only use 1 subnet for some node groups.

fractos:
Yep, just hit this with changing the instance type for a managed node group. Wondering what to do about it.

barryib:
I just opened this https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1188. Can you please review and test it.

Issue #1109, opened 2020-11-19
Update a launch template of a managed node group without recreating that managed node group


pre:
# I have issues with updating a launch template of a Managed Node Group

## I'm submitting a...

* [X] bug report

## What is the current behavior?

Practically any change to a Launch Template will force recreation of a Managed Node Group.

## If this is a bug, how to reproduce? Please include a code sample if relevant.

* Use eg. the examples from https://github.com/terraform-aws-modules/terraform-aws-eks/tree/e4520d2c2e1b1a182d4abb8bdaf032534f5ed89f/examples/launch_templates_with_managed_node_groups

* See concrete examples at the bottom 

* Using node_group `name` instead of `name_prefix` to remove any traces of `random_pet` does not have an effect.

* Run `terraform apply` twice, so that the second run does not show any changes.

* Update any attribute in the Launch Template. For example, do a change in the userdata template or change `block_device_mappings > ebs > volume_size`. 

## What's the expected behavior?

* Changing the Launch Template happens in-place (as it currently does), this works and Launch Template gets a new version number.

* Existing Managed Node Group is updated to the new Launch Template version number:
```
      launch_template_id      = aws_launch_template.ceph_osd.id
      launch_template_version = aws_launch_template.ceph_osd.default_version
```

* Existing Managed Node Group is not replaced with a new resource. Instead the existing MNG is kept. 

* Something causes `ami_type` , `disk_size` and `node_group_name` to change and force a replacement. I have tried setting them explicitly in `node_groups_defaults`, or in the `node_group` but it does not have an effect. I have tried leaving them out - no effect, still trying to recreate the MNG. 

## Related

* https://github.com/terraform-aws-modules/terraform-aws-eks/issues/737 Forced Replacement of launch configurations without changes in code 

## Environment details

* Affected module version: 13.2.0
* OS: MacOS 
* Terraform version: 
```
Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/aws v3.15.0
+ provider registry.terraform.io/hashicorp/kubernetes v1.13.3
+ provider registry.terraform.io/hashicorp/local v1.4.0
+ provider registry.terraform.io/hashicorp/null v2.1.2
+ provider registry.terraform.io/hashicorp/random v3.0.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
```
## Any other relevant info

<details>
<summary>Terraform output</summary>

```

[...]

PR #1189, opened 2021-01-20
fix: Updating Launch Template must not recreate Managed Node Group

pre:
- Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1109
- Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1188

# PR o'clock

## Description

Updating a Launch Template must not recreate the Managed Node Group.

* The `random_pet` resource is only used for dynamic name of the Managed Node Group. 
* The mere existence of modules/node_groups/random.tf causes a bad side-effect of MNG being recreated. 

=> Expect explicit name for each node_group and remove random.tf.

```
node_groups = {
  test-workers = {
    name                    = "${var.cluster_name}-test-workers"
    # ..
  }
}
```

### Checklist

- [ ] CI tests are passing


pre:
This PR fixes the Managed Node Group unintended recreation. 

However, this also exposes that the aws-eks-terraform module has a bug how it handles Launch Templates. 

Given a Managed Node Group:

```hcl
  node_groups = {
    test-workers = {
      name                    = "${var.cluster_name}-test-workers"
      launch_template_id      = aws_launch_template.worker_beer.id
      launch_template_version = aws_launch_template.worker_beer.default_version
      # ..
   }
  }
```

This will instruct Terraform to create a single Launch Template `worker_beer` and a single Managed Node Group `clusteName-test-workers`:

```
  # aws_launch_template.worker_beer will be created
  + resource "aws_launch_template" "worker_beer" {
      + arn                    = (known after apply)
      + default_version        = (known after apply)
      + description            = "Worker Launch-Template"
      + id                     = (known after apply)
      + instance_type          = "m5.large"
      + latest_version         = (known after apply)
      + name                   = "aws-eu-central-1-3-worker-beer"
      + tags                   = {
          + "NodeRole" = "worker"
        }
      + update_default_version = true
      # ... 
    }

  # module.eks.module.node_groups.aws_eks_node_group.workers["test-workers"] will be created
  + resource "aws_eks_node_group" "workers" {
      + ami_type        = (known after apply)
      + arn             = (known after apply)
      + capacity_type   = (known after apply)
      + cluster_name    = "aws-eu-central-1-3"
      + id              = (known after apply)
      + instance_types  = (known after apply)
      + node_group_name = "aws-eu-central-1-3-test-workers"
      + node_role_arn   = "arn:aws:iam::864348345328:role/aws-eu-central-1-32021010709125500060000000a"
      + version         = (known after apply)

      + launch_template {
          + id      = (known after apply)
          + name    = (known after apply)
          + version = (known after apply)
        }

      # ... 
    }
```

2021-01-20 at 14 58


What goes wrong is that after this Terraform is applied, AWS EKS has created a *second* Launch Template which is bound to the Worker Nodes of the Managed Node Group.

As a result, the creation of the Managed Node Group succeeds, but the updates later will fail, because the Worker Nodes are bound to the Launch Template created implicitly by EKS (*not* Terraform). 

Once the Launch Template has been modified, a newer version of the Launch Template is created. The problem is, that the nodes are bound to a wrong Launch Template! (the one created by EKS).

This issue was masked before, because the `random_pet` caused the Managed Node Group always be recreated when Launch Template got changed. Hence, the Managed Node Group was in "pristine" state after each "terraform apply". 

Now, when the Managed Node Group is _updated_ (instead of recreated) to use a newer version of the Launch Template, the 2..n attempts of "terraform apply" will fail because there exists an Launch Template created implicitly by EKS. 


## Userdata of Launch Template updated, Launch Template gets new version 
```hcl
Terraform will perform the following actions:

  # aws_launch_template.worker_beer will be updated in-place
  ~ resource "aws_launch_template" "worker_beer" {
      ~ default_version         = 1 -> (known after apply)
        id                      = "lt-0fa87354118a95faa"
      ~ latest_version          = 1 -> (known after apply)
        name                    = "aws-eu-central-1-3-worker-beer"
        tags                    = {
            "NodeRole" = "worker"
        }
      ~ user_data               = "original" -> "changed"
      # .. 
    }

  # module.eks.module.node_groups.aws_eks_node_group.workers["test-workers"] will be updated in-place
  ~ resource "aws_eks_node_group" "workers" {
        id              = "aws-eu-central-1-3:aws-eu-central-1-3-test-workers"
        tags            = {
            "NodeRole"                                     = "worker"
            "k8s.io/cluster-autoscaler/aws-eu-central-1-3" = "true"
            "k8s.io/cluster-autoscaler/enabled"            = "true"
        }

      ~ launch_template {
            id      = "lt-0fa87354118a95faa"
            name    = "aws-eu-central-1-3-worker-beer"
          ~ version = "1" -> (known after apply)
        }
      # .. 
    }
```

2021-01-20 at 15 06


TL;DR The existing behaviour of `random_pet` is indeed incorrect, but it only masks under itself a more serious bug about how the Launch Templates are currently handled.

pre:
I tried really hard and failed. I have no idea what causes the second Launch Template to be created. 

I commented out all possible references from any place I could find in the aws-eks-terraform module and I hard-coded over default values but nothing helps - this implicit Launch Template always gets created and it breaks all subsequent updates with "terraform apply". The first "terraform apply" (create) always succeeds. 

This issue might have something to do with the general complexity of this module (as discussed in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635) . The default values and resources are shared for different use cases, some of which conflict with each other. 

For example, the result is different when 
- you want explicit Launch Template with Managed Node Group and you define values of your own
- you don't specify a Launch Template at all but expect an implicit Launch Template still be created with default values
- you don't use Managed Node Group but still use Launch Templates with Autoscaling Groups

I'm taking a look into https://github.com/cloudposse/terraform-aws-eks-cluster and https://github.com/cloudposse/terraform-aws-eks-node-group 

The current behaviour of Launch Templates with Managed Node Groups in aws-eks-terraform is badly broken and I don't know how to fix it.

pre:
https://docs.amazonaws.cn/en_us/eks/latest/userguide/create-managed-node-group.html
>  If you don't use your own launch template, the Amazon EKS API creates a default Amazon EC2 launch template in your account and deploys the node group using the default launch template. 

So somehow the Managed Node Group is not correctly bound to the Launch Template, even though it's created an updated by Terraform. Hence, the EKS API creates the implicit Launch Template even though the explicit one was defined.

surajnarwade:
I tested out this solution and I can see, it is still re-creating the node group and there are multiple factors apart from launch template which are responsible:

```
  # module.eks.module.node_groups.aws_eks_node_group.workers["example"] must be replaced
+/- resource "aws_eks_node_group" "workers" {
      ~ ami_type        = "AL2_x86_64" -> (known after apply) # forces replacement

      ~ capacity_type   = "ON_DEMAND" -> (known after apply)
        cluster_name    = "test-eks-lt-smq9vtin"
      ~ disk_size       = 0 -> (known after apply) # forces replacement
      ~ id              = "test-eks-lt-smq9vtin:example" -> (known after apply)
      ~ instance_types  = [] -> (known after apply) # forces replacement
      ~ labels          = {} -> (known after apply)
      ~ node_group_name = "example" -> (known after apply) # forces replacement
      ~ node_role_arn   = "arn:aws:iam::xxxxxxxxx:role/test-eks-lt-smq9vtin2021012115495815060000000b" -> (known after apply) # forces replacement


      ~ subnet_ids      = [
          - "subnet-04fcefa5867f12ff6",
          - "subnet-0be499e0dbc251ceb",
          - "subnet-0c0f2fca22c117804",
        ] -> (known after apply) # forces replacement
      ~ tags            = {
          - "CustomTag" = "EKS example"
        } -> (known after apply)
        
      ~ version         = "1.17" -> (known after apply)

      ~ launch_template {
          ~ id      = "lt-02aa34d795bf5d77f" -> (known after apply) # forces replacement
          ~ name    = "eks-example-20210121153441183000000009" -> (known after apply)
          ~ version = "3" -> (known after apply)
        }

      + remote_access { # forces replacement
          + ec2_ssh_key               = (known after apply) # forces replacement
          + source_security_group_ids = (known after apply) # forces replacement
        }

  
    }
```

surajnarwade:
interesting point is:

```
~ node_group_name = "example" -> (known after apply) # forces replacement
```

where I am not at all changing the name

pre:
I would be happy with any solution which does not recreate the node group but which would update the associated launch template in-place (and not create the implicit launch template at all).

For some reason it does not seem to be possible with the aws-eks-terraform module.

surajnarwade:
@pre you can refer: https://github.com/umotif-public/terraform-aws-eks-node-group

darkhelmet:
After messing around a bunch, I've come to a few conclusions. These are obviously just my assumptions.

**EKS creates the second launch template so that it can manage the AMI.** If you try to specify a launch template with the AMI set to the node groups, you get `InvalidParameterException: You cannot specify an image id within the launch template, since your nodegroup is configured to use an EKS optimized AMI.` I imagine the ASG needs an AMI to function properly, so the LT it gets is the duplicate one that EKS manages. I also notice EKS doesn't update the default version of it's managed LT, because if you click from the ASG to the LT you get the default version 1, even though the ASG is configured to use the latest verison.

pre:
Removing the random_pet does not fix the problem. I ended up building my own terraform without all the extra stuff provided by aws-eks-terraform which I didn't really need.

PR #1204, opened 2021-01-29
docs: Fix cluster autoscaler tags in IRSA example

gisripa:
# PR o'clock

## Description

Based on [this](https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html) reference , Looks like `k8s.io/cluster-autoscaler/${local.cluster_name}` should be `owned`. Please let me know if I misinterpreted this.

### Checklist

- [ ] CI tests are passing


barryib (approved):
Nice catch. Thansk @gisripa

PR #1228, opened 2021-02-07
improvement: use Fargate profile name as for_each index for aws_eks_fargate_profile

this-chris:
# PR o'clock

## Description

We made the experience that adding and/or removing Fargate profiles regenerates all profiles that belong to the cluster. That drives in a maintenance problem, as all pods will be shut down when a related profile will be re-created.

To keep the Fargate profiles a bit more maintainable (when removing or/and adding a profile) the key of the `set`  `fargate_profiles_expanded` is now the name of the profile. The profile name needs to be always unique (constraint by AWS).

Note: That's a breaking change, as all created Fargate profiles by this module need to be re-created with the new index.

### Checklist

- [x] CI tests are passing


barryib:
> We made the experience that adding and/or removing Fargate profiles regenerates all profiles that belong to the cluster.

This is not normal. `var.fargate_profiles` is a map, so until you change the map key, your fargate profile won't be destroyed.

```hcl
fargate_profiles = {
  example = {
    foo = "default"
    tags = {
      Owner = "test"
    }
  },
  bar = {
    namespace = "default"
    tags = {
      Owner = "test"
    }
  }
}
```

So please just use unique key and destroy them.

Issue #1162, opened 2020-12-25
tf destroy fails to remove aws_auth: unauthorized


astanciu:
# I have issues

## I'm submitting a...

* [x ] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
When I do a destroy operation, I receive 
```
Error: Unauthorized
```

The only remaining piece of state is the aws_auth module:
```
 # module.kubernetes_cluster.module.eks.kubernetes_config_map.aws_auth[0] will be destroyed
```

## Environment details
```
module "eks" {
  source = "terraform-aws-modules/eks/aws"

  cluster_name     = var.cluster_name
  cluster_version  = "1.18"
  subnets          = module.vpc.private_subnets
  iam_path         = "/eks/"
  write_kubeconfig = false

  kubeconfig_aws_authenticator_command = "aws"
  kubeconfig_aws_authenticator_command_args = [
    "--region",
    var.region,
    "eks",
    "get-token",
    "--cluster-name",
    var.cluster_name,
  ]

  cluster_encryption_config = [
    {
      provider_key_arn = aws_kms_key.eks.arn
      resources        = ["secrets"]
    }
  ]

  tags = merge(
    var.tags
  )

  vpc_id = module.vpc.vpc_id

  worker_groups = [
    {
      name                 = "worker"
      instance_type        = var.worker_instance_type
      asg_desired_capacity = var.number_of_workers
      asg_max_size         = 5
      iam_role_id          = "k8s-node"
    }
  ]

  workers_additional_policies = [
    "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore",
    aws_iam_policy.nodes-ec2-policy.arn
  ]
}
```

* Affected module version: 3.2.1
* OS: osx 10.15 
* Terraform version: 0.14.3

spaziran:
I'm also getting this error... using a bare-bones install:
```
data "aws_eks_cluster" "cluster" {
  name = module.my-cluster.cluster_id
}

data "aws_eks_cluster_auth" "cluster" {
  name = module.my-cluster.cluster_id
}

provider "kubernetes" {
  host                   = data.aws_eks_cluster.cluster.endpoint
[...]

PR #1230, opened 2021-02-10
fix: update supported tf version to have < 0.14.0

aidanmelen:
modules fails to destroy EKS cluster in 0.14.5 because destroys don't refresh.

# PR o'clock

## Description

> In terraform 0.14+ destroy command no loner refreshes the state of resources before generating execution plan (like it did in 0.13.X). terraform apply still does the refresh, that's why this issue is more frequent the more time elapses after applying.

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1162#issuecomment-767700530

As a result, the module will fail to destroy all of the resources it manages.

I should say that there is a workaround in 0.14.5. The user will be required to run `terraform refresh` then `terraform destroy`.

### Checklist

- [x] CI tests are passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 
 | Name | Version |
 |------|---------|
-| terraform | >= 0.12.9, != 0.13.0 |
+| terraform | >= 0.12.9, != 0.13.0, < 0.14.0 |
 | aws | >= 3.22.0 |
 | kubernetes | >= 1.11.1 |
 | local | >= 1.4 |
--- a/versions.tf
+++ b/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 0.12.9, != 0.13.0"
+  required_version = ">= 0.12.9, != 0.13.0, < 0.14.0"
 
   required_providers {
     aws        = ">= 3.22.0"


daroga0002:
I dont think does limiting terraform version is good approach. I would rather try to solve a root cause than limiting terraform updates, especially that 0.14 is considered now as desired and stable version

aidanmelen:
@daroga0002 I agree, let's fix it because this problem isn't going away. But this module claims to support 0.14.0 when that is not entirely true. Perhaps we should add a note to the README that says:

> If you are using a terraform version greater than 0.14.0, you will need to run `terraform refresh` prior to running `terraform destroy`.

I just thought something like this would save people the 30 minutes of head scratching when the terraform destroy fails.

daroga0002:
@aidanmelen can you describe a issue more (where it is failing, does those are only some module inputs required)?

aidanmelen:
@daroga0002 The error can be reproduced by running the [basic example](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/basic) on terraform 0.14.X.

The specific resource that the module fails to delete is:

> Error: Unauthorized

> module.kubernetes_cluster.module.eks.kubernetes_config_map.aws_auth[0] will be destroyed

There is a lot more context in this related issue #1162 than I can provide here on my own. This seems to be causing confusion for a number of people.

dak1n1:
Prior to Terraform version 0.14, data sources would be [refreshed before destroy](https://github.com/hashicorp/terraform/pull/27408). This functionality has been added back to [Terraform 0.15](https://github.com/hashicorp/terraform/releases/tag/v0.15.0-beta1). It's an issue with Terraform core. Unless they decide to backport that fix to 0.14, it makes sense to limit the versions that are used with this module.

dak1n1:
It looks like the fix will be too difficult to backport to 0.14. https://github.com/hashicorp/terraform/issues/28179#issuecomment-805797078

antonbabenko:
We don't have to restrict the max version of the Terraform in the module.

Btw, [v15.1.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v15.1.0) has been just released where Terraform 0.15 is supported and the minimum version of Terraform is updated to be 0.13.1.

PR #1239, opened 2021-02-12
fix: ManagedNodeGroups using LaunchTemplate and while specifying Disk Size

theothermike:
If using Managed Node Groups with a Launch Template, the disk size MUST be specified in the Launch Template and not in the Managed Node Group definition otherwise AWS API produces an error that Disk Size is specified

# PR o'clock

## Description

If the `launch_template_id` is specified, specify null for disk size in managed node group resource

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -14,7 +14,7 @@ resource "aws_eks_node_group" "workers" {
   }
 
   ami_type        = lookup(each.value, "ami_type", null)
-  disk_size       = lookup(each.value, "disk_size", null)
+  disk_size       = each.value["launch_template_id"] != null ? null : lookup(each.value, "disk_size", null)
   instance_types  = lookup(each.value, "instance_types", null)
   release_version = lookup(each.value, "ami_release_version", null)
   capacity_type   = lookup(each.value, "capacity_type", null)


theothermike:
I don't understand what the error is on Docs, the failure doesn't seem to indicate a reason

ArchiFleKs:
This is included here https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1138/files#diff-7a3fc6c7df17fda0c341e61255461bf1f149256a9ddf14d4a18ab6f020d08136R17

philicious:
@theothermike do you actually face an error here? 

I use MNG + LT since #997 and dont have problems with disk_size. in fact, disk_size is not defaulted for MNG, so if you dont specify it on the node_group and only in the LT, there shouldnt be a problem

theothermike:
> @theothermike do you actually face an error here?
> 
> I use MNG + LT since #997 and dont have problems with disk_size. in fact, disk_size is not defaulted for MNG, so if you dont specify it on the node_group and only in the LT, there shouldnt be a problem

Yes, the API throws an error that either you don't have a disk defined in the launch template, or if you do, and don't have this fix, another error saying disk size is defined.  but the other fix mentioned above is more comprehensive, so will close this one and wait for that to be merged

theothermike:
closing in favour of #1138

PR #1253, opened 2021-02-23
improvement: Use time_sleep instead of local-exec

shoekstra:
# PR o'clock

## Description

Using `time_sleep` is a more platform agnostic way to wait for something than using a `local-exec`. As the
previous way had a `5m` timeout, the same is implemented using `time_sleep`.

With this change we're able to provision EKS clusters and manage the `aws_auth` config map while still using the `hashicorp/tfc-agent` docker image.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,191 +140,129 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.13.1 |
-| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.22.0 |
-| <a name="requirement_kubernetes"></a> [kubernetes](#requirement\_kubernetes) | >= 1.11.1 |
-| <a name="requirement_local"></a> [local](#requirement\_local) | >= 1.4 |
-| <a name="requirement_null"></a> [null](#requirement\_null) | >= 2.1 |
-| <a name="requirement_random"></a> [random](#requirement\_random) | >= 2.1 |
-| <a name="requirement_template"></a> [template](#requirement\_template) | >= 2.1 |
+| terraform | >= 0.13.1 |
+| aws | >= 3.22.0 |
+| kubernetes | >= 1.11.1 |
+| local | >= 1.4 |
+| null | >= 2.1 |
+| random | >= 2.1 |
+| template | >= 2.1 |
 
 ## Providers
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 3.22.0 |
-| <a name="provider_kubernetes"></a> [kubernetes](#provider\_kubernetes) | >= 1.11.1 |
-| <a name="provider_local"></a> [local](#provider\_local) | >= 1.4 |
-| <a name="provider_null"></a> [null](#provider\_null) | >= 2.1 |
-| <a name="provider_random"></a> [random](#provider\_random) | >= 2.1 |
-| <a name="provider_template"></a> [template](#provider\_template) | >= 2.1 |
-
-## Modules
-
-| Name | Source | Version |
-|------|--------|---------|
-| <a name="module_fargate"></a> [fargate](#module\_fargate) | ./modules/fargate |  |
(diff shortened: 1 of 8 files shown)


daroga0002:
Issue which I see here is that AWS doesn't guarantee Control planes availability. This means this can take 2 minutes, but can take also 15 minutes. 

In your PR you assuming that this will take max 5 minutes, in result this will cause that module can fail even if everything is going as should. From other side increasing time_sleep to 10 minutes can cause non required delay in module execution.



Just to be clear I dont like this `local-exec` hack but I dont see currently any other solution to solve it in user friendly way without causing more issues.

shoekstra:
Hi @daroga0002 !

> Issue which I see here is that AWS doesn't guarantee Control planes availability. This means this can take 2 minutes, but can take also 15 minutes.
> 
> In your PR you assuming that this will take max 5 minutes, in result this will cause that module can fail even if everything is going as should. From other side increasing time_sleep to 10 minutes can cause non required delay in module execution.

I can understand the hesitation, it's certainly not the ideal situation... Re max 5 minutes, this is already the case when you look at `var.wait_for_cluster_cmd`, so no change there. The only perk to the existing method is if it's ready within 5 minutes the creation of the rest of the resources happens sooner.

> Just to be clear I dont like this `local-exec` hack but I dont see currently any other solution to solve it in user friendly way without causing more issues.

I can appreciate you can't take my word for it, but I've deployed a few clusters and they've regularly been ready when the 5m timer is reached... 

Due to using Terraform Cloud to bootstrap our EKS clusters, we're confined to using the `hashicorp/tfc-agent` image (I assume it's this anyway) which doesn't have the `wget` or `curl` binaries installed. If the concern is the 5 minute timer, I'd be more in favour with increasing this to a higher number (or making it configurable) and doing this "natively" - waiting a few more minutes to stand up an EKS cluster wouldn't be the end of the world.

Look forward to your feedback!

Stephen

daroga0002:
Yup, I understand your arguments. So I think they are reasonable so from my side  👍 

Some technical think is that you must add  Semantric pull Request prefix into PR subject choosing one from:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/a26c9fd0c9c880d5b99c438ad620e91dda957e10/.github/semantic.yml#L18-L26

I think it should be:
`improvement: Use time_sleep instead of local-exec`

shoekstra:
Great!

I've updated the title as requested, will look at the docs lint job later.

daroga0002:
I am adding this link https://aws.amazon.com/about-aws/whats-new/2021/03/amazon-eks-reduces-cluster-creation-time-40-percent/ as this is also good news in this context. 


@barryib  can you review this?

shoekstra:
Hey @daroga0002, I've rebased this PR and regenerated the docs using `pre-commit run -a`. 

Is there anything still needed to get this merged and released?

daroga0002:
> Is there anything still needed to get this merged and released?

For me it looks good, but I am not maintainer of this repo so I cannot approve :(

@barryib

shoekstra:
Hi @max-rocket-internet @barryib!

Are either of you able to take a look?

Thanks in advance.

barryib:
Hello @shoekstra Thanks for opening this PR and @daroga0002 for all your reviews (really appreciated).

I truly understand your concern here. The first time we introduced this check with curl (or wget as fallback), we knew it was not great solution. But as @daroga0002 mentioned it, the EKS availability is quite random and waiting 5mn, will slow down every EKS cluster creation by 5mn.

There is a issue on AWS side to track this https://github.com/aws/containers-roadmap/issues/654
And here is a try to resolve this in the AWS provider https://github.com/hashicorp/terraform-provider-aws/pull/11426

Since we now publish our own provider with terraform registry, I had in mind to publish the http provider with the support of http timeout https://github.com/hashicorp/terraform-provider-http/pull/71. This fork will be used only for this timeout feature until we go a real solution.

barryib:
FIW, I started this https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1339 for tests.

barryib:
@shoekstra thank you for your work. I'm closing your PR, since we merged https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1339. I assume that PR addressed your issue. Feel free to reopen it if needed.

Thanks @daroga0002 for your help.

PR #1289, opened 2021-03-31
feat: Launch Template ASGs pin to created version

lowjoel:
# PR o'clock

## Description

When creating worker node groups with Launch Templates, the created ASGs currently use $Latest as the version of the Launch Template to start nodes with. When `asg_recreate_on_change` is false, that's expected, but when `asg_recreate_on_change` is true, this means that as soon as a new Launch Template is created the old ASGs will start using the new launch template.

This changes the logic to support a `$Created` value, which would then reference the version created in Terraform. When `asg_recreate_on_change` is false, the ASG still points to the latest (as soon as Terraform gets to it). When `asg_recreate_on_change` is true, the older ASGs (when using `asg_recreate_on_change` would still reference the older LT version, and so compute capacity won't be lost should the new LT have problems spinning up nodes.

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -76,7 +76,7 @@ locals {
     root_block_device_name               = data.aws_ami.eks_worker.root_device_name # Root device name for workers. If non is provided, will assume default AMI was used.
     root_kms_key_id                      = ""                                       # The KMS key to use when encrypting the root storage device
     launch_template_id                   = null                                     # The id of the launch template used for managed node_groups
-    launch_template_version              = "$Latest"                                # The lastest version of the launch template to use in the autoscaling group
+    launch_template_version              = "$Created"                               # The created version of the launch template to use in the autoscaling group
     launch_template_placement_tenancy    = "default"                                # The placement tenancy for instances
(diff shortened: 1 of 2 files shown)


lowjoel:
Not sure why docs are failing - let me know if something requires fixing.

lowjoel:
Hey @barryib @james-callahan - anything I need to do to get a review on this please?

barryib:
I think this will be solved in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1370. What do you think ?

lowjoel:
Yeah I think that could work too. It implements the same thing in a different way. If that gets merged I think this one can be closed.

barryib:
> Yeah I think that could work too. It implements the same thing in a different way. If that gets merged I think this one can be closed.

Great. So let close this one. Thanks again for your help.

barryib:
It's now implemented in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1370

PR #1292, opened 2021-04-06
feat: support overriding worker security groups

almonteb:
in order to support node groups which span VPCs, there needs to be a way
to not attach the default worker security group id whether or not its
managed by this module as SGs do not span VPCs.

while not a "supported" configuration from EKS, it is possible to
accomplish. see: https://github.com/aws/containers-roadmap/issues/153

this functionality is supported on both worker_groups and
worker_group_launch_templates, and does not have a default.
security_group_ids takes priority and will be the only security groups
attached to a launch config or template when present. if not provided,
it will fall back to the existing logic of flattening
worker_security_group_id with any overrides or defaults provided by
additional_security_group_ids.


daroga0002:
closing this PR as mentioned this configuration is not supported by AWS. 


Additionally if users want to make such configuration then they can always disable creation of security group by module using input `worker_create_security_group`

PR #1165, opened 2020-12-29
fix: Updated code and version requirements to work with Terraform 0.15

TjeuKayim:
# PR o'clock

## Description

Replaced use of [deprecated](https://www.terraform.io/docs/configuration/functions/list.html) `list()` function with `[]` syntax and `map()` function with `{}` syntax.

Fix #1164

### Checklist

- [x] CI tests are passing
- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #1307, opened 2021-04-16
fix: Fixed list and map usage

antonbabenko:
Follow up for #1165


antonbabenko:
[v15.1.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v15.1.0) has been just released.

Issue #1124, opened 2020-11-27
Capacity Rebalance support


zeyaddeeb:
# I have issues

I was looking to add capacity rebalance to some worker launch templates and ran into issues:

exampe:

```terraform
resource "aws_autoscaling_group" "example" {
  capacity_rebalance = true
}

```
as per this PR: [https://github.com/hashicorp/terraform-provider-aws/pull/16127](https://github.com/hashicorp/terraform-provider-aws/pull/16127)

## I'm submitting a...

* [ ] bug report
* [X] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [X] kudos, thank you, warm fuzzy

## What is the current behavior?

When adding `capacity_rebalance = true` in the worker launch template will not work since there are no lookup values.

## Environment details

* Affected module version: latest
* OS: ubuntu18.04
* Terraform version: 0.13

## Any other relevant info

[https://github.com/hashicorp/terraform-provider-aws/pull/16127](https://github.com/hashicorp/terraform-provider-aws/pull/16127)

AdamTylerLynch:
@zeyaddeeb  you mention a Launch Template, but your example is showing creating an AutoScaling group. Can you please share your code so we can help troubleshoot?

AdamTylerLynch:
Capacity Rebalance is not a property of a launch template, it is a property of the autoscaling group through. You can create an Auto Scaling group and associate it with a launch template (or multiple launch templates in the latest versions of the SDK).

zeyaddeeb:
Maybe using the wrong terminology, but it would look something like:

```terraform
...

  worker_groups_launch_template = [
    {
      name                                     = "worker1"
      subnets                                  = var.subnet_ids
      asg_min_size                             = 1
      asg_desired_capacity                     = 1
      asg_max_size                             = 1
      autoscaling_enabled                      = true
      capacity_relabance                       = true   # desired variable
    },

...
```

AdamTylerLynch:
@zeyaddeeb are you potentially referring to the Terraform AWS EKS module?
https://github.com/terraform-aws-modules/terraform-aws-eks


`module "eks" {
  source          = "../.."
  cluster_name    = local.cluster_name
  cluster_version = "1.17"
  subnets         = module.vpc.public_subnets
  vpc_id          = module.vpc.vpc_id

  worker_groups_launch_template = [
    {
[...]

PR #1326, opened 2021-04-29
feat: Add `capacity_rebalance` support for self-managed worker groups

LAKostis:
## Description

- See https://docs.aws.amazon.com/autoscaling/ec2/userguide/capacity-rebalance.html. We already have most functionality, only this parameter is missing

Resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1124

### DIFF
--- a/workers_launch_template.tf
+++ b/workers_launch_template.tf
@@ -96,6 +96,11 @@ resource "aws_autoscaling_group" "workers_launch_template" {
     "health_check_grace_period",
     local.workers_group_defaults["health_check_grace_period"]
   )
+  capacity_rebalance = lookup(
+    var.worker_groups_launch_template[count.index],
+    "capacity_rebalance",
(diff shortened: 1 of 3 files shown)


workers_launch_template.tf
```diff
@@ -97,6 +97,11 @@ resource "aws_autoscaling_group" "workers_launch_template" {
     "health_check_grace_period",
     local.workers_group_defaults["health_check_grace_period"]
   )
+  capacity_rebalance = lookup(
```
barryib, line 100: Can you please do the same for ASG for workers (Launch Configuration). See workers.tf.

barryib (changes requested):

workers_launch_template.tf
LAKostis: done

barryib (approved):
Thanks @LAKostis for your contribution.

barryib:

PR #1342, opened 2021-05-07
avoid overwrite default worker group tags

marcobarducci:
# PR o'clock

## Description

this change avoid the overwrite of worker_groups_default tags if you add a specific worker_groups_launch_template tag.
so if you add tags in worker_group_defaults (for all nodes) they're not overwritten by assigning a tag in a worker_groups_launch_template (for a specific node-group).

### Checklist

- [ ] CI tests are passing


daroga0002:
I think that this was general purpose of this part of code. So you can overwrite tags by setting `worker_groups_launch_template` in map defining worker group launch template or if you dont specify that parameter then it will use `default` which is `local.workers_group_defaults["tags"]`


Maybe can you describe more what is real use case and value on some config files?

marcobarducci:
hi, this feature could be a bit annoying when you add the cluster-autoscale tags in the default worker group (as you want to add them on all the worker groups of your eks cluster) and you want a specific tag only to a single worker_group. that single worker group will not inherit the tags from default worker group.

workers_launch_template.tf
```diff
@@ -219,8 +219,9 @@ resource "aws_autoscaling_group" "workers_launch_template" {
       lookup(
         var.worker_groups_launch_template[count.index],
         "tags",
-        local.workers_group_defaults["tags"]
-      )
+        []
```
daroga0002, line 222: this will produce empty tag which will be concat in line 193, so this whole list should be also compacted (in line 193)

daroga0002:
@marcobarducci does you will be able to correct this PR?

lisfo4ka:
Hey, hey, folks, that's not so easy)

@marcobarducci mentioned quite a real use case. But if we decide to provide an option to merge the tags, we should define the strong order which tags are first and which last.

Moreover, the changes should be provided all over the module - for instances, volumes, ENIs in worker groups with launch template and launch config. And do not forget to check node groups resources tagging as well.

lisfo4ka:
@marcobarducci are you ready to update and push this PR further?

Issue #1348, opened 2021-05-12
Terraform managed worker security group doesn't honour its description


jack1902:
## Description

`worker` Security Group: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/781f673295dd69d24c85d10b47cb03507f394402/workers.tf#L344

The worker security group created by Terraform, doesn't honour its description. It states that it is for all node groups, however it is only used if `create_launch_template` is set to `true` (its default is false 👀 ) per node group. This in turn results in a weirdness were the default behaviour of a Node/Worker is to have the AWS EKS created Cluster Security group (this is seperate to the `cluster` security group created by the module).

This issue becomes even more apparent when you have configured a mesh overly on your cluster where for example:

- Node Group 1 has `create_launch_template` set to `false` (or unset, defaulting to false in the module)
- Node Group 2 has `create_launch_template` set to  `true`.

This results in the two seperate Node groups being completely unable to communicate with each other, completely breaking a mesh network 😢.

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/781f673295dd69d24c85d10b47cb03507f394402/modules/node_groups/node_groups.tf#L47

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

 - Terraform: v0.15.3

 - Provider(s): AWS v3.39.0 (others too, but not relevant to this issue)

 - Module: 15.2.0

### Code Snippet to Reproduce

Using https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/managed_node_groups/main.tf set node_groups to the following:

```tf
node_groups = {
    no_worker_sg = {
      desired_capacity = 1
      max_capacity     = 1
      min_capacity     = 1

      instance_types = ["m5.large"]
      capacity_type  = "SPOT"
    },
    worker_sg = {
      desired_capacity = 1
      max_capacity     = 1
      min_capacity     = 1

      instance_types = ["m5.large"]
      capacity_type  = "SPOT"

      # This being the key difference between the two node_groups
      create_launch_template = true
    }
}
```

## Expected behavior

[...]

PR #1349, opened 2021-05-12
fix: add create_launch_template to node group pet marker

jack1902:
# PR o'clock

relates to: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1348

## Description

When changing `create_launch_template` ONLY on my node_groups, it caused terraform to blow up because it couldn't create a new node group with the same name. This took me by surprise.

### Checklist

- [x] CI tests are passing


jack1902:
@barryib appreciate you are probably very busy, but any chance this could get a review?

jack1902:
@barryib ok, assuming that PR gets merged soon then i'm happy for this to be closed 

This feature missing was a PITA to resolve when i needed to cycle a full node group, which resulted in terraform erroring out as the nodegroup already existed. We went from no Launch Template to having one which didn't change the pet-name but resulted in a re-creation because of #1348 meant our entire mesh network died due to nodes having different security groups based on launch template

PR #1224, opened 2021-02-05
feat: Add support for Auto Scaling Group Instance Refresh for self-managed worker groups

bashims:
# PR o'clock

## Description

Add support for the instance_refresh feature of the AWS provisioner added in https://github.com/hashicorp/terraform-provider-aws/pull/16678

This PR resolves #929 

See:
https://aws.amazon.com/blogs/compute/introducing-instance-refresh-for-ec2-auto-scaling/

### Checklist

- [x] CI tests are passing
- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #1360, opened 2021-05-17
feat: Drop `asg_recreate_on_change` feature to encourage the usage of  Instance Refresh for EC2 Auto Scaling

barryib:
# PR o'clock

## Description

This module used random_pet resources to create a random name for the autoscaling group to force the autoscaling group to be re-created when the launch configuration or launch template was changed (if `recreate_asg_when_lc_changes = true` was set), causing the instances to be removed and re-provisioned each time there was an update. Those random_pet resources has been removed and in its place there is now a set of functionality provided by AWS and the Terraform AWS provider - instance_refresh. We encourage those users to move on `instance_refresh`.

Go with https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1224

### Checklist

- [x] CI tests are passing


antonbabenko (approved):

barryib:
@antonbabenko  this was quick 💪

Issue #1247, opened 2021-02-17
EKS version 1.19 unable to find ami


emilchitas:
# I have issues

## I'm submitting a...

* [X] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

When trying to deploy an EKS cluster, version 1.19, I get the following error: 
`Error: Your query returned no results. Please change your search criteria and try again.` 
It seems as if the module is unable to find a passing ami for the nodes.

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Deploy a version 1.19 EKS cluster
```
module "eks" {
  source                          = "terraform-aws-modules/eks/aws"
  version                         = "14.0.0"
  cluster_name                    = "${var.name}-${var.environment}-cluster"
  cluster_version                 = "1.19"
  subnets                         = var.private-subnets
  vpc_id                          = var.vpc_id
  cluster_endpoint_private_access = true
  cluster_endpoint_public_access  = false
  enable_irsa                     = true
  write_kubeconfig                = false

  node_groups = {
    default = {
      name             = "${var.name}-${var.environment}-node-group"
      min_capacity     = 1
      desired_capacity = var.eks_node_group.desired_capacity
      max_capacity     = var.eks_node_group.max_capacity
      instance_types   = [var.eks_node_group.instance_type]
      key_name         = var.keypair_name
      tags = [
        {
          "key"                 = "k8s.io/cluster-autoscaler/enabled"
          "propagate_at_launch" = "false"
          "value"               = "true"
        },
        {
          "key"                 = "k8s.io/cluster-autoscaler/${local.cluster_name}"
          "propagate_at_launch" = "false"
          "value"               = "true"
        }
      ]
    }
  }
  map_roles = local.roles
  map_users = local.users
  tags      = local.common_tags
}
```

## What's the expected behavior?

Terraform should be able to find a suitable ami for this cluster version.

## Are you able to fix this problem and submit a PR? Link here if you have already.

No

## Environment details

* Affected module version: 14.0.0
* OS: Ubuntu
* Terraform version: 0.14.6

## Any other relevant info

None

alghanmi:
I am able to reproduce this issue.

[...]

Issue #1246, opened 2021-02-17
Upgrade cluster version is not supported


mkamel93:
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
I have provisioned an EKS cluster with Kubernetes version 1.18 successfully then I tried to update the Kubernetes version to 1.19 but when I am trying to run 
```terraform plan``` 
I got the following error: **Error: Your query returned no results. Please change your search criteria and try again.**

## What's the expected behavior?
It should update the cluster kubernetes version same as updating it from the GUI

## Environment details

* Affected module version: v14.0.0
* OS: Linux, Mac
* Terraform version: v0.14.3

jochen-st:
same with

Terraform v0.14.6
+ provider registry.terraform.io/hashicorp/aws v3.27.0
+ provider registry.terraform.io/hashicorp/helm v1.3.2
+ provider registry.terraform.io/hashicorp/kubernetes v2.0.2
+ provider registry.terraform.io/hashicorp/local v2.0.0
+ provider registry.terraform.io/hashicorp/null v3.0.0
+ provider registry.terraform.io/hashicorp/random v3.0.1
+ provider registry.terraform.io/hashicorp/template v2.2.0

italolelis:
This is related to #1141. As proposed by @dbirks, a workaround would be to set 

`worker_ami_name_filter_windows = "Windows_Server-2019-English-Core-EKS_Optimized-1.18-*"` 

even if you don't have any windows servers, at least until AWS releases the windows version to 1.19.

It worked fine for me. Thanks to @dbirks.

aholbreich:
@italolelis work aour don't works for me. I'm using node_groups

barryib:
I just opened https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1371 to address this. Will you please review it and test it ?

Issue #1141, opened 2020-12-10
module returns "query returned no results" when using a non existent cluster_version


mahpatil:
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
Get "Error: Your query returned no results. Please change your search criteria and try again." if the cluster_version specified is incorrect.

## If this is a bug, how to reproduce? Please include a code sample if relevant.
```
module "eks-cluster" {
  source          = "terraform-aws-modules/eks/aws"
  cluster_name    = var.cluster_name
  cluster_version = var.cluster_version //set this to "1.19" currently AWS latest v 1.18
  subnets         = var.subnets
  vpc_id          = var.vpc_id

  worker_groups = [
    {
      instance_type         = var.instance_type
      asg_max_size          = var.asg_max_size
      asg_desired_capacity  = var.asg_desired_capacity
      additional_userdata   = var.additional_userdata
    }
  ]
  tags = var.tags
}
```

## What's the expected behavior?
I would expect it to fail with a bit more friendly message but I get below error, it is confusing and doesn't lead to specific field.
Error: Your query returned no results. Please change your search criteria and try again.

Error: Your query returned no results. Please change your search criteria and try again.

## Environment details

* Affected module version: 13.2.1+
* OS: macOS
* Terraform version: 0.14

btsuhako:
this module is searching for `amazon-eks-node-1.19-v*` in the AMI registry -> https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v13.2.1/local.tf#L18

right now it looks like AWS hasn't published that AMI yet. 1.18 available, but 1.19 isn't:
![image](https://user-images.githubusercontent.com/2069065/101959660-f7156200-3bc2-11eb-9c7c-03092b3e75f7.png)
![image](https://user-images.githubusercontent.com/2069065/101959671-fc72ac80-3bc2-11eb-8c85-3b58e11e09bc.png)

dbirks:
In case anyone else tries updating to 1.19 now that it's available:

![Screenshot from 2021-02-16 21-22-38](https://user-images.githubusercontent.com/7545665/108147687-54cf9a00-709d-11eb-8470-07a16b1f4cae.png)

[...]

Issue #1100, opened 2020-11-14
Prevent creation of windows ami datasource when we don't define windows worker nodes


Skullflow:
# I have issues

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?
I've just created an EKS cluster and I saw that the module is creating eks_worker_windows automatically.

Is there a way to prevent eks_worker_windows creation? I would like to keep my state as clean as possible.

Thanks.

barryib:
Humm... Sounds wired. The module create linux boxes by default.

It creates windows boxes only when `platform=windows`.

Please share your code, module version, terraform versions and plan output.

winelivery:
Sure,

module version: `13.1.0`

code:
```
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "13.1.0"

  cluster_name                    = var.cluster_name
  cluster_version                 = "1.18"
  cluster_create_timeout          = "1h"
  cluster_endpoint_private_access = true

  subnets = [
    module.vpc.k8s_subnet_eu_central_1a,
    module.vpc.k8s_subnet_eu_central_1b
  ]

  vpc_id = module.vpc.main_vpc_id

  node_groups = [
    {
      name                 = "default-node-group"
      instance_type        = "t2.medium"
      asg_desired_capacity = 1
      min_capacity         = 1
      max_capacity         = 3
      key_name             = ""
    }
  ]
}
```
plan:
```
Terraform will perform the following actions:

  # data.aws_eks_cluster.cluster will be read during apply
  # (config refers to values not yet known)
 <= data "aws_eks_cluster" "cluster"  {
      + arn                       = (known after apply)
      + certificate_authority     = (known after apply)
      + created_at                = (known after apply)
      + enabled_cluster_log_types = (known after apply)
      + endpoint                  = (known after apply)
      + id                        = (known after apply)
      + identity                  = (known after apply)
      + name                      = (known after apply)
      + platform_version          = (known after apply)
      + role_arn                  = (known after apply)
      + status                    = (known after apply)
      + tags                      = (known after apply)
      + version                   = (known after apply)
      + vpc_config                = (known after apply)
    }

  # data.aws_eks_cluster_auth.cluster will be read during apply
[...]

PR #1364, opened 2021-05-17
fix: Add fallback for missing Windows AMI

dbirks:
# PR o'clock

## Description

Currently if you try to upgrade to 1.20, you get a error that the Windows AMI can't be found, even if you're just using Linux nodes:

```
╷
│ Error: Your query returned no results. Please change your search criteria and try again.
│ 
│   on .terraform/modules/eks/data.tf line 27, in data "aws_ami" "eks_worker_windows":
│   27: data "aws_ami" "eks_worker_windows" {
│ 
╵
```

This PR adds a fallback so that when the Windows AMI can't be found, the most recent one is used instead.

This fixes issues like:
- #1100 
- #1141 
- #1246
- #1247 

I tested that this PR works by setting `cluster_version = 1.20` and `platform = "windows"`, and looking at the AMI in the plan diff:

Screenshot from 2021-05-17 17-15-55

Confirmed that the `ami-03be3e9ae875d10ad` AMI was for Windows 1.19:

Screenshot from 2021-05-17 17-18-42

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -27,7 +27,7 @@ data "aws_ami" "eks_worker" {
 data "aws_ami" "eks_worker_windows" {
   filter {
     name   = "name"
-    values = [local.worker_ami_name_filter_windows]
+    values = [local.worker_ami_name_filter_windows, "Windows_Server-2019-English-Core-EKS_Optimized-*"]
   }
 
   filter {


barryib:
What is the difference between your change and this https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/local.tf#L21 ? It sounds like the module actually fallback on

```hcl
"Windows_Server-2019-English-Core-EKS_Optimized-${tonumber(var.cluster_version) >= 1.14 ? var.cluster_version : 1.14}-*"
```

When `var. worker_ami_name_filter_windows` is empty.

barryib:
@dbirks  Thanks for working on this. I'm closing your PR in favor of https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1371 which is more generics and will let user define their filter if they really need to. Because we can't handle defaults for every usecases.

Feel free to re-open it if needed.

dbirks:
Thanks, your PR looks good :+1: 

That 1.14 fallback will only happen if the cluster version is less than 1.14 right? That reminds me that I should make a PR to remove that, since 1.14 isn't supported by amazon anymore.

Issue #1365, opened 2021-05-18
Instance Refresh not triggering on launch_template changes


Patricol:
## Description

With the new instance refresh functionality, changes to the launch template will not trigger a refresh, because the `autoscaling_group`'s `launch_template` block does not actually change if its `version = $Latest`, as is the case by [default](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/acbf51c33342ce2b1d6b05f55ee9e149e7c26621/local.tf#L78).

This is unexpected based on the comment [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/acbf51c33342ce2b1d6b05f55ee9e149e7c26621/local.tf#L102): `A refresh will always be triggered by a change in any of launch_configuration, launch_template, or mixed_instances_policy.`

I tested by changing `ami_id` multiple times; which never triggered an instance refresh.

All of the terraform resources appear to be created correctly; and I'm on the newest AWS provider (`3.40.0`).

I think this was just a misunderstanding of the intended behavior of `aws_autoscaling_group`'s `instance_refresh` block.

I suppose this module could handle bumping an explicitly set `launch_template` `version`.

## Versions

 - Terraform: `v0.14.11`

 - Provider(s):
```
+ provider registry.terraform.io/hashicorp/aws v3.40.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/external v2.1.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.2.0
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/terraform-aws-modules/http v2.3.0
```
 - Module: `16.0.0`

## Reproduction
Skipping most repro descriptions because the module does correctly setup all the resources; they just don't behave as documented.

barryib:
@Patricol Thanks for opening this. @bashims can you please have a look ?

barryib:
@Patricol did you tried the [instace refresh's example](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/instance_refresh) ? Is that example works for you ? Can you please share your configuration ?

bashims:
@barryib absolutely.

[...]

Issue #1315, opened 2021-04-21
Allow to specify update_default_version in Launch Templates


drmaciej:
## Is your request related to a new offering from AWS?

Not a very new feature; available in terraform: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#update_default_version

## Is your request related to a problem? Please describe.

When I eyeball the launch template in AWS Console, I have to remember to select the latest version from the dropdown. If I don't, I look at a wrong version and focus on incorrect values, which leads me to troubleshooting an issue which may not be there.

## Describe the solution you'd like.

There should be a variable that allows module users to specify the value of update_default_version. To maintain backwards compatibility, it could default to `null`; `null` is what's currently set in the state file.

## Describe alternatives you've considered.

None.

barryib:
Closing because it's now implemented in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1370

Issue #1293, opened 2021-04-08
How to inject a launch template version for worker_groups_launch_template


alexmnyc:
Our IAM policy prohibits creation launch templates with $Default and $Latest versions. It seems that the default behavior is to use "$Latest"

Using worker_groups_launch_template, is there any way to overcome this requirement and make the module auto-generate launch template version on each run?

barryib:
Closing because it's now implemented in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1370

PR #1370, opened 2021-05-19
fix: set an ASG's launch template version to an explicit version to automatically trigger instance refresh

bashims:
Set an ASG's launch template version to an explicit version automatically. This will ensure that an instance refresh will be
triggered whenever the launch template changes.

# PR o'clock

## Description

In order to properly trigger instance refresh whenever a launch template is modified the version of the template must be updated in the autoscaling group. Previous versions of this module only supported setting the template version manually. This PR adds a new feature that will either set the version to the latest (real) version or the default (real) version with the use of the boolean `use_latest_version`.  It is still possible to pin a specific  version of template by setting `launch_template_version` in `worker_groups_launch_template`. The default `launch_template_version` is ignored in the case of autoscaling group launch template, however it is preserved for the `node_group` case.

This PR also adds the ability to optionally update the default version for any autoscaling group launch template.

Fixes: #1365 #1315 #1293

### Checklist

- [x] CI tests are passing

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -75,7 +75,9 @@ locals {
     root_block_device_name               = data.aws_ami.eks_worker.root_device_name # Root device name for workers. If non is provided, will assume default AMI was used.
     root_kms_key_id                      = ""                                       # The KMS key to use when encrypting the root storage device
     launch_template_id                   = null                                     # The id of the launch template used for managed node_groups
-    launch_template_version              = "$Latest"                                # The lastest version of the launch template to use in the autoscaling group
+    launch_template_version              = "$Latest"                                # The latest version of the launch template to use in the autoscaling and node groups. Must be explicitly set in `worker_groups_launch_template`.
+    use_latest_version                   = true                                     # Set the autoscaling group to use the latest version of the launch template, otherwise the default template version will be used. Ignored when `launch_template_version` is set in `worker_groups_launch_template`.
+    update_default_version               = false                                    # Update the autoscaling group launch template's default version upon each update
     launch_template_placement_tenancy    = "default"                                # The placement tenancy for instances
     launch_template_placement_group      = null                                     # The name of the placement group into which to launch the instances, if any.
     root_encrypted                       = false                                    # Whether the volume should be encrypted or not
--- a/local.tf
+++ b/local.tf
@@ -75,8 +75,7 @@ locals {
     root_block_device_name               = data.aws_ami.eks_worker.root_device_name # Root device name for workers. If non is provided, will assume default AMI was used.
     root_kms_key_id                      = ""                                       # The KMS key to use when encrypting the root storage device
     launch_template_id                   = null                                     # The id of the launch template used for managed node_groups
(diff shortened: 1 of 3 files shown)


barryib:
@bashims can you please rebase your branch from master.

bashims:
@barryib rebased

local.tf
```diff
@@ -75,7 +75,9 @@ locals {
     root_block_device_name               = data.aws_ami.eks_worker.root_device_name # Root device name for workers. If non is provided, will assume default AMI was used.
     root_kms_key_id                      = ""                                       # The KMS key to use when encrypting the root storage device
     launch_template_id                   = null                                     # The id of the launch template used for managed node_groups
-    launch_template_version              = "$Latest"                                # The lastest version of the launch template to use in the autoscaling group
+    launch_template_version              = "$Latest"                                # The latest version of the launch template to use in the autoscaling and node groups. Must be explicitly set in `worker_groups_launch_template`.
+    use_latest_version                   = true                                     # Set the autoscaling group to use the latest version of the launch template, otherwise the default template version will be used. Ignored when `launch_template_version` is set in `worker_groups_launch_template`.
```
barryib, line 79: Instead of creating this new local, can't we just check if `launch_template_version == "$Latest"` ?

workers_launch_template.tf
```diff
@@ -141,7 +141,13 @@ resource "aws_autoscaling_group" "workers_launch_template" {
...
             var.worker_groups_launch_template[count.index],
             "launch_template_version",
-            local.workers_group_defaults["launch_template_version"],
+            lookup(
+              var.worker_groups_launch_template[count.index],
+              "use_latest_version",
+              local.workers_group_defaults["use_latest_version"]
+            )
+            ? aws_launch_template.workers_launch_template.*.latest_version[count.index]
+            : aws_launch_template.workers_launch_template.*.default_version[count.index]
```
barryib, line 150:
```suggestion
            lookup(
              var.worker_groups_launch_template[count.index],
              "launch_template_version",
              local.workers_group_defaults["launch_template_version"]
            ) == "$Latest"
            ? aws_launch_template.workers_launch_template.*.latest_version[count.index]
            : aws_launch_template.workers_launch_template.*.default_version[count.index]
```

We always create LT for self-managed worker groups, so there are no reason to allow external version. This was first implemented for Managed Node Group, to let users provide external LT for MNG (only).

local.tf
bashims: Yeah, I think that would be okay as well. It makes the default a default again :)

barryib: ^^

barryib (approved):

barryib:
Thanks again @bashims for your great work.

bashims:
@barryib do you think you might be able to include this fix in a patch release today?

barryib:
> @barryib do you think you might be able to include this fix in a patch release today?

Yep. I'll do that tonight.

barryib:
[v16.1.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v16.1.0) is released.

PR #1373, opened 2021-05-19
docs: Use IRSA for Node Termination Handler IAM policy attachement in Instance Refresh example

stevehipwell:
# PR o'clock

## Description

I've updated the example for how to use instance refresh.

- Removed policy attachment to worker security group
- Added spot instance termination support
- Enable lifecycle for all worker groups
- Rename Terraform resources to make it clearer

### Checklist

- [ ] CI tests are passing

### DIFF
--- a/examples/instance_refresh/main.tf
+++ b/examples/instance_refresh/main.tf
@@ -1,3 +1,5 @@
+# Based on the official aws-node-termination-handler setup guide at https://github.com/aws/aws-node-termination-handler#infrastructure-setup
+
 provider "aws" {
   region = var.region
 }
@@ -50,7 +52,7 @@ module "vpc" {
   enable_dns_hostnames = true
 }
 
-data "aws_iam_policy_document" "node_term" {
+data "aws_iam_policy_document" "aws_node_termination_handler" {
   statement {
     effect = "Allow"
     actions = [
@@ -76,22 +78,17 @@ data "aws_iam_policy_document" "node_term" {
       "sqs:ReceiveMessage"
     ]
     resources = [
-      module.node_term_sqs.sqs_queue_arn
+      module.aws_node_termination_handler_sqs.sqs_queue_arn
     ]
   }
 }
 
-resource "aws_iam_policy" "node_term" {
-  name   = "node-term-${local.cluster_name}"
-  policy = data.aws_iam_policy_document.node_term.json
+resource "aws_iam_policy" "aws_node_termination_handler" {
+  name   = "${local.cluster_name}-aws-node-termination-handler"
+  policy = data.aws_iam_policy_document.aws_node_termination_handler.json
 }
 
-resource "aws_iam_role_policy_attachment" "node_term_policy" {
-  policy_arn = aws_iam_policy.node_term.arn
-  role       = module.eks.worker_iam_role_name
-}
-
-data "aws_iam_policy_document" "node_term_events" {
+data "aws_iam_policy_document" "aws_node_termination_handler_events" {
   statement {
     effect = "Allow"
     principals {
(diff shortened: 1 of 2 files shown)


barryib (approved):
Awesome ❤️ ! Thanks @stevehipwell for your contribution.

bashims:
@stevehipwell  I believe the example was already using IRSA according to aws-node-termination-handler's logs:

```
│ 2021/05/19 01:53:08 DBG AWS Credentials retrieved from provider: WebIdentityCredentials                                                             │
│ 2021/05/19 01:53:08 INF Started watching for interruption events                                                                                    │
│ 2021/05/19 01:53:08 INF Kubernetes AWS Node Termination Handler has started successfully!                                                             
```

Thanks for removing the unnecessary policy attachment!

stevehipwell:
@bashims it was but the policy was also attached to the worker SG, I assumed it was a typo. Thanks for the great work on getting the instance refresh integrated and released.

bashims:
This change seems to break the `instance_refresh` example:

```

  on main.tf line 220, in resource "aws_autoscaling_lifecycle_hook" "aws_node_termination_handler":
 220:   for_each = toset(module.eks.workers_asg_names)

The "for_each" value depends on resource attributes that cannot be determined
resources that the for_each depends on.
```

examples/instance_refresh/main.tf
```diff
@@ -173,29 +192,35 @@ resource "helm_release" "anth" {
...
 }
 
 # Creating the lifecycle-hook outside of the ASG resource's `initial_lifecycle_hook`
 # ensures that node termination does not require the lifecycle action to be completed,
 # and thus allows the ASG to be destroyed cleanly.
-resource "aws_autoscaling_lifecycle_hook" "node_term" {
-  name                   = "node_term-${local.cluster_name}"
-  autoscaling_group_name = module.eks.workers_asg_names[0]
+resource "aws_autoscaling_lifecycle_hook" "aws_node_termination_handler" {
+  for_each = toset(module.eks.workers_asg_names)
```
bashims, line 220: I am not sure, but maybe adding the termination handler to each ASG might cause some delays in node draining if the ASG is not properly tagged for ANTH.

barryib: This is because for_each needs to know keys during the refresh staage. That means, we can't use for_each here, but `count` will do the trick.

bashims: The following seems to resolve the issue:

I can commit it under #1370 

```
diff --git a/examples/instance_refresh/main.tf b/examples/instance_refresh/main.tf
index 7c60621..1b87d8e 100644
--- a/examples/instance_refresh/main.tf
+++ b/examples/instance_refresh/main.tf
@@ -217,10 +217,9 @@ resource "helm_release" "aws_node_termination_handler" {
 # ensures that node termination does not require the lifecycle action to be completed,
 # and thus allows the ASG to be destroyed cleanly.
 resource "aws_autoscaling_lifecycle_hook" "aws_node_termination_handler" {
-  for_each = toset(module.eks.workers_asg_names)
-
+  count                  = length(module.eks.workers_asg_names)
   name                   = "aws-node-termination-handler"
-  autoscaling_group_name = each.value
+  autoscaling_group_name = module.eks.workers_asg_names[count.index]
   lifecycle_transition   = "autoscaling:EC2_INSTANCE_TERMINATING"
   heartbeat_timeout      = 300
   default_result         = "CONTINUE"
```

stevehipwell: Sorry guys, our implementation is a module wrapping the `terraform-aws-eks` module and containing sub modules for doing things like installing the `aws-node-termination-handler` and I copied our config out and flattened it.

andronux: Hello, I'm trying to get the asg name and arns output of my node_group with `module.eks.workers_asg_names` but the output is empty.

Am I missing something when using node_groups (not self-managed nodes) ?

barryib: `module.eks.workers_asg_names` is for self-managed node group only. If you're using managed node group,  there is `node_groups` output.

I don't think you need node termination handler if you're using managed node group. AWS handle node draining for you. Please see AWS docs for managed node groups.

andronux: @barryib even when using spot instances?

stevehipwell: @andronux spot termination for managed node groups should be automated.

andronux: ok thank you @stevehipwell @barryib !

barryib: To me, you don't need it for MNG. AWS handle drains for you. Please see:
- https://aws.amazon.com/fr/about-aws/whats-new/2020/12/amazon-eks-support-ec2-spot-instances-managed-node-groups/
- https://aws.amazon.com/fr/blogs/containers/amazon-eks-now-supports-provisioning-and-managing-ec2-spot-instances-in-managed-node-groups/ (spot section)
- https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html (spot section)

Or I'm probably missing something.

PR #1404, opened 2021-05-28
fix: Default `root_volume_type` must be `gp2`

barryib:
# PR o'clock

## Description

### Checklist

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -46,7 +46,7 @@ locals {
     spot_price                        = ""                          # Cost of spot instance.
     placement_tenancy                 = ""                          # The tenancy of the instance. Valid values are "default" or "dedicated".
     root_volume_size                  = "100"                       # root volume size of workers instances.
-    root_volume_type                  = "gp3"                       # root volume type of workers instances, can be "standard", "gp3", "gp2", or "io1"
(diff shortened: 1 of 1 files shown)


nauxliu:
Just curious, why the default must be gp2?

barryib:
> Just curious, why the default must be gp2?

If I remember, it's because it doest support for launch configuration. I'll with gp3 when we'll drop LC support (soon).

nauxliu:
> > Just curious, why the default must be gp2?
> 
> 
> 
> If I remember, it's because launch configuration doesn’t support it. I'll with gp3 when we'll drop LC support (soon).

Thanks for the explanation

yalimpop:
On a global scale this commit causes AWS billing to increase for everyone using this module :)

ivandir:
@barryib Is there plan to support Launch Template instead? See https://forums.aws.amazon.com/thread.jspa?messageID=969349

ivandir:
I agree with @yalimpop this module will become undesirable in the future with only GP2 support.

bryantbiggs:
@ivandir this issue is quite old and the latest version of the module provides support for any EBS type desired

Issue #1416, opened 2021-05-31
Make tags on security groups consistent


IvanDechovsky:
## Is your request related to a new offering from AWS?
No
## Is your request related to a problem? Please describe.
Tags for cluster and worker security groups mismatch. 

Cluster SG only has 
```
  tags = merge(
    var.tags,
    {
      "Name" = "${var.cluster_name}-eks_cluster_sg"
    },
  )
```
while workers SG has:
```
  tags = merge(
    var.tags,
    {
      "Name"                                      = "${var.cluster_name}-eks_worker_sg"
      "kubernetes.io/cluster/${var.cluster_name}" = "owned"
    },
  )
```
## Describe the solution you'd like.
Could you please add the extra `"kubernetes.io/cluster/${var.cluster_name}"` tag to the cluster security group so we can consistently query all security groups created by the module for a specific cluster in a given account.

barryib:
Thanks @IvanDechovsky. Right ! Feel free to open a PR. I'll happy to review it.

IvanDechovsky:
Done! [PR 1418](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1418)

daroga0002:
basing on [docs](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) and issue #1441 this should be closed.

PR #1418, opened 2021-05-31
feat: Add `kubernetes.io/cluster/<cluster_name>="owned"` tag to the cluster security group

IvanDechovsky:
…t with worker group

# PR o'clock

## Description

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1416

### Checklist

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -60,7 +60,8 @@ resource "aws_security_group" "cluster" {
   tags = merge(
     var.tags,
     {
-      "Name" = "${var.cluster_name}-eks_cluster_sg"
+      "Name"                                      = "${var.cluster_name}-eks_cluster_sg"
+      "kubernetes.io/cluster/${var.cluster_name}" = "owned"
     },
   )
 }


barryib:
After digging a bit in AWS docs, I found this:

image

It mention "_**One, and only one**, of the security groups associated to your nodes should have the following tag applied._". I don't know if we're not breaking something here.

marcelobartsch:
Issue #1441

daroga0002:
basing on [docs](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) and issue #1441 I think this PR should be closed. I checked and this security group `aws_security_group.cluster` is not assigned directly to workers (only API control plane), but this that never-less we shouldn`t follow practice which is against docs.

@antonbabenko what you think?

antonbabenko:
@daroga0002 I don't have enough understanding of the implications (docs are too long) so I would close this one.

daroga0002:
Closing as proposed approach doesn't follow AWS documentation and EKS requirements.

PR #1419, opened 2021-06-01
feat: Allow to choose launch template version for Managed Node Groups when `create_launch_template` is set to `true`

ArchiFleKs:
# PR o'clock

## Description

By default choose latest LT version. This trigger a rollout of all the instances. I noticed this because of the description change [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/6ad44056e3e18aacb64cbdfe8f3de0b6ec0c595b/modules/node_groups/launch_template.tf#L29)

This PR allow to override launch template version to postpone instances rollout if needed

### Checklist


barryib:
😢 

In docs https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html it says:

> When you update your node group to a different version of your launch template, all of the nodes in the group are recycled to match the new configuration of the specified launch template version. Existing node groups that do not use a custom launch template can't be updated directly. **Rather, you must create a new node group with a custom launch template to do so**.

ArchiFleKs:
> 😢
> 
> In docs https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html it says:
> 
> > When you update your node group to a different version of your launch template, all of the nodes in the group are recycled to match the new configuration of the specified launch template version. Existing node groups that do not use a custom launch template can't be updated directly. **Rather, you must create a new node group with a custom launch template to do so**.

@barryib Do you see any issue with that ? 

For me when not using Create/Custom LT the only thing that is going to trigger an update if changing the label maybe and the version right ?

modules/node_groups/node_groups.tf
```diff
@@ -48,7 +48,7 @@ resource "aws_eks_node_group" "workers" {
   dynamic "launch_template" {
     for_each = each.value["launch_template_id"] == null && each.value["create_launch_template"] ? [{
       id      = aws_launch_template.workers[each.key].id
-      version = aws_launch_template.workers[each.key].latest_version
+      version = each.value["launch_template_version"] != null ? each.value["launch_template_version"] : aws_launch_template.workers[each.key].latest_version
```
barryib, line 51: I was just wondering how user will get the `launch_template_version` if it's created directly by the module ? They have to go through the aws cli or aws console ?

barryib:
> @barryib Do you see any issue with that ?

I was just saying that it is intended behavior ^^.

> For me when not using Create/Custom LT the only thing that is going to trigger an update if changing the label maybe and the version right ?

I think you're right. I don't use MNG at all (for now).

modules/node_groups/node_groups.tf
ArchiFleKs: Yes I think they can see it in the TF or in cli/console. Most of the time if think you want to upgrade (that's one of the perk of MNG vs Self managed without instance refresh) but it might be useful when you have large pool and do not want to trigger an upgrade yet

barryib: I just noticed that `launch_template_version = "$Latest"` by default https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/local.tf#L104. In that case, should we need this ternary evaluation ?

Does `version = each.value["launch_template_version"]` enough ?

ArchiFleKs: Yes it should be, fixed

barryib (approved):

barryib:
Thanks @ArchiFleKs for your contribution.

barryib:

modules/node_groups/node_groups.tf
jaimehrubiks: @ArchiFleKs @barryib 
Can we test this? Somehow "$Latest" is converted into the integer latest automatically in my experiments

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1446

PR #1359, opened 2021-05-17
feat: Add timeout of 5mn when waiting for cluster

barryib:
# PR o'clock

## Description

Add timeout of 5mn when waiting for cluster As the previous way had.

### Checklist

- [x] CI tests are passing
- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #1420, opened 2021-06-01
feat: Add `var.wait_for_cluster_timeout` to allow configuring the wait for cluster timeout

### LINKED ISSUE #1395: Timeout on wait_for_cluster is breaking our terraform code
## Description

The PR #1359 is breaking out installation as it takes more than 5 minutes for the kubernetes api service to be provisioned. 

Either this number should be considerably large or should be configurable. 

I am not sure if this number can be configure. If yes, it is not well documented. Could some help us out please?

## Versions
 - Terraform: v0.14.4
 - Module: eks

### Code Snippet to Reproduce
```

 connect: connection timed out

on .terraform/modules/eks/cluster.tf line 68, in data "http" "wait_for_cluster":
68: data "http" "wait_for_cluster" {
```

The endpoint that wait_for_cluster is trying to wget is accessible outside cluster. I did wget and verify that.

## Expected behavior
cluster gets provisioned

## Actual behavior
cluster does not get provisioned but kubeconfig is acquired

SNA-rh:
# PR o'clock

## Description

During #1359 a default timeout of 300 seconds was set for this.  

We create hundreds of clusters a day and have seen many clusters take longer than 300 seconds.   Our team previously used the `wait_for_cluster_cmd` (**NOTE**: This is no longer available in this module) variable to wait 600 seconds.  Sometimes, even 5 minutes isn't even enough.  This PR will make this configurable, so our clusters are more likely to come up by the time we need to apply aws auth.

Resolves #1395 

### Checklist


variables.tf
```diff
@@ -387,3 +387,8 @@ variable "workers_egress_cidrs" {
   type        = list(string)
   default     = ["0.0.0.0/0"]
 }
+
+variable "wait_for_cluster_create_timeout" {
```
barryib, line 391:
```suggestion
variable "wait_for_cluster_timeout" {
```

barryib (commented):
Thanks for working on this. Just variable rewording, otherwise it's LGTM.

variables.tf
```diff
@@ -387,3 +387,8 @@ variable "workers_egress_cidrs" {
   type        = list(string)
   default     = ["0.0.0.0/0"]
 }
+
+variable "wait_for_cluster_timeout" {
+  description = "A timeout of 5mn when waiting for cluster to be available."
```
barryib, line 392: Can you please specify that the timeout is in second ?

variables.tf
```diff
@@ -387,3 +387,8 @@ variable "workers_egress_cidrs" {
...
+  default     = 300
```
barryib, line 393: Shouldn't we add `type = number` ?

barryib (changes requested):
On last thing to clarify please ^^.

SNA-rh:
@barryib I updated the code. Let me know if it looks better now.

I was not sure of the process, so I resolved the comment issues myself.  I hope that's OK.

barryib (approved):

barryib:
Nice. Thanks @SNA-rh for you contribution.

jonDowdle:
@barryib We're real excited to use this. Any idea when the next release gets cut that'd include it?

barryib:

PR #1421, opened 2021-06-02
fix: Add dash suffix to name_prefix

imranismail:
## Description

This adds a dash suffix to the name_prefix

### Before
```
prod-eks-od20210602013534947400000001
```

### After

```
prod-eks-od-20210602013534947400000001
```

### Checklist

### DIFF
--- a/modules/node_groups/locals.tf
+++ b/modules/node_groups/locals.tf
@@ -32,7 +32,7 @@ locals {
     lookup(
       v,
       "name_prefix",
-      join("-", [var.cluster_name, k])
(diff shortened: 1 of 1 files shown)


daroga0002:
this will be breaking change probably, @imranismail  can you test what will happen after changing this on transition?

barryib:
There is also the `name_prefix` argument you can use for managed node groups. You can add dash there and whatever you want.

imranismail:
@barryib yup, figured so. Okay, I'll use that alternative. But I think this can still be plan to be include this in the next major release, right?

PR #1429, opened 2021-06-07
feat: create the fargate role even if number of fargate profiles is 0

### LINKED ISSUE #1356: Always create fargate role

No

When manually managing aws_auth configmap, you need a manual process to be able to add the fargate role.
However, when you do not specify any fargate profiles, the role is not created: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/781f673295dd69d24c85d10b47cb03507f394402/modules/fargate/locals.tf#L2
This means that the fargate role gets recreated if you have an intermediate state of no active profiles; and hence (because of using `name_prefix` rather than a specific `name`) needing to update your manually managed aws_auth configmap.

Remove the `&& length(var.fargate_profiles) > 0` from the condition of when to create the fargate IAM role

An additional variable to the fargate submodule `always_create_role` -> this seems like needless complexity, though if its less objectionable then that would be fine.

james-callahan:
# PR o'clock

## Description

Closes #1356

### DIFF
--- a/modules/fargate/data.tf
+++ b/modules/fargate/data.tf
@@ -1,5 +1,5 @@
 data "aws_iam_policy_document" "eks_fargate_pod_assume_role" {
-  count = local.create_eks && var.create_fargate_pod_execution_role ? 1 : 0
+  count = var.create_eks && var.create_fargate_pod_execution_role ? 1 : 0
   statement {
     effect  = "Allow"
     actions = ["sts:AssumeRole"]
@@ -12,6 +12,6 @@ data "aws_iam_policy_document" "eks_fargate_pod_assume_role" {
 }
 
 data "aws_iam_role" "custom_fargate_iam_role" {
-  count = local.create_eks && !var.create_fargate_pod_execution_role ? 1 : 0
+  count = var.create_eks && !var.create_fargate_pod_execution_role && var.fargate_pod_execution_role_name != null ? 1 : 0
   name  = var.fargate_pod_execution_role_name
 }
--- a/modules/fargate/fargate.tf
+++ b/modules/fargate/fargate.tf
@@ -1,5 +1,5 @@
 resource "aws_iam_role" "eks_fargate_pod" {
-  count                = local.create_eks && var.create_fargate_pod_execution_role ? 1 : 0
+  count                = var.create_eks && var.create_fargate_pod_execution_role ? 1 : 0
   name_prefix          = format("%s-fargate", substr(var.cluster_name, 0, 24))
   assume_role_policy   = data.aws_iam_policy_document.eks_fargate_pod_assume_role[0].json
   permissions_boundary = var.permissions_boundary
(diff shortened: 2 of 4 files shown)


daroga0002:
I dont think this is best approach from security point of view to create it always (even if you dont need it) as this is causing a lot of security questions

james-callahan:
> as this is causing a lot of security questions

From whom?


We could potentially make it so you can set `create_fargate_pod_execution_role` to `false` without providing the custom role, as long as you don't have any fargate profiles.

daroga0002:
> From whom?

from internal security teams. Questions which will be aksed:
- what is purpose of that role?
- how that role is used?
- if not used why you require it?
- why you cannot remove it?

BeyondEvil:
> We could potentially make it so you can set `create_fargate_pod_execution_role` to `false` without providing the custom role, as long as you don't have any fargate profiles.

Yes, please! 🙏

james-callahan:
> Yes, please! pray

Done.

daroga0002:
I still think this is scenario which will cause issues for multiple users using this module. 


Scenario which you pasted here and in issue #1356 is very specific for you (and maybe few other users). 

Currently I am not using Fargate under EKS at all and running your branch producing me:
```
......
  # module.eks_cluster.module.fargate.aws_iam_role.eks_fargate_pod[0] will be created
  + resource "aws_iam_role" "eks_fargate_pod" {
      + arn                   = (known after apply)
      + assume_role_policy    = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "sts:AssumeRole"
                      + Effect    = "Allow"
                      + Principal = {
                          + Service = "eks-fargate-pods.amazonaws.com"
                        }
                      + Sid       = ""
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
      + create_date           = (known after apply)
      + force_detach_policies = false
      + id                    = (known after apply)
      + managed_policy_arns   = (known after apply)
      + max_session_duration  = 3600
      + name                  = (known after apply)
      + name_prefix           = "eks-tools-fargate"
      + path                  = "/"
      + tags                  = {
          + "Name" = "eks-tools"
        }
      + tags_all              = {
          + "Name"    = "eks-tools"
          + "project" = "scp"
          + "source"  = "tf-scp-tools"
        }
      + unique_id             = (known after apply)

      + inline_policy {
          + name   = (known after apply)
          + policy = (known after apply)
        }
    }

  # module.eks_cluster.module.fargate.aws_iam_role_policy_attachment.eks_fargate_pod[0] will be created
  + resource "aws_iam_role_policy_attachment" "eks_fargate_pod" {
      + id         = (known after apply)
      + policy_arn = "arn:aws:iam::aws:policy/AmazonEKSFargatePodExecutionRolePolicy"
      + role       = (known after apply)
    }
......
```

What generally violating AWS/EKS best practices to follow least privileges approach (in result not create a roles/policies which are not required).


I still dont understand why you trying to generate those roles inside module instead generate them outside (or create company specific module) as in case when you manage `aws-auth` configmap outside module there shouldnt be difference for you.

daroga0002:
@bryantbiggs  do you think this should be considered?

bryantbiggs:
no - we should only create the role if:
1. the module relies on the role and will fail without it
2. users explicitly opt in to creating the role, usually through the use of a variable like `create_this_role`

this looks like like a poor approach because the cluster is being managed through a mixed state of manual and code, and to compensate we are just adding a role for all users whether they need it or not. my recommendation would be to create the role you need separately from the module, or manage auth through the module and let the module create the role for you. this is a very obscure edge case as far as I can tell and we should close this issue/PR

PR #1424, opened 2021-06-02
feat: Add support for Managed Node Groups (`node_groups`) taints

nauxliu:
## Description

Add support for set taints on managed node groups, Resolve #1403

Ref:
https://docs.aws.amazon.com/eks/latest/userguide/node-taints-managed-node-groups.html
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#taint-configuration-block

### Checklist

- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #1433, opened 2021-06-08
feat: Improve managed node group bootstrap

stevehipwell:
# PR o'clock

## Description

The managed node group bootstrap pattern has been modified to use the environment variables already defined in the `bootstrap.sh` script sourced from a new `/etc/profile.d/bootstrap.sh` file. The script still needs to be modified to use `bootstrap.sh` but this pattern can support additional bootstrap options not currently directly supported and other patterns (such as dynamic labels from instance metadata).

I've enabled support for managed node group `--use-max-pods` via `use_max_pods`. This can already be set via `extra_bootstrap_args` for un-managed node groups.

### Checklist


stevehipwell:
@ArchiFleKs

ArchiFleKs:
@stevehipwell I've never tried it this way, but it seems way cleaner :)  Did you manage to test the extra args also (like taint etc) ?

stevehipwell:
@ArchiFleKs I think native taint support has been merged ready to be released, which other arguments were you thinking of? I've tested an additional `pre_userdata` that sets the subnet-id in the `/etc/profile.d/bootstrap.sh` file and uses it in a custom label passed in to the module via `kubelet_extra_args = "--node-labels=subnet=$${BOOTSTRAP_SUBNET_ID}"`.

ArchiFleKs:
Yes you're right. Nice.

stevehipwell:
@barryib could you take a look this would be a good to release with https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1424.

stevehipwell:
@barryib could you take a look a this? The pattern can be re-used with the new [containerd support](https://github.com/awslabs/amazon-eks-ami/releases/tag/v20210716) (via `CONTAINER_RUNTIME="containerd"`) and I'd be happy to add this to the PR if you'd like.

stevehipwell:
@barryib I've added support for `--container-runtime` now as I'd like to use this. Could you take a look at this PR?

ArchiFleKs:
This PR is very useful to use new feature in VPC CNI 1.9

stevehipwell:
> This PR is very useful to use new max pod feature in VPC CNI 1.9

@ArchiFleKs that's one of the things I'm already using this pattern for, but it's not only useful for the new prefix behaviour but also the existing custom networking.

ArchiFleKs:
> > This PR is very useful to use new max pod feature in VPC CNI 1.9
> 
> @ArchiFleKs that's one of the things I'm already using this pattern for, but it's not only useful for the new prefix behaviour but also the existing custom networking.

Been using it for a couple of weeks with a fork on ARM micro with new network prefix and containerd, working like a charm, would love to see this merged

modules/node_groups/launch_template.tf
```diff
@@ -9,8 +9,10 @@ data "cloudinit_config" "workers_userdata" {
     content_type = "text/x-shellscript"
     content = templatefile("${path.module}/templates/userdata.sh.tpl",
```
bryantbiggs, line 10: I think this userdata keeps coming up often, maybe its time to stop adding more options here and just have a generic "bring your own userdata script" option. so either you can use the standard userdata script provided here which only supports `pre_userdata` and `kubelet_extra_args`, or you can provide your own userdata script that it will use instead

bryantbiggs: @daroga0002 thoughts?

stevehipwell: @bryantbiggs & @daroga0002 I disagree, this PR solves a lot of the need for regular changes. It allows important parameters to be captured as strongly typed variables (`container_runtime`, `use_max_pods` &`kubelet_extra_args`) and enables any additional userdata via `pre_userdata` which can now also interact with the AMI _boostrap.sh_ script by adding variables to the new _/etc/profile.d/bootstrap.sh_ file.

bryantbiggs: @stevehipwell it solves the here and now for yourself and maybe a handful of others, but it doesn't solve for tomorrow when someone else wants to add a new piece of functionality, and the next user, and the next ... etc. 

this approach is common amongst our modules - we try to provide a good, default "out of box" experience that will hopefully cover the 80% of use cases, but the number of possible combinations and customizations can explode very quickly to meet individual use cases. we see this most often with IAM roles and policies and we have used this pattern successfully - provide a baseline default IAM role/policy that makes provisioning the module easy and seamless, but also provide the ability to override the role/policy that the module uses with one provided by the user of the module

we use this pattern in this module already as well:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/local.tf#L3-L7

also, we are looking for ways to reduce the complexity of this module and abstract away some of these niche configurations. if we can do this, it allows for more focus on larger initiatives within the project and more agility

daroga0002: I am thinking and looking here 
https://github.com/awslabs/amazon-eks-ami/blob/548b1650e63f3ab582d79418c7b3336943642fd1/files/bootstrap.sh#L108-L118
 and see there are still options which are not handled currently. So imagine that somebody can come and ask about next parameters.

Maybe some better solution will be just 
move `pre_user_data` part before 
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/0ed8af1e987ca9e59f2b62acec9cfce75fa30e77/modules/node_groups/templates/userdata.sh.tpl#L11

what will allow users to apply any variables they will want.

So pre user data in case will be looking (optionally if users really want to have it):
```
# Define extra environment variables for bootstrap
printf '#!/bin/bash
export CONTAINER_RUNTIME="%s"
export USE_MAX_PODS="%s"
export KUBELET_EXTRA_ARGS="%s"
' "${container_runtime}" "${use_max_pods}" "${kubelet_extra_args}" > /etc/profile.d/bootstrap.sh
``` 

and then executing:
```
sed -i '/^set -o errexit/a\\nsource /etc/profile.d/bootstrap.sh' /etc/eks/bootstrap.sh
```

bryantbiggs: or have two - `pre_userdata` and `post_userdata` which wrap what exists

stevehipwell: @bryantbiggs the current design needs the user data to be templated to support some very common use cases and this PR extends this capability to some additional very common use cases. If you wish to change this you'd need to release a breaking change and significantly reduce capability for the 80% use case.

@daroga0002 variables need to be persisted or else they don't work between init steps, this is why I've designed the user data in this way for this PR.

stevehipwell: @bryantbiggs @daroga0002 the pattern in this PR simplifies both the implementation and maintenance of customising _/etc/eks/bootstrap.sh_. It is now easy to add new managed values, such as I've done here for `container_runtime` and `use_max_pods`, by updating the initial creation of _/etc/profile.d/eks-bootstrap.sh_. It's also easy for end users to set additional environment variables by appending to this file.

As there is nothing happening besides linking to _/etc/profile.d/eks-bootstrap.sh_ from _/etc/eks/bootstrap.sh_  before the `pre_userdata` is templated there is no reason to add a second variable. A power user could always clear _/etc/profile.d/eks-bootstrap.sh_ from `pre_userdata` so there is no reason to even need to provide a userdata override variable.

An example of what this PR offers end-users can be seen by the following snippet setting an environment variable in _/etc/eks/bootstrap.sh_.

```terraform
node_group = {
  pre_userdata = <<-EOT
    printf 'PAUSE_CONTAINER_VERSION="%s"
    ' "3.1-eksbuild.1"  >> /etc/profile.d/eks-bootstrap.sh
  EOT
}
```

stevehipwell:
@bryantbiggs @daroga0002 @antonbabenko could you take another look at this and the comments above? I really think that this PR makes the managed node group module more generic while supporting important parameters in a strongly typed way.

modules/node_groups/launch_template.tf
daroga0002: I really like sourcing `source /etc/profile.d/eks-bootstrap.sh` into bootstrap, but still I dont think that we dont need a seperate variables to customize bootstrap script.

Currently we have https://github.com/awslabs/amazon-eks-ami/blob/548b1650e63f3ab582d79418c7b3336943642fd1/files/bootstrap.sh#L108-L118 

so I am still pushing to make this more generic and allow to push any variable into bootstrap than create variables in TF code for each of them.

stevehipwell: @daroga0002 we already have a strongly typed variable for `KUBELET_EXTRA_ARGS` so for consistency the approach in this PR makes sense as the alternative would be to document multiple ways of setting `KUBELET_EXTRA_ARGS`; which after further thought I could do pretty easily. If we're happy with a breaking change then it's even simpler.

stevehipwell:
Closed in favour of https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1577.

PR #1434, opened 2021-06-09
fix: userdata_template_file in workers_group_defaults is now used if defined

DanielRis:
…] if defined

# PR o'clock

## Description

I am trying to create a company specific wrapper around this module to simplify usage and standardize a few things. We are using our own customized AMIs which work a little bit different then the default AMIs provided by AWS.

I wanted to define a default `userdata_template_file` that is used for all workers created via `worker_groups` without asking users to provide this every time.

I quickly realized that `workers_group_defaults` are not used for userdata in general. My PR tried to fix this behavior. It should be a non breaking change and only affect deployments which actually defined a default.

### Checklist

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -189,8 +189,8 @@ locals {
         var.worker_groups[index],
         "userdata_template_file",
         lookup(var.worker_groups[index], "platform", local.workers_group_defaults["platform"]) == "windows"
-        ? "${path.module}/templates/userdata_windows.tpl"
-        : "${path.module}/templates/userdata.sh.tpl"
+        ? coalesce(local.workers_group_defaults["userdata_template_file"], "${path.module}/templates/userdata_windows.tpl")
+        : coalesce(local.workers_group_defaults["userdata_template_file"], "${path.module}/templates/userdata.sh.tpl")
       ),
       merge({
         platform            = lookup(var.worker_groups[index], "platform", local.workers_group_defaults["platform"])


daroga0002:
does usage of `pre_userdata` doesnt fit into your use case?


I am trying to find here understand use case, as module providing in general minimal set of user_data which must be executed in EKS worker and allowing to add own pre/additional user data arguments

DanielRis:
Hi @daroga0002, I am using BottleRocket as AMI here which requires different user_data.

daroga0002:
can you paste here difference?

I know we have https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/bottlerocket/main.tf which is bottlerocket AMI so wondering what else is required

DanielRis:
I wanted to use simplified `worker_groups` as user provided inputs, and not `worker_groups_launch_template` for my inputs, and I was wondering why there is inconsistent in defaults handling for this.

This is my code now, works fine:

```
module "eks" {
  source  = "/home/dristic/terraform-public-modules/terraform-aws-eks"
  # version = "17.0.3"

  cluster_version = var.cluster_version
  subnets         = local.control_subnet_ids
  vpc_id          = local.vpc_id

  # Worker configuration
  workers_group_defaults = {
    userdata_template_file = "${path.module}/templates/userdata.toml"
    ami_id                 = data.aws_ami.bottlerocket_ami.id
    additional_security_group_ids = [
      module.gatherer.security_group_ids["outbound"],
      module.gatherer.security_group_ids["service"],
    ]
  }

  worker_groups = [
    {
      instance_type        = "t3.small"
      asg_desired_capacity = 1
    }
  ] 
}
```

daroga0002:
but what you got in `${path.module}/templates/userdata.toml` ?

DanielRis:
```
[settings.kubernetes]
api-server = "${endpoint}"
cluster-certificate = "${cluster_auth_base64}"
cluster-name = "${cluster_name}"

[settings.kernel]
lockdown = "integrity"

[settings.host-containers.admin]
enabled = false

[settings.host-containers.control]
enabled = false
```

daroga0002:
maybe better approach instead changing template will be allowing to replace whole content of user data. Then users will be able to make any template outside this module and then pass it filled into module.

The drawback can be that if users will be requiring to fill it with some data from EKS module then it can create cycle dependency.


Just loudly thinking …

PR #1443, opened 2021-06-15
feat: Add support for addons

tnimni:
# PR o'clock

## Description

Add support for eks addons recommended by was
1. VPC CNI
2. CoreDns
3. kube-proxy

The default behavior is to install the addons to the cluster.

Requires irsa to be enabled to deploy the VPC CNI addon

### Checklist

- [V] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

### DIFF
--- a/modules/addons/coredns.tf
+++ b/modules/addons/coredns.tf
@@ -1,0 +1,12 @@
+resource "aws_eks_addon" "coredns" {
+  count = var.create_coredns_addon ? 1 : 0
+
+  cluster_name      = var.cluster_name
+  addon_name        = "coredns"
+  resolve_conflicts = "OVERWRITE"
+  addon_version     = lookup(var.coredns_versions, var.cluster_version, "Not Found")
+
+  depends_on = [
+    var.eks_depends_on
+  ]
+}
--- a/modules/addons/outputs.tf
+++ b/modules/addons/outputs.tf
@@ -1,0 +1,5 @@
+# output "iam_role_arn" {
+#   description = "role arn"
+#   value       = module.iam_assumable_role_with_oidc.iam_role_arn
+# }
+
--- a/modules/addons/outputs.tf
+++ b/modules/addons/outputs.tf
@@ -1,5 +1,0 @@
-# output "iam_role_arn" {
-#   description = "role arn"
-#   value       = module.iam_assumable_role_with_oidc.iam_role_arn
-# }
-
--- a/modules/addons/variables.tf
+++ b/modules/addons/variables.tf
@@ -77,7 +77,7 @@ variable "cluster_oidc_issuer_url" {
 
 variable "enable_irsa" {
   type        = bool
-  description = "Whether to create iam role for vpc cni and attach it to the service account"
+  description = "Whether to create iam role for vpc cni and attach it to the service account, if irsa=false vpc cni plugin will not be deployed"
   default     = true
 }
 
--- a/modules/addons/variables.tf
+++ b/modules/addons/variables.tf
@@ -1,0 +1,83 @@
+variable "create_vpc_cni_addon" {
+  type        = bool
+  description = "Controls if vpc cni addon should be deployed"
+  default     = true
+}
+
+variable "create_kube_proxy_addon" {
+  type        = bool
+  description = "Controls if kube proxy addon should be deployed"
+  default     = true
+}
+
+variable "create_coredns_addon" {
+  type        = bool
+  description = "Controls if coredns addon should be deployed"
+  default     = true
+}
+
+variable "cluster_name" {
+  type        = string
+  description = "Name of parent cluster"
+}
+
+
+variable "cluster_version" {
+  type        = string
+  description = "Kubernetes version to use for the EKS cluster."
+}
+
+variable "coredns_versions" {
+  # Versions are taken from https://docs.aws.amazon.com/eks/latest/userguide/managing-coredns.html#updating-coredns-add-on
+  type        = map(any)
+  description = "The CoreDns plugin version for the corresponding version"
+  default = {
+    "1.18" = "v1.8.3-eksbuild.1"
+    "1.19" = "v1.8.3-eksbuild.1"
+    "1.20" = "v1.8.3-eksbuild.1"
+  }
+}
+
+
+variable "kube_proxy_versions" {
+  # Versions are taken from https://docs.aws.amazon.com/eks/latest/userguide/managing-kube-proxy.html#updating-kube-proxy-add-on
+  type        = map(any)
+  description = "The Kube proxy plugin version for the corresponding eks version"
+  default = {
+    "1.18" = "v1.18.8-eksbuild.1"
+    "1.19" = "v1.19.6-eksbuild.2"
+    "1.20" = "v1.20.4-eksbuild.2"
+  }
+}
+
+variable "vpc_cni_versions" {
+  # Versions are taken from https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html#updating-vpc-cni-add-on
+  # Latest patch version is taken from https://github.com/aws/amazon-vpc-cni-k8s
+  type        = map(any)
+  description = "The VPC CNI plugin version for the corresponding eks version"
+  default = {
+    "1.18" = "v1.7.10-eksbuild.1"
+    "1.19" = "v1.7.10-eksbuild.1"
+    "1.20" = "v1.7.10-eksbuild.1"
+  }
+}
+
+# Hack for a homemade `depends_on` https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2
+# Will be removed in Terraform 0.13 with the support of module's `depends_on` https://github.com/hashicorp/terraform/issues/10462
+variable "eks_depends_on" {
+  description = "List of references to other resources this submodule depends on."
+  type        = any
+  default     = null
+}
+
+variable "cluster_oidc_issuer_url" {
+  type        = string
+  description = "The cluster oidc issuer url"
+}
+
+variable "enable_irsa" {
+  type        = bool
+  description = "Whether to create iam role for vpc cni and attach it to the service account"
+  default     = true
+}
+
--- a/README.md
+++ b/README.md
@@ -152,17 +152,18 @@ Apache 2 Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraf
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 3.40.0 |
-| <a name="provider_http"></a> [http](#provider\_http) | >= 2.4.1 |
-| <a name="provider_kubernetes"></a> [kubernetes](#provider\_kubernetes) | >= 1.11.1 |
-| <a name="provider_local"></a> [local](#provider\_local) | >= 1.4 |
+| <a name="provider_aws"></a> [aws](#provider\_aws) | 3.45.0 |
+| <a name="provider_http"></a> [http](#provider\_http) | 2.4.1 |
+| <a name="provider_kubernetes"></a> [kubernetes](#provider\_kubernetes) | 2.3.2 |
+| <a name="provider_local"></a> [local](#provider\_local) | 2.1.0 |
 
 ## Modules
 
 | Name | Source | Version |
 |------|--------|---------|
-| <a name="module_fargate"></a> [fargate](#module\_fargate) | ./modules/fargate |  |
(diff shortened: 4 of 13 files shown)


modules/addons/variables.tf
```diff
@@ -0,0 +1,83 @@
...
+    "1.18" = "v1.7.10-eksbuild.1"
+    "1.19" = "v1.7.10-eksbuild.1"
+    "1.20" = "v1.7.10-eksbuild.1"
+  }
+}
+
+# Hack for a homemade `depends_on` https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2
+# Will be removed in Terraform 0.13 with the support of module's `depends_on` https://github.com/hashicorp/terraform/issues/10462
+variable "eks_depends_on" {
+  description = "List of references to other resources this submodule depends on."
```
daroga0002, line 68: maybe this should me moved to module level as terraform from some time allowing it?

daroga0002 (commented):
can we add some example into `examples` directory which will be using this?

modules/addons/variables.tf
```diff
@@ -0,0 +1,83 @@
...
+
+
+variable "cluster_version" {
+  type        = string
+  description = "Kubernetes version to use for the EKS cluster."
+}
+
+variable "coredns_versions" {
+  # Versions are taken from https://docs.aws.amazon.com/eks/latest/userguide/managing-coredns.html#updating-coredns-add-on
+  type        = map(any)
```
daroga0002, line 32: I think better will be `map(string)` just to not loose type over time

modules/addons/variables.tf
```diff
@@ -0,0 +1,83 @@
...
+    "1.18" = "v1.8.3-eksbuild.1"
+    "1.19" = "v1.8.3-eksbuild.1"
+    "1.20" = "v1.8.3-eksbuild.1"
+  }
+}
+
+
+variable "kube_proxy_versions" {
+  # Versions are taken from https://docs.aws.amazon.com/eks/latest/userguide/managing-kube-proxy.html#updating-kube-proxy-add-on
+  type        = map(any)
```
daroga0002, line 44: I think better will be `map(string)` just to not loose type over time

modules/addons/variables.tf
```diff
@@ -0,0 +1,83 @@
...
+    "1.18" = "v1.18.8-eksbuild.1"
+    "1.19" = "v1.19.6-eksbuild.2"
+    "1.20" = "v1.20.4-eksbuild.2"
+  }
+}
+
+variable "vpc_cni_versions" {
+  # Versions are taken from https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html#updating-vpc-cni-add-on
+  # Latest patch version is taken from https://github.com/aws/amazon-vpc-cni-k8s
+  type        = map(any)
```
daroga0002, line 56: I think better will be `map(string)` just to not loose type over time

modules/addons/variables.tf
```diff
@@ -0,0 +1,83 @@
...
+    "1.19" = "v1.7.10-eksbuild.1"
+    "1.20" = "v1.7.10-eksbuild.1"
+  }
+}
+
+# Hack for a homemade `depends_on` https://discuss.hashicorp.com/t/tips-howto-implement-module-depends-on-emulation/2305/2
+# Will be removed in Terraform 0.13 with the support of module's `depends_on` https://github.com/hashicorp/terraform/issues/10462
+variable "eks_depends_on" {
+  description = "List of references to other resources this submodule depends on."
+  type        = any
```
daroga0002, line 69: probably `string` will be better

modules/addons/outputs.tf
```diff
@@ -0,0 +1,14 @@
+output "vpc_cni_id" {
+  description = "The id of the Amazon VPC CNI addon"
+  value       = aws_eks_addon.vpc_cni[0].id
```
daroga0002, line 3: does it will be working when I will not create this addon (there will be no element 0)?

modules/addons/coredns.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_eks_addon" "coredns" {
+  count = var.create_coredns_addon ? 1 : 0
+
+  cluster_name      = var.cluster_name
+  addon_name        = "coredns"
+  resolve_conflicts = "OVERWRITE"
+  addon_version     = lookup(var.coredns_versions, var.cluster_version, "Not Found")
```
daroga0002, line 7: how AWS API will react when we pass `"Not Found"`, does there is validation on their end?

modules/addons/variables.tf
tnimni: This pattern is used in modules/fargate and modules/node_groups, I just copied it from there to keep everything consistent.
I'm not sure which example you require, as eks_depends_on is defined in addons.tf in the root of the main module, https://github.com/tnimni/terraform-aws-eks/blob/feature/addons/addons.tf#L11-L14

modules/addons/variables.tf
tnimni: Changed to `map(string)`

modules/addons/variables.tf
tnimni: Changed to `map(string)`

modules/addons/variables.tf
tnimni: Changed to `map(string)`

modules/addons/variables.tf
tnimni: based on https://github.com/tnimni/terraform-aws-eks/blob/feature/addons/modules/fargate/variables.tf#L59-L65 , I copied it to keep everything consistent

modules/addons/outputs.tf
tnimni: Fixed the count=0 for all addons

modules/addons/coredns.tf
tnimni: This is on purpose, it will not even reach aws api. terraform plan will throw an error stating the addon version can not be `"Not Found"` giving a hint for the user that the cluster version is not supported by the module

modules/addons/variables.tf
daroga0002: yup, this was created before modules were supporting dependncies. Lets maintainers decide does we want to leave this for consistency or we will start using module dependency.

daroga0002 (commented):
for me it looks good, @barryib  can you take a look on it?

stevehipwell:
Please can the defaults not be to enable these.

tnimni:
@stevehipwell 
changed defaults to false

tnimni:
Get we llease get this approved? @barryib

thecosmicfrog:
Good work here @tnimni - hoping for it to be merged soon.

tnimni:
> Good work here @tnimni - hoping for it to be merged soon.

Thanknyo

> Good work here @tnimni - hoping for it to be merged soon.

Thank you

mmerickel:
So if we enable the vpc cni addon then should we pass `attach_worker_cni_policy=false`? Would be good to mention that in the docs if so.

tnimni:
> attach_worker_cni_policy=false

Hi,
it is not required to do so,

I checked and aws recommend it here https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html

I added to the example and also updated the main module readme to reflect that the variable attach_worker_cni_policy should be set to false if deploying the vpc cni addon

I appreciate your input

nikitacr7:
Hi

Awesome. We're waiting for it.

thapakazi:
Same Same, nice work @tnimni  :clap:

tnimni:
@nikitacr7 @thapakazi thank you both

csschwe:
@tnimni This is great!!!

I do have one request though.  Can you add support for tags like the node_groups module

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L9

Thanks

tnimni:
> @tnimni This is great!!!
> 
> I do have one request though. Can you add support for tags like the node_groups module
> 
> https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L9
> 
> Thanks

I'll have a look into it

wellermann:
Agreed, this is great.

EKS 1.21 just came out this week so it might be worth adding it now.

Either way this is great work @tnimni Thank you for doing this.

aftouh:
Awesome. I'm waiting for it.
Thank you @tnimni.

kuntalkumarbasu:
You just save a ton of my time. thank you @tnimni .
When can we get it merged?

tnimni:
> You just save a ton of my time. thank you @tnimni .
> When can we get it merged?

Hi,
I'm not able to merge or approve
need to ask @barryib

tnimni:
> EKS 1.21 just came out this week so it might be worth adding it now.
> 
> Either way this is great work @tnimni Thank you for doing this.

Hi,
I added it, also updated older cluster versions with the latest recommended addons

stevehipwell:
@tnimni I've just reviewed this and the whole addons module should be conditional, not the resources inside it.

stevehipwell:
@tnimni my second observation is that this would introduce a significant maintenance overhead to the module for it to function as expected without user interaction. Every time a new addon version is released you'd need to update the module addon version maps or else the paved road for users would be to get old versions. You'd also need to change the module for each new EKS version. The code for this module hasn't been modified for about 2 months at this point so I really don't think that the current design works.

If the user is providing the addon versions they have to understand the addon release cycle by which point they would be better off coding it themselves; as is this pattern adds more overhead as you need to know the state of the module as well as the state of the addon releases.

There is the [aws eks describe-addon-versions](https://docs.aws.amazon.com/cli/latest/reference/eks/describe-addon-versions.html) command to get the available addon versions for a specific EKS version, meaning that the Terraform provider should be able to support this as a data source. I'd suggest holding off on this PR until this is available as it'd solve the above concerns.

bryantbiggs:
hi all 👋🏼 - this looks interesting, but also a bit scary. the EKS module is already quite large and complex, we've even discussed how best to move forward since it is becoming so large and complex. I don't know if adding this functionality onto the module directly is the best course of action - this seems to be more of an independent module or a sub-module that is standalone and not tied into the root of the overall EKS module.

@terraform-aws-modules/triage-supporters what are your thoughts?
- add as independent sub-module
- do not add to this module
- other?

antonbabenko:
Good point, Bryant! We should try not to increase the complexity by adding more and more features if they don't belong there but I think EKS-addons _belong_ to the root module.

I think this PR can be simplified by having one resource `aws_eks_addon` in the root module which will accept a map of elements (one per addon). There is no need to maintain it as a separate submodule.

nitrocode:
@antonbabenko like this?

```hcl
module "eks" {
  # ...
  addons = [
    {
      addon_name               = "vpc-cni"
      addon_version            = "v1.8.0-eksbuild.1"
      resolve_conflicts        = "NONE"
      service_account_role_arn = null
    }
  ]
  # ...
}
```

antonbabenko:
@nitrocode Yes, but instead of a `list` it can be `map` (benefits: no need to make a conversion in `for_each` and it indicates that key in the map will be used as a key for resource):

```hcl
module "eks" {
  # ...
  addons = {
    "vpc-cni" = {
      addon_version     = "v1.8.0-eksbuild.1"
      resolve_conflicts = "NONE"
    }
  }
  # ...
}
```

stevehipwell:
Thanks @antonbabenko your design deals with my concerns and will allow anyone who wants to use addons to do so without impacting any of us who choose not to.

daroga0002:
adding https://github.com/hashicorp/terraform-provider-aws/issues/20725 as if this will be added to provider will allow to remove `coredns_versions`, `kube_proxy_versions` and `vpc_cni_versions` variables

ivankovnatsky:
still needed.

bryantbiggs:
not stale

beingamarnath:
Any updates on this? Now that EKS version 1.21 is out there, can we introduce this to the module directly

Lillecarl:
Not stale

antonbabenko:

mmerickel:
I think this was closed in error... I do not see any of the changes from this PR in master or the 18.0.3 release.

bryantbiggs:
@mmerickel https://github.com/terraform-aws-modules/terraform-aws-eks/blob/3c6686379de9489d4c044cd13a63e2329e2ae4e1/main.tf#L220

bryantbiggs:
yes, see convo above https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1443#issuecomment-897416141

there are examples of using this features as well https://github.com/terraform-aws-modules/terraform-aws-eks/blob/3c6686379de9489d4c044cd13a63e2329e2ae4e1/examples/complete/main.tf#L29

PR #1451, opened 2021-06-17
feat: Allow users to add more Audiences to OpenID Connect

### LINKED ISSUE #1145: enable_irsa=true creates OIDC provider with audience pointing to global sts endpoint (sts.amazonaws.com)
# I have issues

## I'm submitting a...

* [ ] bug report
* [x ] feature request

EKE- 1.18, Fargate

when enable_irsa = true an OIDC provider is created correctly and the audience is set to "sts.amazonaws.com". However, in most corporate network one is encouraged to use local endpoint "sts.eu-central-1.amazonaws.com". In most of the corporate network I have experienced the  "sts.amazonaws.com" is blocked which creates issues for role assumption by containers.

Either based on region setting of the EKS cluster or through explicit setting of sts endpoint it would be nice to have to be able to configure OIDC audience (sts endpoint) explicitly with some property like

oidc_audience = sts.xxxx.com (if irsa is set tot true)
 

## Environment details

* Affected module version:
* OS: Linux
* Terraform version: 0.13.4

cabrinha:
# PR o'clock

Fixes #1145

## Description

I'd like to add more audiences to the OpenID Connect Provider: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1145

### Checklist


cabrinha:
@barryib can I get a review on this one?

daroga0002 (approved):

local.tf
```diff
@@ -44,7 +44,7 @@ locals {
   )
 
   ec2_principal = "ec2.${data.aws_partition.current.dns_suffix}"
-  sts_principal = "sts.${data.aws_partition.current.dns_suffix}"
+  sts_principal = concat(["sts.${data.aws_partition.current.dns_suffix}"], var.openid_connect_audiences)
```
antonbabenko, line 47: Shouldn't we remove empty elements from the list by wrapping this with `compact`? 

```suggestion
  sts_principal = compact(concat(["sts.${data.aws_partition.current.dns_suffix}"], var.openid_connect_audiences))
```

daroga0002: yup, you are right

antonbabenko:
Hi @cabrinha and thanks for the PR!

[v17.5.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.5.0) has been just released.

lisfo4ka:
@antonbabenko @daroga0002 folks, maybe it's too late to claim, but I assume that the name of the variable with the client id list should be something other than `sts_principal` since it can contain not only STS endpoint but any client ID that identifies the application. See an example, in the doc https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_openid_connect_provider
So I'm back to propose the change from my PR to have the separate variable for a full list with client ids. What do you think about this?

daroga0002:
if I understand correctly you want just to change a `locals` variable name?

I think as this PR was arleady merged and there was created release input variable should stay as is (`openid_connect_audiences`) but locals probably can be adjusted as this will make sense and will not cause any user facing changes.

antonbabenko:
@lisfo4ka Isn't `sts` a required item in the context of this module? I mean, sometimes one may want to have another value but in this module, EKS always expects to have at least one `sts...` item, right?

lisfo4ka:
@daroga0002 yes, you're right, I've meant just a local variable renaming.

But I see now that it's really not so important change since for the IRSA option the client id list will contain STS endpoints only. Thanks, @antonbabenko.

So let's leave as it is?

antonbabenko:
We can rename `locals` name without any problems. `client_id_list` as a local name, and value wrapped with `distinct()` would be great. Please make a PR.

lisfo4ka:
@antonbabenko  @daroga0002  please, find the discussed changes in #1561.
I've tested it with the latest `master`.

PR #1453, opened 2021-06-21
chore: Extract only tflint file in GH Actions

ArchiFleKs:
# PR o'clock

## Description

Fix CI error with pre-commit

### Checklist


daroga0002 (approved):
I checked and under their zip didnt find other files, but it should be neutral change and if they will change something in their releases then we will be secured

@antonbabenko  any thoughts here?

antonbabenko:
Hi @ArchiFleKs !

A similar fix we have already applied for `terraform-docs` where sometimes multiple files were present in a zip archive and it broke CI sometimes.

This PR is safe to merge but it won't fix anything unless `tflint` put more files into zip, so merging for the future generations anyway :)

PR #1445, opened 2021-06-15
feat: Added throughput support for root and EBS disks

junaid-ali:
# PR o'clock

## Description

Fixes #1444
When `gp3` disk type is used, throughput should be configurable for an ASG launch configuration. This PR adds that support.

Example usage:
```terraform
module "eks" {
  ...

  worker_groups_launch_template = [
    {
      name                   = "worker-group-gp3"
      instance_type          = "t3.small"
      asg_desired_capacity   = 1
      public_ip              = true
      root_volume_size       = 150
      root_volume_type       = "gp3"
      root_volume_throughput = 300 # setting throughput to 300 MiB/s
    },
  ]
}
```

### Checklist

~- [ ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation~

PR #1454, opened 2021-06-21
fix: Make minimum AWS provider version common

shoekstra:
# PR o'clock

## Description

The `node_group` module requires a more recent version of the AWS provider and as this is use by the EKS module, it should also have its version updated, and since the `fargate` module is called from the EKS module, it should also have it's version updated.

### Checklist

☝️ should happen via terraform-docs.

### DIFF
--- a/modules/fargate/versions.tf
+++ b/modules/fargate/versions.tf
@@ -2,6 +2,6 @@ terraform {
   required_version = ">= 0.13.1"
 
   required_providers {
-    aws = ">= 3.40.0"
+    aws = ">= 3.43.0"
   }
 }
--- a/versions.tf
+++ b/versions.tf
@@ -2,7 +2,7 @@ terraform {
   required_version = ">= 0.13.1"
 
   required_providers {
-    aws        = ">= 3.40.0"
+    aws        = ">= 3.43.0"
     local      = ">= 1.4"
     kubernetes = ">= 1.11.1"
     http = {


junaid-ali:
@shoekstra this PR should fix this as well: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1445
haven't got any reviews from maintainers yet.

daroga0002:
@shoekstra  could you update this in all `versions.tf` files:
```
$ find . -name versions.tf
./versions.tf
./examples/create_false/versions.tf
./examples/fargate/versions.tf
./examples/managed_node_groups/versions.tf
./examples/basic/versions.tf
./examples/spot_instances/versions.tf
./examples/instance_refresh/versions.tf
./examples/launch_templates_with_managed_node_groups/versions.tf
./examples/irsa/versions.tf
./examples/launch_templates/versions.tf
./examples/secrets_encryption/versions.tf
./modules/node_groups/versions.tf
./modules/fargate/versions.tf
```

antonbabenko (commented):
I don't think this PR is necessary because `modules/fargate` can be used as a standalone module and thus it has another version requirements.

There is no need to make versions of providers to match.

The only requirement is that the minimum version of providers contain features used by the module (see [CHANGELOG.md in Terraform AWS Provider repo](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md)).

@daroga0002 Let's close this PR unless version 3.43.0 is the minimum required version for certain features in the `master` branch?

shoekstra:
> There is no need to make versions of providers to match.

I agree with this, but IMO if a module is calling a module that has a higher requirement, then the caller module should match the minimum requirement of the called module.

I can no longer recall why I opened this, I guess there was some issue or I wouldn't have created the PR. If there is no functional difference with current master and this, or if no one else is affected, please feel free to close it. We are specifying our AWS provider version in our `versions.tf` so no longer affected.

PR #1447, opened 2021-06-16
fix: Fixed launch template version infinite plan issue and improved rolling updates

jaimehrubiks:
# PR o'clock

Resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1446

## Description

The recent PR (https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1419) added support to set the launch template version on auto-generated launch templates for managed node groups, with the purpose of delaying rolling updates. **However**, it adds a bug caused by setting "$Latest" as the parameter to the node groups API call, which causes a never settled terraform plan, as the $Latest (or $Default) string is accepted by the API but translated into the corresponding integer afterwards. Read here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#version

This PR fixes this issue and sets the corresponding integer directly. When the user specifies either "$Latest" or "$Default" on the "launch_template_version" variable, the module will now resolve the corresponding version number as an integer and use that value instead.

This PR also adds the variable "update_default_version" which will allow delaying the rolling update without knowing a specific version number.

**How to delay a Rolling Update?**

Using the default values:

    launch_template_version: "$Latest"
    update_default_version: true

We achieve: 1) Always use the latest changes on the launch template, 2) Push the $Default tag to the latest version

Whenever we need to delay the rolling update, we can simply set:

    launch_template_version: "$Default"
    update_default_version: false

And when we apply, only the launch template will be changed, but not the node groups, as they will reference to the old "default" which was the last time when update_default_version was set to true

When we revert those changes, the node groups will be updated.

Also, by setting: 

        launch_template_version: "$Default"

In all cases, will achieve the same result and we only need to play with update_default_version variable

Note that you cant make it such that "new instances use the new launch template but we don't roll update the old one". You can, however, delay the rolling update as stated above, and try if the new version is working by manually going to the AWS cli and creating a new VM out of the launch template version desired (the latest number for example).

PR #1461, opened 2021-06-25
fix: worker security group handling when worker_create_security_group=false

sunghospark-calm:
# PR o'clock

## Description
Currently when you enable create_launch_template=true on the nodegroup, it switches from using primary cluster security group to worker security group which can cause various problems.

The first work around attempted was to use cluster_primary_security_group_id as worker_security_group_id, but that caused variable cycle error.
```
Error: Cycle: module.eks.local.cluster_primary_security_group_id (expand), module.eks.output.cluster_primary_security_group_id (expand), module.eks.var.worker_security_group_id (expand), module.eks.local.worker_security_group_id (expand), module.eks.aws_security_group_rule.cluster_https_worker_ingress, module.eks.aws_eks_cluster.this
```

The second workaround was to add the primary security group as worker_additional_security_group_ids, however that causes the ALB ingress controller to fail in AWS because the ingress controller cannot work when there are multiple SG attached to the network interface.

The solution was to delete the unused worker security group, however that causes the failure on launch template creation due to the empty security group id.

This PR cleans up the security groups for launch template if theres an empty string in the list.
There was also a `cluster_private_access_sg_source` rule which was trying to add the worker security group when the worker security group didn't exist, which is fixed in this PR.

### Checklist

### DIFF
--- a/cluster.tf
+++ b/cluster.tf
@@ -77,7 +77,7 @@ resource "aws_security_group_rule" "cluster_egress_internet" {
 }
 
 resource "aws_security_group_rule" "cluster_https_worker_ingress" {
-  count                    = var.cluster_create_security_group && var.create_eks ? 1 : 0
+  count                    = var.cluster_create_security_group && var.create_eks && var.worker_create_security_group ? 1 : 0
   description              = "Allow pods to communicate with the EKS cluster API."
   protocol                 = "tcp"
   security_group_id        = local.cluster_security_group_id
--- a/modules/node_groups/launch_template.tf
+++ b/modules/node_groups/launch_template.tf
@@ -48,7 +48,7 @@ resource "aws_launch_template" "workers" {
   network_interfaces {
     associate_public_ip_address = lookup(each.value, "public_ip", null)
     delete_on_termination       = lookup(each.value, "eni_delete", null)
-    security_groups = flatten([
+    security_groups = compact(flatten([
       var.worker_security_group_id,
       var.worker_additional_security_group_ids,
       lookup(
@@ -56,7 +56,7 @@ resource "aws_launch_template" "workers" {
         "additional_security_group_ids",
         null,
       ),
-    ])
+    ]))
   }
 
   # if you want to use a custom AMI


sunghospark-calm:
Hello @barryib @max-rocket-internet , I'm hoping to get some reviews on this PR. Please let me know if I should be pinging different person or channel. Thank you for maintaining this project.

yafanasiev:
Just hit this as well as https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1447. Had to fork the base repo, and everything now works beautifully. Hope this gets merged soon.

aidan-mundy:
I am having the same issue and this appears to be a direct fix for the problem, would appreciate a speedy merge on this small PR.

yafanasiev:
This is the only PR preventing us to go back from fork to official module. I can confirm that the fix is actually working and battle-tested. Please, review this 😢

antonbabenko:
Hi @yafanasiev ! I will review it during this week and merge it. Sorry that it takes so long time.

antonbabenko:
I couldn't make it to work on an existing cluster created with `examples/launch_templates`.

Error messages are like this:

```
╷
│ Error: Error updating Auto Scaling Group: ValidationError: You must use a valid fully-formed launch template. Value () for parameter groupId is invalid. The value cannot be empty
│       status code: 400, request id: 00236532-0ebb-412d-9f64-289d0ff947c2
│ 
│   with module.eks.aws_autoscaling_group.workers_launch_template[2],
│   on ../../workers_launch_template.tf line 3, in resource "aws_autoscaling_group" "workers_launch_template":
│    3: resource "aws_autoscaling_group" "workers_launch_template" {
│ 
╵
```

Could you please try to create a cluster using examples, then apply this PR (add `worker_create_security_group = false` and other required inputs), and run terraform plan/apply to confirm that it works as expected.

Please update the example, if necessary.

sunghospark-calm:
Thanks for the review @antonbabenko .
I could confirm that just adding `worker_create_security_group = false` alone did not work with `example/launch_templates`
The required blob was

```
  worker_create_security_group = false
  worker_create_cluster_primary_security_group_rules = true
  worker_additional_security_group_ids = [module.eks.cluster_primary_security_group_id]
```

I think it makes sense because there gotta be a security group in place before the cluster to work.

I also tested with `example/managed_node_groups`. For this one I could just add `worker_create_security_group = false` and the apply ran successfully.

antonbabenko:
Thanks, @sunghospark-calm !

[v17.13.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.13.0) has been just released.

Issue #1462, opened 2021-06-26
Implicit node_group object creation is flawed


apeschel:
## Description

The node_group object used in this module has the "optional" type attribute "name" that significantly changes the behavior of this module if it's set.

This module manages to do this by avoiding declaring an explicit object definition for node_group, and instead dynamically constructing the node_group object, as seen here:

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v17.1.0/modules/node_groups/locals.tf

This is a huge flaw, because terraform doesn't normally support objects with optional type attributes, and there is no practical way to conditionally set the "name" attribute without running into type errors.

We have several existing clusters that need to have the name field set, but we don't want to set it for new clusters. Unfortunately, there's no way to use common code to do this with the current node_group object. The terraform type system just explodes.

This module should explicitly define the node_group object, which would help with the type issues around the "name" attribute.

If this module wants to use an optional attribute, there is an experimental terraform feature for using optional attributes. While it would mean using an experimental module, it would have the massive benefit of not completely breaking the terraform type system, and would allow for users to come up with ways to optionally set the name attribute.

https://www.terraform.io/docs/language/expressions/type-constraints.html#experimental-optional-object-type-attributes

jaimehrubiks:
Just for clarification, can you explain how you reproduce this bug? Maybe that can help someone finding a solution. Does this happen when you mix node gruops with and without name field on the same cluster?

apeschel:
@jaimehrubiks 

Here's an example of what I'm trying to do:

```
locals {
  node_group_base = {
    desired_capacity  = 6
    max_capacity      = 9
    min_capacity      = 6
    instance_types    = ["r5.xlarge"]
    subnets           = [element(module.vpc.private_subnets, 0)]
    disk_size         = 20
  }
  node_group_name_hack = {
    name              = var.cluster_node_groups_name
  }
  node_groups_expanded = var.cluster_node_groups_name != null ? merge(local.node_group_base, local.node_group_name_hack) : local.node_group_base
}

module "eks" {
  source          = "terraform-aws-modules/eks/aws"
  version         = "17.1.0"
  cluster_name    = var.cluster_name
  subnets         = module.vpc.private_subnets
  tags            = var.custom_tags

  vpc_id = module.vpc.vpc_id

  node_groups = {
    pool = local.node_groups_expanded
  }
}
```

[...]

PR #1464, opened 2021-06-26
fix: allow node_group attributes to be omitted

apeschel:
# PR o'clock

## Description

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1462

This fix allows the node_group attributes name and name_prefix to be
explicitly omitted by setting them to null.

This fixes an edge case where the attribute was set, but null, which
would cause the conditional logic to be incorrect.

### Checklist

### DIFF
--- a/modules/node_groups/locals.tf
+++ b/modules/node_groups/locals.tf
@@ -32,13 +32,19 @@ locals {
     v,
   ) if var.create_eks }
 
-  node_groups_names = { for k, v in local.node_groups_expanded : k => lookup(
-    v,
-    "name",
-    lookup(
-      v,
-      "name_prefix",
-      join("-", [var.cluster_name, k])
-    )
-  ) }
+  # This node_groups_names construct is a consequence of not explicitly
+  # declaring the node_group object. For more information, refer to this issue:
+  # https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1462
+  node_groups_names = {
+    for k, v in local.node_groups_expanded :
+      k => (
+        lookup(v, "name", null) != null
+          ? v["name"]
+          : (
+            lookup(v, "name_prefix", null) != null
+              ? v["name_prefix"]
+              : join("-", [var.cluster_name, k])
+          )
+      )
+  }
 }


apeschel:
I quickly tested and verified locally that this corrects the ability to omit the name and name_prefix node_group attributes by setting them to null.

azizea (approved):
LGTM!

apeschel:
@barryib 

Apologies for pinging you, but is there a chance someone could take a look at this, please?

apeschel:
@barryib This is a pretty big bug with a pretty easy fix that prevents us from using a newer version of terraform, could we please get some attention on this?

apeschel:
I've created a fork with this fix until the maintainer returns.

You can use it by setting the source as follows:

```
module "eks" {
  source = "git@github.com:ElementAnalytics/terraform-aws-eks?ref=v17.1.1"
}
```

modules/node_groups/locals.tf
```diff
@@ -27,13 +27,23 @@ locals {
...
+  # if it does exist. We cannot take a shortcut by chaining the lookups in the
+  # default condition, since the default condition will not be reached if the
+  # attribute exists and is null.
+  #
+  # It would be ideal to rework this module to explicitly declare
+  # the node_group object, as that would simplify the logic.
+  #
+  # For more information, refer to this issue:
+  # https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1462
+  node_groups_names = { for k, v in local.node_groups_expanded : k => lookup(v, "name", null) != null ? v["name"] : lookup(v, "name_prefix", null) != null ? v["name_prefix"] : join("-", [var.cluster_name, k]) }
```
bryantbiggs, line 48: hi @apeschel - could you re-expand the formatting so that its a bit easier to read (hard to read on one line)

apeschel: @bryantbiggs absolutely, should only take a bit

apeschel:
@bryantbiggs Just pushed an update to break it up into multiple lines

apeschel:
@bryantbiggs Hello, and chance we could get some follow up on this, please?

bryantbiggs:
hi @apeschel I haven't had time to test this yet, apologies - could you update an example to show how this functionality is intended to work please? https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples

apeschel:
@bryantbiggs Hello, I've updated the `managed_node_groups` with details on how to take advantage of terraform functionality to use newer versions of this module with clusters built using older versions of this module.

This is only possible with the bug fix in place. As previously mentioned, the bug breaks the normal terraform behavior, and this example isn't possible.

apeschel:
@bryantbiggs Hello, could I get some follow up on this please?

bryantbiggs:
apologies, I haven't had time - @daroga0002 what do you think?

daroga0002:
I feel a pain of upgrading, but as we arleady decided to release breaking changes in module I really wish to avoid this kind tricks in code (as this code looks really hacky style).

Those are just my open thoughts.

daroga0002:
I am looked over this from mobile, maybe I will later take a look from computer as maybe it is not so scary as looks on mobile.


I will take a look into this tomorrow.

apeschel:
> I really wish to avoid this kind tricks in code (as this code looks really hacky style).

@daroga0002 With all respect, the original code snippet was incredibly arcane and hacky to the point that it was broken and no one realized. This is just a bandaid with lots of comments to make it clear that it's arcane and hacky, and why it's being done.

It also has the benefit that it actually works.

daroga0002:
I am checking a releases and see in v17.0.0 was attached this note https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/upgrades.md

does this is not working?

apeschel:
@daroga0002 No, because you can't ommit the node_group attribute by setting it to `null`. This attribute is broken with the current code.

daroga0002:
Ok, I analysed this again and now understand issue. 

Correct me if I am wrong but you want to use common code where you use node groups definition:
```
name = var.cluster_node_groups_name
```

and for some clusters is:
`cluster_node_groups_name=null`
where for others:
`cluster_node_groups_name=something`

apeschel:
@daroga0002 Yes, that's exactly right - unfortunately the current code doesn't work correctly with the `null` branch.

examples/managed_node_groups/main.tf
```diff
@@ -74,6 +74,16 @@ module "eks" {
 
   node_groups = {
     example = {
+      # In version 1.17 of this module, the way that node group names are
```
daroga0002, line 77: I dont think we require change examples. Adding those comments can be misleading for new users. Lets dont touch examples at all as I dont think is required

examples/managed_node_groups/variables.tf
```diff
@@ -50,3 +50,9 @@ variable "map_users" {
     },
   ]
 }
+
+variable "cluster_node_groups_name" {
```
daroga0002, line 54: same as above

modules/node_groups/locals.tf
```diff
@@ -31,13 +31,34 @@ locals {
...
-  node_groups_names = { for k, v in local.node_groups_expanded : k => lookup(
-    v,
-    "name",
-    lookup(
-      v,
-      "name_prefix",
-      join("-", [var.cluster_name, k])
-    )
-  ) }
+  # This node_groups_names construct is a consequence of not explicitly
```
daroga0002, line 34: I think that this comment is also not required as if somebody will try to understand this then will need analyse that conditional

daroga0002 (commented):
In general it is now clear what we trying to do and I am ok to merge it. The only think is that we shouldn`t change examples as this is not required. 

ALso comment from before conditional maybe is not required or can be limited to 1-2 lines explaining why this structure is required, something like:
```
This structure is required to allow passing string or null variables into `name_prefix`/ `name` variables. As terraform expect defined variable types.
```

daroga0002:
@apeschel  can you update PR as per comments?

apeschel:
@daroga0002 Just pushed an update - I left a small comment for some context, please let me know if you want that removed as well. Should only take a moment.

daroga0002:
I am trying to replicate this issue (for testing) and seems on current master of module I am not able to achieve error. 

I have:
```
$ terraform version
Terraform v1.0.6
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v3.57.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.4.1
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1
```

and code of module:
```
module "eks_cluster" {
  create_eks = var.eks_create
  # https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest
  version         = "17.15.0"
................
  node_groups = {
    pool = {
      name              = var.cluster_node_groups_name
      desired_capacity  = 2
      max_capacity      = 2
      min_capacity      = 2
      instance_types    = ["r5.xlarge"]
      subnets           = ["subnet-07a78a7e8184e3702"]
      disk_size         = 20
    }
  }
  ```


variable set to:
```
cluster_node_groups_name = null
```


@apeschel which version of terraform you saw this errors?

apeschel:
I was using terraform 0.14.X at the time, I believe. Let me see if I can reproduce it on 1.0.X

apeschel:
@daroga0002 I'm not able to replicate with 1.0.6 either. Looks like terraform changed the behavior of the ternary operator at some point. Feel free to close this, I'll reopen it if the issue comes up again.

PR #1209, opened 2021-01-30
feat: Allow customization of aws_eks_node_group timeouts

Dhouti:
# PR o'clock

## Description

Allows configuration of timeouts for create/update/delete on the aws_eks_node_group resource.

Relevant issue: #1131

### Checklist

- [x] CI tests are passing
- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

Issue #1131, opened 2020-12-02
Configurable timeouts on creating workers node group.


samcday:
# I have issues

## I'm submitting a...

* [ ] bug report
* [x] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

Creating a new node group can take longer than an hour, which causes the `node_groups.aws_eks_node_group.workers` resource to timeout.

## What's the expected behavior?

Timeout behaviour for `node_groups.aws_eks_node_group.workers` should be configurable!

## Environment details

* Affected module version: v12.2.0
* OS: Linux
* Terraform version: 0.12.x

ruckc:
This is a constant pain point with automating EKS clusters

bryantbiggs:
should be fixed shortly

antonbabenko:
Fixed in #1552 

Released in [v17.9.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.9.0).

PR #1467, opened 2021-06-28
fix: Node Group Timeout Configuration

njgibbon:
# PR o'clock

## Description

Set the terraform 'Operation timeouts'[[0]] for the Node Groups[[1]].

I recently found that sometimes updating node groups with large nodes can sometimes take more than 1 hour which is the default. When terraform times out in these circumstances it gets into quite an ugly State mess.

This pull request handles these scenarios without requiring changes to any interfaces by just setting the defaults to 3 hours instead of 1.

I've tagged this as 'fix' because the default configuration causes errors within a fairly standard expected use-case. But, i'm comfortable changing that tag 🙂 

There is another PR open (but out of sync) which does the same work, but enables via a map var so that the timeouts are configurable. I honestly think this (change defaults to 3h) is sufficient and will immediately help users with minimal change and should be merged even if you ultimately want to make these values configurable.

This is currently blocking some users from using the module or causing them to maintain a fork.

**Linked Issues**
* https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1209
* https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1131

[0]: https://www.terraform.io/docs/language/resources/syntax.html#operation-timeouts
[1]: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#timeouts
### Checklist

### DIFF
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -80,6 +80,12 @@ resource "aws_eks_node_group" "workers" {
     lookup(var.node_groups[each.key], "additional_tags", {}),
   )
 
+  timeouts {
+    create = "2h"
+    update = "2h"
+    delete = "2h"
+  }
+
   lifecycle {
     create_before_destroy = true
     ignore_changes        = [scaling_config.0.desired_size]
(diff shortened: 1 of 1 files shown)


daroga0002:
similar PR https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1552

bryantbiggs:
closing in favor of #1552 since it has a better setup for flexibility

PR #1469, opened 2021-07-01
feat: Allow multiple SGs to added to EKS endpoint

stargriv:
# PR o'clock

## Description

Added logic to allow adding existing AWS Security Groups to EKS endpoint.

### Checklist

- [] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
 | cluster\_log\_kms\_key\_id | If a KMS Key ARN is set, this key will be used to encrypt the corresponding log group. Please be sure that the KMS Key has an appropriate key policy (https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/encrypt-log-data-kms.html) | `string` | `""` | no |
 | cluster\_log\_retention\_in\_days | Number of days to retain log events. Default retention - 90 days. | `number` | `90` | no |
 | cluster\_name | Name of the EKS cluster. Also used as a prefix in names of related resources. | `string` | n/a | yes |
-| cluster\_security\_group\_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingress/egress to work with the workers | `string` | `""` | no |
+| cluster\_security\_group\_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingress/egress to work with the workers. | `string` | `""` | no |
+| cluster\_additional\_security\_group\_ids | If provided, additional security groups will be attachjed to EKS cluster. If not given, cluster\_security\_group\_id will be used instead. | `list(string)` | `[]` | no |
 | cluster\_service\_ipv4\_cidr | service ipv4 cidr for the kubernetes cluster | `string` | `null` | no |
 | cluster\_version | Kubernetes version to use for the EKS cluster. | `string` | n/a | yes |
 | config\_output\_path | Where to save the Kubectl config file (if `write_kubeconfig = true`). Assumed to be a directory if the value ends with a forward slash `/`. | `string` | `"./"` | no |
--- a/cluster.tf
+++ b/cluster.tf
@@ -15,7 +15,7 @@ resource "aws_eks_cluster" "this" {
   tags                      = var.tags
 
   vpc_config {
-    security_group_ids      = compact([local.cluster_security_group_id])
+    security_group_ids      = compact(concat(tolist([local.cluster_security_group_id]), local.cluster_additional_security_group_ids))
     subnet_ids              = var.subnets
     endpoint_private_access = var.cluster_endpoint_private_access
     endpoint_public_access  = var.cluster_endpoint_public_access
--- a/local.tf
+++ b/local.tf
@@ -1,10 +1,11 @@
 locals {
 
-  cluster_security_group_id         = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
-  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : null
-  cluster_iam_role_name             = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
-  cluster_iam_role_arn              = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
-  worker_security_group_id          = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
+  cluster_security_group_id             = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -183,7 +183,8 @@ MIT Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-a
...
 | cluster\_log\_retention\_in\_days | Number of days to retain log events. Default retention - 90 days. | `number` | `90` | no |
 | cluster\_name | Name of the EKS cluster. Also used as a prefix in names of related resources. | `string` | n/a | yes |
-| cluster\_security\_group\_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingress/egress to work with the workers | `string` | `""` | no |
+| cluster\_security\_group\_id | If provided, the EKS cluster will be attached to this security group. If not given, a security group will be created with necessary ingress/egress to work with the workers. | `string` | `""` | no |
+| cluster\_additional\_security\_group\_ids | If provided, additional security groups will be attachjed to EKS cluster. If not given, cluster\_security\_group\_id will be used instead. | `list(string)` | `[]` | no |
```
daroga0002, line 187:
```suggestion
| cluster\_additional\_security\_group\_ids | If provided, additional security groups will be attached to EKS cluster. If not given, cluster\_security\_group\_id will be used instead. | `list(string)` | `[]` | no |
```

variables.tf
```diff
@@ -27,6 +27,12 @@ variable "cluster_security_group_id" {
   default     = ""
 }
 
+variable "cluster_additional_security_group_ids" {
+  description = "A list of additional security group ids to attach to EKS cluster"
```
daroga0002, line 31: this should be exactly same as in `README.md` as that file is generated by `terraform-docs` basing on variable/output descriptions

cluster.tf
```diff
@@ -15,7 +15,7 @@ resource "aws_eks_cluster" "this" {
   tags                      = var.tags
 
   vpc_config {
-    security_group_ids      = compact([local.cluster_security_group_id])
+    security_group_ids      = compact(concat(tolist([local.cluster_security_group_id]), local.cluster_additional_security_group_ids))
```
daroga0002, line 18: Maybe more clear to read and understand code will be making whole merging operation in `locals.tf` and here just pass one variable something like: `security_group_ids = local.cluster_security_group_ids` ?

local.tf
```diff
@@ -1,10 +1,11 @@
...
 
-  cluster_security_group_id         = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
-  cluster_primary_security_group_id = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : null
-  cluster_iam_role_name             = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.name) : var.cluster_iam_role_name
-  cluster_iam_role_arn              = var.manage_cluster_iam_resources ? join("", aws_iam_role.cluster.*.arn) : join("", data.aws_iam_role.custom_cluster_iam_role.*.arn)
-  worker_security_group_id          = var.worker_create_security_group ? join("", aws_security_group.workers.*.id) : var.worker_security_group_id
+  cluster_security_group_id             = var.cluster_create_security_group ? join("", aws_security_group.cluster.*.id) : var.cluster_security_group_id
+  cluster_primary_security_group_id     = var.cluster_version >= 1.14 ? element(concat(aws_eks_cluster.this[*].vpc_config[0].cluster_security_group_id, list("")), 0) : null
+  cluster_additional_security_group_ids = var.cluster_additional_security_group_ids
```
daroga0002, line 5: maybe better will be something like:
```
cluster_security_group_ids = compact(concat([cluster_security_group_id], var.cluster_additional_security_group_ids)
```

as from EKS perspective there is no primary or additional security groups meaning

daroga0002 (commented):
I am thinking as initially EKS was allowing to define just single security group so this is reason why we have `cluster_security_group_id` but later they added possibility to add multiple groups. 

So maybe we should think in long term to deprecate `cluster_security_group_id` input and replace by `cluster_security_group_ids` (instead `cluster_additional_security_group_ids`). In transition period we can handle both, but in some bigger breaking change refactor just deprecate  singular `cluster_security_group_id` in favour of
 `cluster_security_group_ids`.
 
Then here we can start this deprecation period from this PR and also update all examples to follow new plural input variable.

@antonbabenko @bryantbiggs @lisfo4ka any thoughts?

local.tf
lisfo4ka: Agreed with @daroga0002 that this interpolation should be in locals, but not in cluster.tf.
And I assume flatten() can be used here as well

lisfo4ka:
@daroga0002 maybe we can leave `var.cluster_additional_security_group_ids` for case with `cluster_create_security_group=true` which creates the main `aws_security_group.cluster` SG with required rules and additionally provide an option to pass some pre-existing SGs via `cluster_additional_security_group_ids`. What do you think?

daroga0002:
> @daroga0002 maybe we can leave `var.cluster_additional_security_group_ids` for case with `cluster_create_security_group=true` which creates the main `aws_security_group.cluster` SG with required rules and additionally provide an option to pass some pre-existing SGs via `cluster_additional_security_group_ids`. What do you think?

In general any group which we creating here is always `additional` even if we call it `aws_security_group.cluster`. This is because EKS cluster silently always create primary security group which we dont manage by this module.

More details about this here https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1542

lisfo4ka:
> > @daroga0002 maybe we can leave `var.cluster_additional_security_group_ids` for case with `cluster_create_security_group=true` which creates the main `aws_security_group.cluster` SG with required rules and additionally provide an option to pass some pre-existing SGs via `cluster_additional_security_group_ids`. What do you think?
> 
> In general any group which we creating here is always `additional` even if we call it `aws_security_group.cluster`. This is because EKS cluster silently always create primary security group which we dont manage by this module.
> 
> More details about this here #1542

Yeah, I saw that hidden SG)) I just think that we can mix up the pre-existing SGs with that one which we create with `cluster_create_security_group=true` option, so we need two separate variables.

daroga0002:
> Yeah, I saw that hidden SG)) I just think that we can mix up the pre-existing SGs with that one which we create with `cluster_create_security_group=true` option, so we need two separate varia

I general if `cluster_create_security_group=true` then we have empty input cluster_security_group_id. If we have this to `false` then in general we dont care at all security groups as this is on user side to connect them (if I remember correctly).


So as per me plural `cluster_security_group_ids` will help to reduce a input (which we currently have so many)

bryantbiggs:
> I am thinking, as initially EKS was allowing to define just single security group so this is reason why we have `cluster_security_group_id` but later they added possibility to add multiple groups.
> 
> So maybe we should think in long term to deprecate `cluster_security_group_id` input and replace by `cluster_security_group_ids` (instead `cluster_additional_security_group_ids`). In transition period we can handle both, but in some bigger breaking change refactor just deprecate singular `cluster_security_group_id` in favour of `cluster_security_group_ids`.
> 
> Then here we can start this deprecation period from this PR and also update all examples to follow new plural input variable.
> 
> @antonbabenko @bryantbiggs @lisfo4ka any thoughts?

yes I agree - I think we could easily include this in the next breaking change

antonbabenko:

jaredready:
Is there any possibility of back-porting this to 17.x?

bryantbiggs:
unfortunately that gets very difficult and burdensome and so we only support the current version

PR #1505, opened 2021-07-28
feat: Support offline bootstrap of Windows nodes

jaroslav-fedor-swi:
# PR o'clock

## Description

In case you need to spin nodes in environment, that does not have egress to internet windows nodes does not connect to cluster. The reason behind that is that `Start-EKSBootstrap.ps1` which is baked in windows EKS AMI is executed with minimal set of parameters which then requires egress to `eks.<region>.amazonaws.com`. This endpoint does not support private link and does not come with managed prefix list. The only way to make the Boostrap script offline is to pass additional arguments - private EKS endpoint and Cluster CA. 

### Checklist


daroga0002:
could you add working example of EKS with Windows nodes to https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples ?

bryantbiggs:
this looks like an area where we just make a generic variable and its up to users to provide whatever arguments they want for input rather than building out this massive string of input options

daroga0002:
I think I require to run procedure to enable windows support over EKS.

I will try that approach

examples/windows_launch_templates/README.md
```diff
@@ -0,0 +1,65 @@
+# Launch templates example
+
+This is EKS example using workers launch template with worker groups feature.
```
daroga0002, line 3: Please update that this require executing steps from https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html#macos-and-linux because windows nodes will not join cluster

examples/windows_launch_templates/main.tf
```diff
@@ -0,0 +1,114 @@
...
+          propagate_at_launch = true
+        },
+        {
+          key                 = "k8s.io/cluster-autoscaler/enabled"
+          value               = true
+          propagate_at_launch = true
+        }
+      ]
+    },
+    // default Linux worker group launch template for big workloads
```
daroga0002, line 52: Please add one node group or linux worker, for example:

```
  node_groups = {
    example = {
      desired_capacity = 1
      max_capacity     = 10
      min_capacity     = 1
      instance_types = ["t3.large"]
    }
}
```

as only windows cluster will be not working in example

daroga0002 (changes requested):
@jaroslav-fedor-swi  2 comments to address before merging

examples/windows_launch_templates/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  ]
```
daroga0002, line 53:
```suggestion
  ]
  node_groups = {
    linux = {
      desired_capacity = 1
      max_capacity     = 10
      min_capacity     = 1
      instance_types = ["t3.large"]
    }
  }
```

daroga0002: @jaroslav-fedor-swi  can you commit those two suggestions into your branch?

antonbabenko:

PR #1528, opened 2021-08-12
feat: Ignore NG's Launch Template Version Change

Efrat19:
# PR o'clock
As a heavy user of this module, every time I run the `terraform apply` I get frustrated by the fact I all of my managed nodegroups are updated, regardless of the change I made to the TF stack.
The update is to the LT version, from "2" to "$Latest", even though in my configuration the version is already set to "latest":

My configuration look like this:
```terrraform

module "eks" {
  .....
  workers_group_defaults = {
    .....
    create_launch_template   = true
    launch_template_version = "$Latest"
  }
}
```

But when I run terraform plan, I see that terraform is aiming to change the version:
image

## Description
I added the LT version to the ignore list in hope this may help other people using this module

### DIFF
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -82,7 +82,10 @@ resource "aws_eks_node_group" "workers" {
 
   lifecycle {
     create_before_destroy = true
-    ignore_changes        = [scaling_config.0.desired_size]
+    ignore_changes        = [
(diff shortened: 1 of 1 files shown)


johngmyers:
#1447 looks like a much better fix, since it handles the case where the specified launch template version is not "$Latest".

Efrat19:
I agree, thanks!

Issue #1245, opened 2021-02-16
Unable to have more than one profile in fargate_profiles


brunoyb:
# I have issues

I'm using Terraform v0.14.6. When I try to add a new item to `fargate_profiles` I get an error during `terraform plan`.

## I'm submitting a...

* [X] bug report
* [ ] feature request
* [ ] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [ ] kudos, thank you, warm fuzzy

## What is the current behavior?

`terraform plan` shows this:

```
Error: Inconsistent conditional result types

  on .terraform/modules/my_eks/modules/fargate/fargate.tf line 17, in resource "aws_eks_fargate_profile" "this":
  17:   for_each               = local.create_eks ? local.fargate_profiles_expanded : {}
    |----------------
    | local.create_eks is true
    | local.fargate_profiles_expanded is object with 4 attributes

The true and false result expressions must have consistent types. The given
expressions are object and object, respectively.
```

## If this is a bug, how to reproduce? Please include a code sample if relevant.

Have an instance of this module with `fargate_profiles`, after a successful `terraform apply` try to add a new item to `fargate_profiles` and run `terraform plan`.

EDIT: Looks like this happens whenever you have more than one fargate profile, not only when adding a new one (see comments below).

## What's the expected behavior?

`terraform plan` should work and show the new `aws_eks_fargate_profile` resources being created.

## Are you able to fix this problem and submit a PR? Link here if you have already.

I don't have a PR because I'm not sure what is correct solution for this, but I made a copy of this module locally and changed this:

```
resource "aws_eks_fargate_profile" "this" {
  for_each               = local.create_eks ? local.fargate_profiles_expanded : {}
```

With this:

```
resource "aws_eks_fargate_profile" "this" {
  for_each               = local.create_eks ? local.fargate_profiles_expanded : null
```

To workaround the issue for now.

In other words, I changed `{}` with `null`. Not even sure how `for_each` works with `null`, but since it didn't complain, I decided to leave it like this until we find the correct fix.

## Environment details

* Affected module version: 14.0.0
* OS: Some recent version of Amazon Linux I believe
* Terraform version: 0.14.6

Thanks!

JohnHedman:
> Are you able to fix this problem and submit a PR? Link here if you have already.
[...]

PR #1532, opened 2021-08-16
fix: unable to have more than one fargate profile

ehrenkret:
# PR o'clock

## Description

Resolves issue #1245

`local.fargate_profiles_expanded` is always empty when `local.create_eks` is false so the ternary operator here can be omitted.

`local.create_eks` is defined as `var.create_eks && length(var.fargate_profiles) > 0`

`local.fargate_profiles_expanded` is defined as `{ for k, v in var.fargate_profiles : k => merge(
    v,
    { tags = merge(var.tags, lookup(v, "tags", {})) },
  ) if var.create_eks }`

If `var.create_eks` is false, then `local.fargate_profiles_expanded` will have all entries filtered out by the conditional in the for expression.
If `var.fargate_profiles` is length 0, then `local.fargate_profiles_expanded` will have no entries to iterate over in the for expression.

### Checklist

- [x] Add semantics prefix to your PR or Commits (at least one of your commit groups)

### DIFF
--- a/modules/fargate/fargate.tf
+++ b/modules/fargate/fargate.tf
@@ -14,7 +14,7 @@ resource "aws_iam_role_policy_attachment" "eks_fargate_pod" {
 }
 
 resource "aws_eks_fargate_profile" "this" {
-  for_each               = local.create_eks ? local.fargate_profiles_expanded : {}
+  for_each               = local.fargate_profiles_expanded
   cluster_name           = var.cluster_name
   fargate_profile_name   = lookup(each.value, "name", format("%s-fargate-%s", var.cluster_name, replace(each.key, "_", "-")))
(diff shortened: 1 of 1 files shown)


daroga0002:
local.fargate_profiles_expanded:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/85ca4ca93d519fb4e2799c510a3323ef4116d7d7/modules/fargate/locals.tf#L6-L9

antonbabenko:
Fixed in #1583 (9885ac0)

antonbabenko:
[v17.19.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.19.0) has been just released with this fix.

PR #1537, opened 2021-08-19
fix: Add dependens_on eks cluster to the kubeconfig file

joanayma:
# PR o'clock

## Description

Simple depends_on addition. Prevent deletion/creation of kubeconfig file before the cluster actually exists or deleted.
In case the terraform destroy fails and the cluster is still there with some issue, we might need to use the kubeconfig.
Right now the behaviour deletes the kubeconfig file too early.

### Checklist

### DIFF
--- a/kubectl.tf
+++ b/kubectl.tf
@@ -4,4 +4,7 @@ resource "local_file" "kubeconfig" {
   filename             = substr(var.kubeconfig_output_path, -1, 1) == "/" ? "${var.kubeconfig_output_path}kubeconfig_${var.cluster_name}" : var.kubeconfig_output_path
   file_permission      = var.kubeconfig_file_permission
   directory_permission = "0755"
+  depends_on = [
+    aws_eks_cluster.this,
+  ]
 }


aidan-mundy:
Add fix: to the front of the PR name if you want to pass the automated workflow checks

daroga0002:
I doubt that this will make sense as here is implicit dependency created, but even if you create this explicit then kubeconfig will must be destroyed before cluster (so nothing will change) as dependencies for destroy are working in reverse order, so:

create process:
1. create EKS
2. create kubeconfig

destroy process:
1. destroy kubeconfig
2. destroy EKS

nathanblair:
This should be re-opened. I am currently experiencing problems because the kubeconfig file is deleted and it has the required connection configuration to bring down the cluster when running `terraform destroy`. So instead of using the correct connection information it tries to ping against localhost resulting in a connection refused and failing to delete the cluster.

daroga0002:
> This should be re-opened. I am currently experiencing problems because the kubeconfig file is deleted and it has the required connection configuration to bring down the cluster when running `terraform destroy`. So instead of using the correct connection information it tries to ping against localhost resulting in a connection refused and failing to delete the cluster.

Could you explain me how depends_on will resolve it?

nathanblair:
I may be working from a misunderstanding of 'depends_on' as I try to avoid it if I can. But my thought process is that if the kube configmap (AWS auth) resource depends on the local file resource then the terraform system will postpone deleting the local file resource until all dependent resources have been deleted, including, in my case, the AWS auth configmap.

daroga0002:
> I may be working from a misunderstanding of 'depends_on' as I try to avoid it if I can. But my thought process is that if the kube configmap (AWS auth) resource depends on the local file resource then the terraform system will postpone deleting the local file resource until all dependent resources have been deleted, including, in my case, the AWS auth configmap.

Seems you have wrong understanding about dependencies. Please check https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1537#issuecomment-913560769 where I have explained it

nathanblair:
If that's how the behavior works I'd consider that a bug. That's not how depends_on should work.

joanayma:
@daroga0002 is right.

nathanblair:
If that's the case then there is literally no valid path to having the kubernetes 'load_config' option set to true. And should be stated in the documentation.

I maintain that if that's the way this works then it is a bug and the reasoning for stating otherwise seems tautological. Should an explicit 'depends_on' be made for the local kubeconfig if it is required then? Otherwise there is still no valid path for deleting a cluster that relies on information in a kubeconfig file.

PR #1538, opened 2021-08-20
fix: Allow instance `Name` tag to be overwritten

aidan-mundy:
# PR o'clock

## Description

Allow instance `Name` tag to be overwritten by the user-defined node-group variable `additional_tags`. Without this PR, it is impossible to define custom names for instances created as part of a managed node group while using a generated launch template. It was possible to affect the name of the instance by setting a different name for the node-group, but there does not appear to be a technical reason for this to be mandated. 

### Checklist

### DIFF
--- a/modules/node_groups/launch_template.tf
+++ b/modules/node_groups/launch_template.tf
@@ -77,11 +77,11 @@ resource "aws_launch_template" "workers" {
 
     tags = merge(
       var.tags,
-      lookup(var.node_groups_defaults, "additional_tags", {}),
-      lookup(var.node_groups[each.key], "additional_tags", {}),
       {
         Name = local.node_groups_names[each.key]
-      }
+      },
+      lookup(var.node_groups_defaults, "additional_tags", {}),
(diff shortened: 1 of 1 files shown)


goetzc:
Is this ([v17.18.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v17.4.0)) only for _launch_template_?  Instances generated by _node_groups_ still have no `Name` tag.

aidan-mundy:
@goetzc As far as I know, an ASG cannot propagate tags to instances without a launch template. Create a launch template and use it for your node group (or use create_launch_template=true) in order to support tags (including the "Name" parameter, which is just a resource tag that is treated specially in the Console UI)

lisfo4ka:
Folks, it's a nice idea, but what about worker groups? Should we provide the same option to define custom names to instances for worker groups as well?

There is quite a different approach, but it's hardcoded to the exact pattern without the option to redefine.

@aidan-mundy  @daroga0002 @antonbabenko

PR #1552, opened 2021-08-28
feat: Allow override of timeouts in node_groups

RobertKozak:
# PR o'clock

## Description

This allows timeouts to be configured for [aws_eks_node_group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#timeouts) using the standard timeout block. 

Updating node groups with large nodes often takes more than the default 1 hour and terraform can time out.

This pull request handles this by allowing you to add a timeout block to either the **node_groups_defaults** or to individual **node_groups**.

This PR also respects the timeouts described [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#timeouts)

NOTE: You can set either / or create, update, delete and you will get defaults for the ones not supplied 

There are a couple other PRs for this same issue #1173 but one PR #1209 is very old with no changes since Jan 30 and the other PR #1467 is just too simplistic for a long term solution. 

### Linked PRs and Issues

- #1209
- #1467
- #1131

### Checklist


modules/node_groups/node_groups.tf
```diff
@@ -69,6 +69,12 @@ resource "aws_eks_node_group" "workers" {
     }
   }
 
+  timeouts {
+    create = lookup(each.value["timeouts"], "create", "60m")
```
bryantbiggs, line 73: for the fallback on all of these, can we use `null` so that we use whats provided by the upstream provider (looks like it defaults to an hour https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#timeouts )

```suggestion
    create = lookup(each.value["timeouts"], "create", null)
    update = lookup(each.value["timeouts"], "update", null)
    delete = lookup(each.value["timeouts"], "delete", null)
```

bryantbiggs:
actually, looks like theres a bit more - I think what is captured in this PR is what we are looking for https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1209 

if you update with those changes we can get this merged

modules/node_groups/node_groups.tf
RobertKozak: I'll try that. When I was doing my testing it didn't actually fallback and created empty values that broke the plan. But I will try again.

RobertKozak:
I'll try again and get this updated after some testing on my end

RobertKozak:
@bryantbiggs OK. all checked in and ready

antonbabenko:
Thanks @RobertKozak !

[v17.9.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.9.0) has been just released.

local.tf
```diff
@@ -96,6 +96,7 @@ locals {
     additional_ebs_volumes            = []                          # A list of additional volumes to be attached to the instances on this Auto Scaling group. Each volume should be an object with the following: block_device_name (required), volume_size, volume_type, iops, throughput, encrypted, kms_key_id (only on launch-template), delete_on_termination. Optional values are grabbed from root volume or from defaults
     additional_instance_store_volumes = []                          # A list of additional instance store (local disk) volumes to be attached to the instances on this Auto Scaling group. Each volume should be an object with the following: block_device_name (required), virtual_name.
     warm_pool                         = null                        # If this block is configured, add a Warm Pool to the specified Auto Scaling group.
+    timeouts                          = {}                          # A map of timeouts for create/update/delete operations
```
csschwe, line 99: should this have been:
`timeouts                          = var.timeouts`

The `timeouts` variable in the variable.tf file does not seem to actually be used during my testing

daroga0002:
> This merge was basically reverted by #1583

I still see this https://github.com/RobertKozak/terraform-aws-eks/blob/d46c49562771f247bf012031337968aef616332b/modules/node_groups/locals.tf#L27

so this feature is still available

PR #1560, opened 2021-09-01
feat: Supporting update_config in EKS managed node groups

### LINKED ISSUE #1554: Supporting `update_config` in EKS managed node groups
## What am I really asking for?

We need to use an AWS feature, I'd like to suggest a change to this terraform module, to enable it. I am looking for permission to share an implementation, after discussing API and any other concerns with you fine folks.

Setting update_config to either a max unavailable instances or percentage of instances has been supported since the latest release of the provider, as of the time of writing.

Rolling node groups with the current settings goes slowly as EKS only rolls one node at a time. We can sometimes tolerate more disruption, and will want to ensure this process goes faster

Some kind of option, when passing in parameters for node groups into the module, to set and update the update_config block settings

I've considered a tracheotomy, but that won't solve the issue.

## Additional context
I have a proof of concept implementation in a fork, that we can start using internally at $dayjob, here's a rough sense of how it works:

I want to set either percentage or a number iff the user of the module has passed in one of those parameters. I am using two parameters, one to indicate type, "number" or "percent", and the other to represent amount. This API could be rewritten, I just set it to this, to prove the concept.

The code adds two dynamic blocks to `./modules/node_groups/node_group.tf`:

```
  dynamic "update_config" { 
    for_each = each.value["max_unavailable_type"] == "number" ? toset([each.value["max_unavailable"]]) : toset([])
    content {
      max_unavailable = update_config.value
    }
  }

  dynamic "update_config" { 
    for_each = each.value["max_unavailable_type"] == "percent" ? toset([each.value["max_unavailable"]]) : toset([])
    content {
      max_unavailable_percentage = update_config.value
    }
  }
```

marianobilli:
# PR o'clock

Fixes #1554

## Description
Adding functionality proposed in [Supporting update_config in EKS managed node groups ](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1554)

### Checklist

### DIFF
--- a/examples/managed_node_groups/main.tf
+++ b/examples/managed_node_groups/main.tf
@@ -96,7 +96,7 @@ module "eks" {
         }
       ]
       update_config = {
-        max_unavailable_percentage = 50
+        max_unavailable_percentage = 50 # or set `max_unavailable`
       }
     }
   }
--- a/examples/managed_node_groups/main.tf
+++ b/examples/managed_node_groups/main.tf
@@ -95,6 +95,9 @@ module "eks" {
           effect = "NO_SCHEDULE"
         }
       ]
+      update_config = {
+        max_unavailable_percentage = 50
+      }
     }
   }
 
--- a/modules/node_groups/locals.tf
+++ b/modules/node_groups/locals.tf
@@ -24,6 +24,7 @@ locals {
       pre_userdata                  = var.workers_group_defaults["pre_userdata"]
       additional_security_group_ids = var.workers_group_defaults["additional_security_group_ids"]
       taints                        = []
+      timeouts                      = var.workers_group_defaults["timeouts"]
       update_default_version        = true
       ebs_optimized                 = null
     },
--- a/modules/node_groups/locals.tf
+++ b/modules/node_groups/locals.tf
@@ -26,7 +26,6 @@ locals {
       taints                        = []
       update_default_version        = true
       ebs_optimized                 = null
-      update_config                 = tomap({ "max_unavailable" : 1 })
     },
     var.node_groups_defaults,
     v,
--- a/modules/node_groups/locals.tf
+++ b/modules/node_groups/locals.tf
@@ -26,6 +26,7 @@ locals {
       taints                        = []
       update_default_version        = true
       ebs_optimized                 = null
+      update_config                 = tomap({ "max_unavailable" : 1 })
     },
     var.node_groups_defaults,
     v,
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -69,6 +69,11 @@ resource "aws_eks_node_group" "workers" {
     }
   }
 
+  update_config {
+    max_unavailable_percentage = lookup(each.value.update_config, "max_unavailable_percentage", null)
+    max_unavailable            = lookup(each.value.update_config, "max_unavailable", null)
+  }
+
   version = lookup(each.value, "version", null)
 
   labels = merge(
--- a/modules/node_groups/node_groups.tf
+++ b/modules/node_groups/node_groups.tf
@@ -78,6 +78,12 @@ resource "aws_eks_node_group" "workers" {
     }
   }
 
+  timeouts {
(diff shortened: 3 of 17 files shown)


modules/node_groups/node_groups.tf
```diff
@@ -69,6 +69,20 @@ resource "aws_eks_node_group" "workers" {
     }
   }
 
+  dynamic "update_config" {
```
antonbabenko, line 72: Please combine these two dynamic blocks into one.

examples/managed_node_groups/main.tf
```diff
@@ -95,6 +95,8 @@ module "eks" {
           effect = "NO_SCHEDULE"
         }
       ]
+      max_unavailable_type = "percent"
+      max_unavailable = 50
```
antonbabenko, line 99: Make sure to run `pre-commit run -a` to fix formatting and documentation.

antonbabenko (changes requested):
Looks good, but minor things are left.

modules/node_groups/node_groups.tf
marianobilli: Done @antonbabenko. 
Please checked how I did it, I believe the code is much simpler than the original proposed solution. 
The idea is that by default if you dont add this block on the tf resource it still creates an update_config by default with value of 1.

Tested scenarios:
- max_unavailable_percentage in node_groups_defaults
- max_unavailable in node_groups_defaults
- max_unavailable_percentage in specific node_group (without node_groups_defaults)
- max_unavailable in specific node_group (without node_groups_defaults)
- max_unavailable_percentage in node_groups_defaults with max_unavailable_percentage overwrite in specific node_group 
- max_unavailable_percentage in node_groups_defaults with max_unavailable_percentage overwrite in specific node_group 
- max_unavailable in node_groups_defaults with max_unavailable overwrite in specific node_group 
- max_unavailable in node_groups_defaults with max_unavailable overwrite in specific node_group

modules/node_groups/locals.tf
```diff
@@ -26,6 +26,7 @@ locals {
       taints                        = []
       update_default_version        = true
       ebs_optimized                 = null
+      update_config                 = tomap({ "max_unavailable" : 1 })
```
antonbabenko, line 29: I am not sure that this is the current behavior for existing users (also, please remove `tomap()`).

The goal is to introduce new features into the module without breaking anything or triggering updates to existing infrastructure for current users.

This should be verified like this:
1. Bring cluster up using the code in `examples` from `master` before this PR
2. Add the changes from this PR
3. Run `terraform plan` and there should be no changes proposed (especially, no destructive changes)

Please provide the final `terraform plan`.

marianobilli: @antonbabenko  done. Please find attached two plans.
1) As you requested the post-change plan overt the infra created pre-change showing there is no change
2) A post-change plan using the parameters I added to the example showing that update_config will be applied

[tfplans.zip](https://github.com/terraform-aws-modules/terraform-aws-eks/files/7098455/tfplans.zip)

antonbabenko:
Yes, I will fix that later tonight or tomorrow.

antonbabenko:
Thanks, @marianobilli for the PR. I have simplified code a bit and verified it locally.

[v17.10.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.10.0) has been just released.

marianobilli:
> Thanks, @marianobilli for the PR. I have simplified code a bit and verified it locally.
> 
> [v17.10.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.10.0) has been just released.

Thank you @antonbabenko ! I was not sure about using lookup() vs try(), and also i didnt know i could put two tries inside the same try() function. So I learnt something new. Thanks again!

antonbabenko:
Since we agreed to increase the minimum supported version of Terraform to 0.13 across all `terraform-aws-modules`, we can use `try()` and `depends_on` in modules (sometimes helpful in examples). It is a bit more elegant than `lookup()` in most places.

PR #1562, opened 2021-09-02
feat: Add add metadata options to managed node groups launch template

xr0n1ck:
# PR o'clock

## Description

Please explain the changes you made here and link to any relevant issues.

### Checklist

- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #1485, opened 2021-07-16
feat: Add metadata_options for node_groups

maxbrunet:
# PR o'clock

## Description

Add `metadata_options` in custom `launch_template` for managed node groups.

Useful for: https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html

### Checklist

- [x] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #1564, opened 2021-09-02
feat: Add metadata_options for node_groups

xr0n1ck:
# PR o'clock

## Description

Add metadata_options in custom launch_template for managed node groups.

Useful for: https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html

### Checklist


modules/node_groups/launch_template.tf
```diff
@@ -75,6 +75,15 @@ resource "aws_launch_template" "workers" {
 
   key_name = lookup(each.value, "key_name", null)
 
+  dynamic "metadata_options" {
+    for_each = lookup(each.value, "metadata_options", null) != null ? [lookup(each.value, "metadata_options", null)] : []
+    content {
+      http_endpoint               = lookup(metadata_options.value, "http_endpoint", null)
+      http_tokens                 = lookup(metadata_options.value, "http_tokens", null)
+      http_put_response_hop_limit = lookup(metadata_options.value, "http_put_response_hop_limit", null)
```
maxbrunet, line 83: Since the defaults are primitive variables, you need to specify them in the lookups or do a merge somewhere if you want to use an object

```hcl
http_endpoint               = lookup(metadata_options.value, "http_endpoint", var.workers_group_defaults["metadata_http_endpoint"])
http_tokens                 = lookup(metadata_options.value, "http_tokens", var.workers_group_defaults["metadata_http_tokens"])
http_put_response_hop_limit = lookup(metadata_options.value, "http_put_response_hop_limit", var.workers_group_defaults["metadata_http_put_response_hop_limit"])
```

or

```hcl
metadata_options = merge(
  {
     http_endpoint               = var.workers_group_defaults["metadata_http_endpoint"]
     http_tokens                 = var.workers_group_defaults["metadata_http_tokens"]
     http_put_response_hop_limit = var.workers_group_defaults["metadata_http_put_response_hop_limit"]
  },
  lookup(each.value, "metadata_options", {}),
)
```

maxbrunet (commented):
I have rebased PR #1485, but if you want to proceed with this one instead, here are a few comments.

Also in my opinion, although a `metadata_options` object could look nice, I would prefer remaining consistent with how self-managed launch_template/launch_configuration workers are configured and not use one

modules/node_groups/launch_template.tf
xr0n1ck: I use `dynamic "metadata_options"` due to https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1562#discussion_r700916693, and I missed about `workers_group_defaults` are primitives. It's my bad.
But looks like here is two ways:
1) Do not use dynamic block and use defaults (as was at https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1564/commits/0873af5bcf4ca7ca57d3877caadefa77654939ae). I don't see the point of using dynamic in this case, because with _defaults_ we always have all of 3 values for `metadata_options`. I think it's bad idea to merge _defaults_, only if at least one of 3 inputs are present, and use provider default behavior if not.
2) Do not use _defaults_ for `metadata_options`. Here we can use object as an input, `dynamic "metadata_options"` block,  and create it only if an input is not empty.

xr0n1ck: As for me - I prefer 1.

xr0n1ck:
@antonbabenko 
Is there any chance to merge this PR or https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1485 instead?

modules/node_groups/launch_template.tf
antonbabenko: I agree, let's close this one, and merge #1485 instead.

antonbabenko:
Continue in #1485

PR #1565, opened 2021-09-02
feat: Add Timeout to aws_autoscaling_group - based on #1552

filintod:
# PR o'clock

## Description

Autoscaling groups timeout is not exposed and sometimes the deletion of an ASG can take more than the default 10m.  This is based on top of PR #1552 to reuse the timeout variables

### Checklist


modules/node_groups/locals.tf
```diff
@@ -24,6 +24,7 @@ locals {
       pre_userdata                  = var.workers_group_defaults["pre_userdata"]
       additional_security_group_ids = var.workers_group_defaults["additional_security_group_ids"]
       taints                        = []
+      timeouts                      = var.workers_group_defaults["timeouts"]
```
csschwe, line 27: `var.workers_group_defaults["timeouts"]` is not defined anywhere so this is always null

local.tf
```diff
@@ -95,6 +95,7 @@ locals {
     additional_ebs_volumes            = []                          # A list of additional volumes to be attached to the instances on this Auto Scaling group. Each volume should be an object with the following: block_device_name (required), volume_size, volume_type, iops, encrypted, kms_key_id (only on launch-template), delete_on_termination. Optional values are grabbed from root volume or from defaults
     additional_instance_store_volumes = []                          # A list of additional instance store (local disk) volumes to be attached to the instances on this Auto Scaling group. Each volume should be an object with the following: block_device_name (required), virtual_name.
     warm_pool                         = null                        # If this block is configured, add a Warm Pool to the specified Auto Scaling group.
+    timeouts                          = {}                          # A map of timeouts for create/update/delete operations
```
csschwe, line 98: should this be something like 
`timeouts                          = var.timeouts`

csschwe: only asking because #1552 created a `timeouts` variable in the variables.rb

filintod: yeah, I did closed this and open this other one https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1566, I did not reuse that variable but probably can add it as last default value

PR #1566, opened 2021-09-02
feat: add delete timeout to aws_autoscaling_group

filintod:
# PR o'clock

## Description

Adding timeout to asg deletion as sometimes it can take more than the default 10m in our case. 

Did not find where to add to README so added to changelog

### Checklist


daroga0002:
we should address this issue in same way as in node groups: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/253f927465ccac36bcadefcfb1fd0e39b6244d73/modules/node_groups/main.tf#L81-L85

to make a consistent user experience.

@filintod do you will be able to make this change?

filintod:
I can do it but next week if that works.

daroga0002:
@filintod do you plan to implement  changes?

filintod:
@daroga0002 sorry for the delay. updated the timeout to follow node groups format

daroga0002:
> @daroga0002 sorry for the delay. updated the timeout to follow node groups format

could you paste here some example of module configuration (this will make a testing faster)?

filintod:
@daroga0002 an example would be like:

```
module "eks" {
  source           = "terraform-aws-modules/eks/aws"
  // other configs....

  workers_group_defaults = {
    asg_force_delete = true

    # double delete timeout default value
    timeout = {
      delete = "20m"
    }
  }
}
```

workers.tf
```diff
@@ -103,7 +103,9 @@ resource "aws_autoscaling_group" "workers" {
     "capacity_rebalance",
     local.workers_group_defaults["capacity_rebalance"]
   )
-
+  timeouts {
+    delete = lookup(local.workers_group_defaults["timeouts"], "delete", null)
```
daroga0002, line 107: this should be customized via `workers_group_defaults["timeouts"]` but also via `timeouts` setting per particulal worker definition as rest of parameters which are defined here https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/locals.tf

filintod: added an extra lookup against the per particular worker definition

daroga0002: This will work, but in general this is in other way as other variables. Please review and use same pattern which we arleady using

antonbabenko:

PR #1576, opened 2021-09-08
fix: Fixed coalescelist() with subnets in fargate module

### LINKED ISSUE #1575: Addition of coalescelist in v17.15.0 breaks create_eks option
## Description

Line 11 of fargate.tf was changed on 6th September for "Added ability to pass different subnets for fargate and the cluster" using a coalescelist.

Coalescelist works only if there is at least one variable in the lists to coalesce.  This excludes it from being passed variables with nothing in them when create_eks is set to False, as the coalescelist requires calculation at plan, and errors.

create_eks is used as a flag to disable the creation of EKS environments, as a work around to the more typical count.  However, this no longer works if variables for a subnet are not supplied.

 - Terraform:
Terraform v1.0.3
on darwin_amd64
 - Provider(s):

Terraform v1.0.3
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v3.57.0
+ provider registry.terraform.io/hashicorp/helm v2.3.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.4.1
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/newrelic/newrelic v2.7.5
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1

 - Module:
terraform-aws-modules/eks/aws 17.15.0 affected, 17.14.0 is not.

## Reproduction

Build two terraform files which use the eks module to provision an EKS cluster.  Set one of them to create_eks = False and give an empty list for the subnets and fargate_subnets (which defaults to an empty list).  Set the other with a populated subnet list.

Running a plan or apply, watch it error with: 
Error: Error in function call │ │ on .terraform/modules/eks_cluster_ap_southeast_1.eks/fargate.tf line 11, in module "fargate": │ 11: subnets = coalescelist(var.fargate_subnets, var.subnets) │ ├──────────────── │ │ var.fargate_subnets is empty list of string │ │ var.subnets is empty list of string │ │ Call to function "coalescelist" failed: no non-null arguments.

Yes.  I have switched to the appropriate workspace

Yes, cleared local cache.

### Code Snippet to Reproduce

module "eks" {
  source     = "terraform-aws-modules/eks/aws"
  create_eks = var.create_eks

  cluster_name     = var.cluster_name
  cluster_version  = var.cluster_version
  enable_irsa      = true
  vpc_id           = one(data.aws_vpc.eks_vpc.*.id)
  subnets          = flatten(data.aws_subnet_ids.worker_subnets.*.ids)
  write_kubeconfig = false
(removed the rest)

You can see that we are using a lookup in order to get the subnets for each region.  The lookup is controlled by the create_eks variable as well (see below), so with that set to false, there's nothing which is sent through.

data "aws_vpc" "eks_vpc" {
  count = local.create_eks_count
  tags = {
    Name = var.vpc_name
  }
}

data "aws_subnet_ids" "worker_subnets" {
  count  = local.create_eks_count
  vpc_id = one(data.aws_vpc.eks_vpc.*.id)
  tags = {
    Tier = "Private"
  }
}

## Expected behavior
Prior to v17.15.0, the create_eks flag would stop an EKS environment from being created as a substitute for the count option which is not viable as per the module's documentation.  By supplying an empty variable for the subnet, terraform would continue to run, as it does not need to create anything.

## Actual behavior
v17.15.0 causes terraform to exit with the error 'Call to function "coalescelist" failed: no non-null arguments.' even when create_eks is set to False.  So one must supply a subnet even if the EKS module is not expected to be created.

Pinning the module version to v17.14.0 results in the error message going away and a successful deployment.

Error: Error in function call
│
│   on .terraform/modules/eks_cluster_ap_southeast_1.eks/fargate.tf line 11, in module "fargate":
│   11:   subnets                           = coalescelist(var.fargate_subnets, var.subnets)
│     ├────────────────
│     │ var.fargate_subnets is empty list of string
│     │ var.subnets is empty list of string
│
│ Call to function "coalescelist" failed: no non-null arguments.

## Additional context
The documentation does state that the input is required, but it defeats the purpose of the create_eks variable if subnets are required for regions where you aren't actually wanting to create an EKS cluster.

antonbabenko:
Resolves #1575

### DIFF
--- a/fargate.tf
+++ b/fargate.tf
@@ -8,7 +8,7 @@ module "fargate" {
   permissions_boundary              = var.permissions_boundary
   iam_path                          = var.iam_path
   iam_policy_arn_prefix             = local.policy_arn_prefix
-  subnets                           = coalescelist(var.fargate_subnets, var.subnets)
+  subnets                           = coalescelist(var.fargate_subnets, var.subnets, [""])
   tags                              = var.tags
 
   # Hack to ensure ordering of resource creation.


antonbabenko:
[v17.16.0](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/v17.16.0) has been just released.

PR #1271, opened 2021-03-10
fix: SSRF Security Issue with IMDS

oed-mertenss:
# PR o'clock

## Description

I tried to use your Module but there is a security issue reported by terrascan:
```
	Description    :	Launch configuration uses IMDSv1 which vulnerable to SSRF
	File           :	../../../../../tmp/jqpxh5/workers.tf
	Line           :	162
	Severity       :	HIGH
```
https://www.accurics.com/blog/security-blog/aws-cloud-security-protect-ssrf/

PR #1579, opened 2021-09-09
feat: Change IDMS metadata defaults to more strict

lisfo4ka:
# PR o'clock

## Description

This will close #1271.
Defaults were set according to the [Amazon EKS security best practices](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html )

### Checklist

### DIFF
--- a/local.tf
+++ b/local.tf
@@ -112,8 +112,8 @@ locals {
     cpu_credits                          = "standard"                                                          # T2/T3 unlimited mode, can be 'standard' or 'unlimited'. Used 'standard' mode as default to avoid paying higher costs
     market_type                          = null
     metadata_http_endpoint               = "enabled"  # The state of the metadata service: enabled, disabled.
-    metadata_http_tokens                 = "optional" # If session tokens are required: optional, required.
-    metadata_http_put_response_hop_limit = null       # The desired HTTP PUT response hop limit for instance metadata requests.
+    metadata_http_tokens                 = "required" # If session tokens are required: optional, required.
+    metadata_http_put_response_hop_limit = 1          # The desired HTTP PUT response hop limit for instance metadata requests.
     # Settings for launch templates with mixed instances policy
     override_instance_types                  = ["m5.large", "m5a.large", "m5d.large", "m5ad.large"] # A list of override instance types for mixed instances policy
     on_demand_allocation_strategy            = null                                                 # Strategy to use when launching on-demand instances. Valid values: prioritized.


local.tf
```diff
@@ -112,8 +112,8 @@ locals {
     cpu_credits                          = "standard"                                                          # T2/T3 unlimited mode, can be 'standard' or 'unlimited'. Used 'standard' mode as default to avoid paying higher costs
     market_type                          = null
     metadata_http_endpoint               = "enabled"  # The state of the metadata service: enabled, disabled.
-    metadata_http_tokens                 = "optional" # If session tokens are required: optional, required.
-    metadata_http_put_response_hop_limit = null       # The desired HTTP PUT response hop limit for instance metadata requests.
+    metadata_http_tokens                 = "required" # If session tokens are required: optional, required.
+    metadata_http_put_response_hop_limit = 1          # The desired HTTP PUT response hop limit for instance metadata requests.
```
antonbabenko, line 116: These changes will trigger an update for existing users unless they provide these values already.

I propose we group such changes for the next major release and make it visible in the upgrade guide.

WDYT?

lisfo4ka: Yes, sure, fully agree.

antonbabenko:
Added this one to milestone v18.0.0 - https://github.com/terraform-aws-modules/terraform-aws-eks/milestone/5

@daroga0002 @lisfo4ka If you find other PRs which introduce breaking changes, let's add them there.

daroga0002:
@lisfo4ka  please rebase

jaimehrubiks:
Note that I tested this yesterday by manually setting "required", and EBS and EFS CSI Provisioner helm charts stopped working because they could not read metadata.

In my case I don't mind if you set it because it is very easy to change to "optional", but wanted you to know that this is giving me important issues.

ArchiFleKs:
This is a good idea to be secure by default I might be wrong but I think `metadata_http_put_response_hop_limit` blocks metadata access for all the pods except the one using host networking right ?

It means user should use IRSA for all the component. I'm not sure you can retrieve region or vpc id or other information without querying metadata service, that mean adding them manually I guess. For example: https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html#deploy-lb-controller

I can also think of aws ebs csi driver which also rely on metadata, and not sure if the csi-node can function without metadata access (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/474). Anyway this is good but will require some adaptation on the user side to be able to upgrade to this major release

daroga0002:
> It means user should use IRSA for all the component.

exactly as per https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

Because:
```
Even if you assign an IAM role to a Kubernetes service account, the pod still also has the permissions assigned to the Amazon EKS node IAM role, unless you block pod access to the IMDS. For more information, see Restricting access to the IMDS and Amazon EC2 instance profile credentials.
```

lisfo4ka:
Hi folks,

Investigating a little bit deeper I suggest increasing TTL (`metadata_http_put_response_hop_limit`) to 2, since 1 is really incompatible with containerised applications on Kubernetes that run in a separate network namespace from the instance.
I know that it can be figured out by using host networking for the pod, but I'm not sure if it's a good recommendation to the community.

Moreover, AWS EKS best practices admit [increasing the hop limit on EC2 instances to 2](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#when-your-application-needs-access-to-idms-use-imdsv2-and-increase-the-hop-limit-on-ec2-instances-to-2).

Even more, official CloudFormation templates for EKS cluster deployment are going [to change the hop limit to 2](https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/) as well. It seems [not yet](https://github.com/aws-quickstart/quickstart-amazon-eks/blob/71e0dd6b908dc1050d66c73f38aa7c41a49a922a/templates/amazon-eks-advanced-configuration.template.yaml#L418)

And of course, we can remind in a Changelog that after restricting access to the instance metadata the pod can't inherit node IAM profile anymore, so the appropriate IAM roles for application service accounts should be provided.
For instance, the first deployment which should be taken into account is Daemonset `aws-node` that is responsible for AWS CNI plugin work and requires `AmazonEKS_CNI_Policy`. The details for configuring IRSA for the CNI plugin can be found here [cni-iam-role](https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html).
The same should be considered for any deployments that rely on instance IAM profile token, e.g. node termination handler, EBS and EFS CSI drivers etc.

And just let me clarify that the parameter `http_tokens = "required" `doesn't block IMDS, but force to use IMDSv2 instead of IMDSv1, which provides improvement for the security and adds an additional defence in depth layer. More details - [Add defense in depth against open firewalls, reverse proxies, and SSRF vulnerabilities with enhancements to the EC2 Instance Metadata Service](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/)

Also, we can leave a suggestion to configure the CloudWatch dashboard based on the `MetadataNoToken` metric.
It'll help users to ensure that no instances are using IMDSv1 before disabling it altogether.

antonbabenko:

PR #1580, opened 2021-09-10
feat: Fix custom AMI bootstrap

stevehipwell:
# PR o'clock

## Description

This PR fixes the issues introduced in PR https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1473 which would have made it impossible to start a fully customised image without requiring a `bootstrap.sh` script.

### Checklist


ArchiFleKs:
Looks good. I’m off until Tuesday I don’t have the time to test it. 

My bad but the “role” was a custom label added by me and not something added by AWS. I think it can be removed

stevehipwell:
Thanks @ArchiFleKs, I did wonder about the role...

ArchiFleKs:
I think `topology.ebs.csi...`  can be ommited also as it was also added by me for other reason. What about these EKS labels ? :

* `eks.amazonaws.com/sourceLaunchTemplateVersion`
* `eks.amazonaws.com/nodegroup`
* `eks.amazonaws.com/sourceLaunchTemplateId`

AndreyBar:
Hi. This is the author of #1473. Could you please elaborate on your PR, I want to understand the problem it's addressing?
If I understand correctly, your PR allows users to ignore `bootstrap.sh` if they have some custom AMI that doesn't have such script, and provide bootstrap logic directly in user data?

stevehipwell:
@ArchiFleKs I did wonder when they started adding the CSI labels, we add them ourselves until the CSI driver starts using the correct topology labels. The other labels you list would be impossible to inject as they're added to the calculated launch template that's created after the MNG is created.

@AndreyBar your PR only supported custom AMIs built from the EKS optimised image or that have added a _/etc/eks/bootstrtap.sh_ script to mimic the behaviour of one. The call to _/etc/eks/bootstrtap.sh_ wouldn't work correctly without providing custom user data and any labels set would be ignored. This PR corrects this behaviour so if you just want to use a fixed AMI you can do so without needing to understand how _/etc/eks/bootstrtap.sh_ works.

modules/node_groups/templates/userdata.sh.tpl
```diff
@@ -1,9 +1,20 @@
...
 
-# Allow user supplied pre userdata code
+sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" ${kubelet_extra_args}"' /etc/eks/bootstrap.sh
+%{endif ~}
+%{if length(pre_userdata) > 0 ~}
+
+# User supplied pre userdata code
 ${pre_userdata}
+%{endif ~}
+%{ if length(ami_id) > 0 && ami_is_eks_optimized ~}
```
endrec, line 11: If think this should be 
```suggestion
%{ if length(ami_id) > 0 && ami_is_eks_optimized == false ~}
```

Shouldn't it?

stevehipwell: No, what's in the PR is correct. We want to call the bootstrap script if the `ami_id` is set and the AMI is an EKS optimised image. If the `ami_id` is set in a launch template then the launch template is used directly and not merged to create a managed launch template; in this case we can call `/etc/eks/bootstrtap.sh` ourselves if the image is an EKS optimised one with the script available.

endrec: Oh, okay, so we need to add this even if we use an AWS supplied AMI here. I did not realised that.

stevehipwell: If you set `ami_id` you need this.

stevehipwell:
@antonbabenko could you take a look at this PR?

antonbabenko:
Sorry, I don't have the capacity to take a look at this one for real (run tests, think about edge cases, compatibility, etc).

Maybe @daroga0002 can put this in his queue for reviews?

daroga0002:
> Maybe @daroga0002 can put this in his queue for reviews?

Yup, I have it on TODO list

daroga0002:
@stevehipwell sorry for delay but I was sick last days, I will find some time during weekend for this but could you rebase this PR to make it easier for testing.

stevehipwell:
@daroga0002 I hope you're feeling better now? I've rebased.

daroga0002 (approved):

daroga0002:
@stevehipwell thank you for your contribution 🎉  , I reviewed PR and I think we can merge it as this should solve multiple issues with custom AMI managed groups.

daroga0002:
@antonbabenko  lets merge this and wait with releasing

modules/node_groups/templates/userdata.sh.tpl
```diff
@@ -1,9 +1,20 @@
 #!/bin/bash -e
+%{ if length(ami_id) == 0 || ami_is_eks_optimized ~}
 
-# Allow user supplied pre userdata code
+sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" ${kubelet_extra_args}"' /etc/eks/bootstrap.sh
+%{endif ~}
+%{if length(pre_userdata) > 0 ~}
```
antonbabenko, line 6: I don't think we need to have this `if/endif` block but just include `${pre_userdata}` always as it was before.

antonbabenko (commented):
There is a minor comment about shell script. The rest looks good to me.

modules/node_groups/templates/userdata.sh.tpl
stevehipwell: It would work without the if block, but it's tidier to use it so the resulting script has consistent white space. It also makes it clear in the template that it's an optional variable.

modules/node_groups/templates/userdata.sh.tpl
```diff
@@ -1,9 +1,20 @@
...
+%{ if length(ami_id) > 0 && ami_is_eks_optimized ~}
 
-sed -i '/^KUBELET_EXTRA_ARGS=/a KUBELET_EXTRA_ARGS+=" ${kubelet_extra_args}"' /etc/eks/bootstrap.sh
-%{ if run_bootstrap_script }
-    /etc/eks/bootstrap.sh ${cluster_name}
-%{ endif }
+# Set variables
+API_SERVER_URL=${cluster_endpoint}
+B64_CLUSTER_CA=${cluster_ca}
+K8S_CLUSTER_DNS_IP=172.20.0.10
```
daroga0002, line 16: I found one issue as this should be variable https://github.com/terraform-aws-modules/terraform-aws-eks/blob/253f927465ccac36bcadefcfb1fd0e39b6244d73/main.tf#L28

stevehipwell: @daroga0002 I think you're right. I'm currently on vacation, so can't make this change until I'm back next week. If you can modify my PR that's fine. Otherwise setting the optimised variable to false would let this to be set by providing the whole user data, I could then update the other PR to make this customisable as it covers this type of behaviour?

daroga0002: np, we can wait, have a great vacations 😄

stevehipwell: @daroga0002 this should be ready to merge.

daroga0002: @stevehipwell are you sure you added change as I still see 
image

stevehipwell: @daroga002 sorry, manic day in the office. I only rebased, I'll add the changes first thing in the morning.

stevehipwell: @daroga0002 I've updated the userdata template to be simpler and so that all custom AMIs use the same initial logic for setting the variables (this is designed ready to support #1577). After checking the _bootstrap.sh_ code I've removed the manual setting of `K8S_CLUSTER_DNS_IP` as _bootstrap.sh_ will handle this correctly; #1577 will enable this to be fully customisable.

endrec:
@JordyWTC This PR was merged _after_ `v17.20.0`, so the issue you are seeing is probably from an earlier change.

(The last commit on `v17.20.0` was 17 Sep, this Pr was merged on 8 Oct.)

stevehipwell:
@JordyWTC that is exactly how it's currently working.

PR #1581, opened 2021-09-10
feat: Make the tagging of IAM Policy and Profile optional

zoltan-fedor:
# PR o'clock

## Description

This is to allow the disabling of the tagging of the IAM Policy and Profile created.
This is necessary for those whose IAM user is not allowed to tag these resources - which is typical in enterprise environment.

### Checklist

- [ x ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ Apache 2 Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraf
 | <a name="input_permissions_boundary"></a> [permissions\_boundary](#input\_permissions\_boundary) | If provided, all IAM roles will be created with this permissions boundary attached. | `string` | `null` | no |
 | <a name="input_subnets"></a> [subnets](#input\_subnets) | A list of subnets to place the EKS cluster and workers within. | `list(string)` | n/a | yes |
 | <a name="input_tags"></a> [tags](#input\_tags) | A map of tags to add to all resources. Tags added to launch configuration or templates override these values for ASG Tags only. | `map(string)` | `{}` | no |
+| <a name="input_tag_iam_policy_and_profile"></a> [tag_iam_policy_and_profile](#input\_tag_iam_policy_and_profile) | Whether to apply the tags to the IAM Policy and Profile created | `bool` | `true` | no |
 | <a name="input_timeouts"></a> [timeouts](#input\_timeouts) | A map of timeouts for create/update/delete operations. | `map(string)` | `{}` | no |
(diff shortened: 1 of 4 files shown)


daroga0002:
Could you elaborate why cannot tag IAM roles and can tag everything else?

Tagging is rather best practice approach, usually if Enterprises have strict rules/requirements regarding roles and iam things then we allow in this module to pass everywhere `externally created` roles

zoltan-fedor:
See Note at the bottom of https://docs.aws.amazon.com/iot/latest/developerguide/tagging-iot-iam.html?utm_source=pocket_mylist

_"If you allow or deny users access to resources based on tags, you must consider explicitly denying users the ability to add those tags to or remove them from the same resources. Otherwise, it's possible for a user to circumvent your restrictions and gain access to a resource by modifying its tags."_

daroga0002:
Just thinking here about this approach and I don`t undestand your security layer here. 

Dont get how IAM conditional policies using tags will adding to your Enterprise security as you still have privileges to create policies and roles (I am just assuming as you want create those objects via module)?

As I am able to imagine that as you can create roles and policies in IAM then you can just create * policy and then escalate your own privileges on AWS account.


Correct me if I am wrong

zoltan-fedor:
See Note at the bottom of https://docs.aws.amazon.com/iot/latest/developerguide/tagging-iot-iam.html?utm_source=pocket_mylist

_"If you allow or deny users access to resources based on tags, you must consider explicitly denying users the ability to add those tags to or remove them from the same resources. Otherwise, it's possible for a user to circumvent your restrictions and gain access to a resource by modifying its tags."_

>> Based on this you can allow the creation of IAM policies, but not the tagging of them in case if you have some sensitive policies which you want to control access to. Which is why it reasonable to have a scenario where the user has rights to create IAM policies, but not to tag them (as that would require both the rights to create and modify tags - which would in turn allow them to circumvent tag-based restrictions).

Also tagging is exactly the one which can prevent the scenario you described: "As I am able to imagine that as you can create roles and policies in IAM then you can just create * policy and then escalate your own privileges on AWS account."

I am sure that this is a fairly rare scenario, so I understand if you decide not to include this option.
Those few who are affected can always just use my fork instead.

Thanks

daroga0002:
Thx @zoltan-fedor, if you are comfortable to forking then it will be maybe better. 

In general we are trying to follow here some main best practices from AWS and EKS side, module arleady is very heavy and we want to avoid a customizations which will be used by small amount of people.

I am closing this PR for now, but thank you for contribution :+1:

zoltan-fedor:
No worries, I am comfortable with forking. I understand.

Thanks for considering!

Issue #1530, opened 2021-08-12
Launch template for node_groups is ignored


v1nc3nt27:
## Description

I'm using managed node_groups and since there is no parameter to specify the cpu_credits, I'm adding a launch template to the node group. However, the actual nodes still launch in "unlimited" mode and for the node_group another launch template is created starting with "eks-" prefix. The automatically created launch template does not contain my launch template setting.

UPDATE: I'm using a launch template for a g4dn instance in which I specify a specific AMI which is definitely used, so the it seems like it's just the credit_specification parameter that is being ignored.

## Versions
 - Terraform: v1.0.4
 - Provider(s):
terraform-aws-eks v17.1.0

### Code Snippet to Reproduce
```
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "17.1.0"#"15.1.0"

  cluster_name    = "${var.cluster_name}"
  cluster_version = "1.20"
  subnets         = module.vpc.private_subnets
  
	node_groups = {
	    cpu-worker-core-0 = {
	      node_group_name     = "cpu-worker-0"
	      desired_capacity    = 2
	      max_capacity        = 4
	      min_capacity        = 0
	      capacity_type       = "ON_DEMAND"
	      subnet_ids          = module.vpc.private_subnets
	      instance_types = ["t3.small"]
	      launch_template_id      = aws_launch_template.cpu-worker-full-on-demand.id
	      launch_template_version = aws_launch_template.cpu-worker-full-on-demand.default_version
	    },
	    ...
	}
}
```
```
resource "aws_launch_template" "cpu-worker-full-on-demand" {
  name_prefix            = "${var.cluster_name}-"
  description            = "CPU 100% on-demand Launch-Template"
  update_default_version = true

  block_device_mappings {
    device_name = "/dev/xvda"

    ebs {
      volume_size           = 50
      volume_type           = "gp2"
      delete_on_termination = true
    }
  }
  credit_specification {
    cpu_credits = "standard"
  }

  monitoring {
    enabled = false
  }

  network_interfaces {
    associate_public_ip_address = false
    delete_on_termination       = true
    security_groups             = [
      module.eks.cluster_primary_security_group_id
      ]
  }


  lifecycle {
    create_before_destroy = true
  }

}
```

## Expected behavior
My launch template or at least its settings are taken.

## Actual behavior
The automatically created launch template is used.

aidan-melen:
[...]

PR #1589, opened 2021-09-16
feat: optional cpu_credits support added for node_groups submodule

DovnarAlexander:
# PR o'clock

## Description

cpu_credits is not supported for managed node groups with launch_template creation option.
This PR adds this optional support (and only works if any not-null `cpu_credits` value is provided.

Potentially can fix https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1530.

### Checklist

### DIFF
--- a/modules/node_groups/launch_template.tf
+++ b/modules/node_groups/launch_template.tf
@@ -47,6 +47,13 @@ resource "aws_launch_template" "workers" {
 
   instance_type = each.value["set_instance_types_on_lt"] ? element(each.value.instance_types, 0) : null
 
+  dynamic "credit_specification" {
+    for_each = lookup(each.value, "cpu_credits", null) != null ? [""] : []
+    content {
+      cpu_credits = each.value.cpu_credits
+    }
+  }
+
   monitoring {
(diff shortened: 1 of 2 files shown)


daroga0002:
does this really gets affected in EKS node gorup?

as from https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1530 looks it will not work

DovnarAlexander:
> does this really gets working in EKS node group?
> 
> as from #1530 looks it will not work

Hi. This case uses user-managed LT (and looks like the RC of the issue is hidden somewhere inside this LT). In that case I use capability to create it using the module. I've tested it for my case and it works as needed.

DovnarAlexander:
> > does this really gets working in EKS node group?
> > as from #1530 looks it will not work
> 
> Hi. This case uses user-managed LT (and looks like the RC of the issue is hidden somewhere inside this LT). In that PR I use capability to create it using the module. I've tested it for my node groups and it works as needed.

I've checked it one more. The issue above is correct. Resolved temporary by global ec2 setting changed on the AWS account level.

PR #1591, opened 2021-09-17
feat: define fargate pod execution role name

emanuelflp:
Set a custom name to "fargate pod execution role" if variable `var.fargate_pod_execution_role_name` is defined.

# PR o'clock

## Description

When creating fargate profiles, we can't define the name of `fargate pod execution role` as we can do to other IAM Roles.

### Checklist


claytonsilva (approved):

daroga0002 (commented):
This require also a update in variables.tf as currently it states:
```
"The IAM Role that provides permissions for the EKS Fargate Profile."
```

PR #1592, opened 2021-09-17
fix: clarify usage of cluster endpoint public CIDR list var

### LINKED ISSUE #1111: Docs should clarify private endpoint need to be true when public endpoint is IP filtered
# I have issues

## I'm submitting a...

* [x] bug report
* [ ] feature request
* [x] kudos, thank you, warm fuzzy

Use the managed node group example and override `cluster_endpoint_public_access_cidrs` to be `[YOUR_WAN_IP/32]` but do not set `cluster_endpoint_private_access`

terraform apply: after 15 min to create EKS cluster, the node group will loop waiting for 25 minutes when TF gives up, with the following error: 
```
Error: error waiting for EKS Node Group (eks-experiment-3:eks-experiment-3-main-wise-gobbler) creation: NodeCreationFailure: Instances failed to join the kubernetes cluster. Resource IDs: [... ... ...]
```

I didn't realize until after a lot of pain trying to figure out causes that it was result of having `cluster_endpoint_public_access_cidrs = [MY_WAN_IP]`, but have only a public cluster endpoint: the nodes would then talk to the cluster through the public endpoint but their IP would not match the above list so could never join. 

So this might be more of an enhancement than a bug report: 
- I should have seen in the docs for `cluster_endpoint_public_access_cidrs` that in almost all cases where you override this value, you will need to set `cluster_endpoint_private_access` to true
- the `cluster_endpoint_private_access` should have defaulted to true since `cluster_endpoint_public_access_cidrs` was not the default value, how could it possibly work with anything but private=true? 

## Environment details

* Affected module version: 13.2.1
* OS: Ubuntu
* Terraform version: 0.13.5

## Any other relevant info

If you override the default of `cluster_endpoint_public_access_cidrs` (which is `[0.0.0.0]` by default), you are pretty much guaranteed, IMO, to need to activate the private endpoint. In fact, enabling nodes to use the private endpoint should probably be a default. What would be the downside? What benefit to have node-to-node traffic *ever* leave the cluster?

schollii:
# PR o'clock

## Description

Fixes #1111. 

If private access is false, then worker nodes have no choice but to access API server through the public access point, ie internet. 

But If you set public endpoint CIDR list, their IP will be blocked (you don't have a way to determine WAN IP of nodes). So worker nodes will never join the cluster. This caused me headache so I wanted to avoid this for others. 

So if you set public access = true and public SG CIDR list, you HAVE to enable private access too so the nodes can directly access the API server from within the VPC. The docs on this point in AWS are very easy to miss / not see / not remember.  

### Checklist

- [x ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation


README.md
```diff
@@ -214,7 +214,7 @@ Apache 2 Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraf
...
-| <a name="input_cluster_endpoint_public_access_cidrs"></a> [cluster\_endpoint\_public\_access\_cidrs](#input\_cluster\_endpoint\_public\_access\_cidrs) | List of CIDR blocks which can access the Amazon EKS public API server endpoint. | `list(string)` | <pre>[<br>  "0.0.0.0/0"<br>]</pre> | no |
+| <a name="input_cluster_endpoint_public_access_cidrs"></a> [cluster\_endpoint\_public\_access\_cidrs](#input\_cluster\_endpoint\_public\_access\_cidrs) | List of CIDR blocks which can access the Amazon EKS public API server endpoint. If you use this, also set `cluster_endpoint_private_access` to true otherwise the nodes will use public access point but their WAN IP will not be in the CIDR list (see _API server endpoint access options_ table in [EKS User Guide](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html) for details). | `list(string)` | <pre>[<br>  "0.0.0.0/0"<br>]</pre> | no |
```
antonbabenko, line 217: Please move your addition into [`notes` section](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/README.md#notes) in README.

The documentation for inputs is extracted from `variables.tf` and we try to exclude very detailed explanations and links to other docs or variables from the description of the variables.

antonbabenko (changes requested):

README.md
daroga0002: I think better place will be https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md#networking where we already have some info about it

schollii: OK, it certainly makes sense to move it there, I will do that shortly.

antonbabenko: I completely forgot that we have FAQ where such details should be written. :)

daroga0002:
I have opened https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1603 for same to fix it

daroga0002:
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1603 has been merged so this PR can be closed

PR #1603, opened 2021-09-28
docs: update faq about endpoints

### LINKED ISSUE #1111: Docs should clarify private endpoint need to be true when public endpoint is IP filtered

daroga0002:
# PR o'clock

## Description

Resolves #1111

Duplicate of https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1592

Additionally fixing some markdown formatting issues

### Checklist

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -12,7 +12,7 @@ Please open Issues or PRs if you think something is missing.
 
 Often caused by a networking or endpoint configuration issue.
 
-At least one of the cluster public or private endpoints must be enabled to access the cluster to work. If you require a public endpoint, setting up both (public and private) and restricting the public endpoint via setting `cluster_endpoint_public_access_cidrs` is recommended. More about communication with an endpoint is available under https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html
+At least one of the cluster public or private endpoints must be enabled to access the cluster to work. If you require a public endpoint, setting up both (public and private) and restricting the public endpoint via setting `cluster_endpoint_public_access_cidrs` is recommended. More about communication with an endpoint is available [here](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html).
 
 Nodes need to be able to contact the EKS cluster endpoint. By default, the module only creates a public endpoint. To access endpoint, the nodes need outgoing internet access:
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -12,15 +12,20 @@ Please open Issues or PRs if you think something is missing.
 
 Often caused by a networking or endpoint configuration issue.
 
-At least one of the cluster public or private endpoints must be enabled in order for access to the cluster to work.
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -12,15 +12,20 @@ Please open Issues or PRs if you think something is missing.
 
 Often caused by a networking or endpoint configuration issue.
 
-At least one of the cluster public or private endpoints must be enabled in order for access to the cluster to work.
+At least one of the cluster public or private endpoints must be enabled to access the cluster to work. If you require a public endpoint, setting up both (public and private) and restricting the public endpoint via setting `cluster_endpoint_public_access_cidrs` is recommended. More about communication with an endpoint is available under https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html
```
antonbabenko, line 15:
```suggestion
At least one of the cluster public or private endpoints must be enabled to access the cluster to work. If you require a public endpoint, setting up both (public and private) and restricting the public endpoint via setting `cluster_endpoint_public_access_cidrs` is recommended. More about communication with an endpoint is available [here](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html).
```

PR #1611, opened 2021-09-30
feat: add tag 'name' to iam roles

lazouz:
# PR o'clock

## Description

The aws_iam_role does not have a specific tag 'name' and sometimes it can be problematic

I only add a tag name corresponding to the iam_role name

### Checklist

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -138,7 +138,7 @@ resource "aws_iam_role" "cluster" {
   tags = merge(
     var.tags,
     {
-      "Name" = var.cluster_iam_role_name != "" ? null : "${var.cluster_name}-eks_worker_iam_role"
+      "Name" = var.cluster_iam_role_name != "" ? null : "${var.cluster_name}-eks_cluster_iam_role"
     },
   )
 }
--- a/main.tf
+++ b/main.tf
@@ -138,7 +138,7 @@ resource "aws_iam_role" "cluster" {
   tags = merge(
     var.tags,
     {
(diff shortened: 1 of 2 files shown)


main.tf
```diff
@@ -135,7 +135,12 @@ resource "aws_iam_role" "cluster" {
   path                  = var.iam_path
   force_detach_policies = true
 
-  tags = var.tags
+  tags = merge(
+    var.tags,
+    {
+      "Name" = var.cluster_iam_role_name != "" ? null : "${var.cluster_name}-eks_cluster_iam_role"
```
aperigault, line 141:
```suggestion
      "Name" = var.cluster_iam_role_name != "" ? var.cluster_iam_role_name : var.cluster_name
```

workers.tf
```diff
@@ -483,7 +483,12 @@ resource "aws_iam_role" "workers" {
   path                  = var.iam_path
   force_detach_policies = true
 
-  tags = var.tags
+  tags = merge(
+    var.tags,
+    {
+      "Name" = var.workers_role_name != "" ? null : "${var.cluster_name}-eks_worker_iam_role"
```
aperigault, line 489:
```suggestion
      "Name" = var.workers_role_name != "" ? var.workers_role_name : local.cluster_name
```

bryantbiggs:
we used to use this approach as some resources in AWS relied on the `Name` tag to set the name value in the UI (some still support this, but through the API this has become a first class property instead of relying on tags). therefore, I am with @daroga0002  and don't see the value of this at this time

PR #1625, opened 2021-10-06
Rebuild examples

### LINKED ISSUE #1622: Multiple examples are not working
## Description

Multiple examples are not working, goal should be that after using that example we receive fully working EKS cluster but now this is not achievable.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions
doesnt matter as this is generic issue

## Reproduction
run all examples from scratch and test

## Expected behavior
Examples should be working

## Actual behavior
some are not working because misconfiguration, example is managed node groups which create worker with taint so coredns deployment cannot be scheduled

daroga0002:
# PR o'clock

## Description

resolves #1622

This is major improvement in our examples. Some examples were broken or were not running correctly, after those fixes:
- all examples are working as should (I tested each of them)
- examples have generic README.md which can be updated more over time
- in IRSA example I automated helm autoscaler deployment in terraform (it was manual step which was error prone)
- fixed bottlerocket example where nodes were not registering to cluster (missing aws-auth configmap management)
- each example is now independent and has random suffix (so can be used multiple times on same AWS account)
- each example has `generic.tf` file which is same across all examples (creating vpc and other basic resources)
- we can override via variable in terraform region were example will be launched (`terraform apply -var=region=us-east-1`)
- core-dns is scheduled correctly in fargate example
- fixed variables description highlighted in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1623
- fixed some markdown linting error in main README.md
- bump some `versions.tf` in examples to allign with overall module
- removed `_bootstrap` example as it was created dependencies and in general didnt allow to run multiple examples in parallel

### Checklist


antonbabenko:
Good improvement overall!

The changes are rather big, and not all of them are in sync with the approach we have in other modules in this organization (e.g. different structures, file names, etc). I will try to find time to review it during this week.

Issue #1644, opened 2021-10-14
Allow alternative user_data format in node_groups


scalen:
## Is your request related to a new offering from AWS?
Yes (ish) - It relates to supporting the Bottlerocket AMI using `node_groups`.

## Is your request related to a problem? Please describe.
The `node_groups` module is more opinionated about the `user_data` for a node group than the `worker_groups` configuration was, forcing either:

* Use of `sh`-based cloud-init user data.
* Specification of the complete Launch Template outside of the module

Bottlerocket (a hardened AMI designed for use as a worker node in EKS) [uses raw TOML for its user data](https://github.com/bottlerocket-os/bottlerocket#using-user-data), not embedded in cloud-init or a shell.

As such, it is very easy to create an older `worker_group` based on Bottlerocket, but very labour intensive and brittle to create a newer `node_group`, even when wishing to reuse all of the other Launch Template features provided in this module.

## Describe the solution you'd like.
I would like to see the `userdata_template_file` and `userdata_template_extra_args` (or similar) module variables replicated from the parent module to the `node_groups` module.  This could easily be made to work with the `cloudinit_config` resource, which already does templating.  To be completely accurate, it should also be possible to set the MIME type of the resulting config file with something along the lines of `userdata_mime_type`, or to leave it unset when needing raw user data.

I would probably modify the shape of the above recommended variables to be the following, but this is just because I like nested configuration:
```
variable "user_data" {
  type    = map
  default = {
    template_file       = "${path.module}/templates/userdata.sh.tpl"
    template_extra_args = {}
    mime_type           = "text/x-shellscript"
  }
}
```

## Describe alternatives you've considered.
The only alternative to the above, whilst still using `node_groups`, appears to be to set `create_launch_template` to `false`, duplicate all of the behaviour outside of the module, then pass in the `launch_template_id`.

However, this is brittle (liable to falling behind the launch configuration in this module) and hard to achieve (as many of the variables set in this module's launch templates are derived from values derived from the inner workings of the module).  As such, I do not see it as a truly viable alternative.

[...]

PR #1645, opened 2021-10-14
feat: Create node group user data from given template

scalen:
# PR o'clock

Closes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1644

## Description

Prior to this commit, user data for node groups was generated from a
prescribed template, and user data in other formats (such as the TOML
used to configure Bottlerocket instances, see link) was not supported.

This change allows a node_group to specify its own user data template
file, the template's extra arguments and the user data's mime type;
this in turn supports alternative forms of user data as required by any
given AMI.

https://github.com/bottlerocket-os/bottlerocket#using-user-data

### Checklist

### DIFF
--- a/modules/node_groups/locals.tf
+++ b/modules/node_groups/locals.tf
@@ -15,6 +15,7 @@ locals {
       create_launch_template               = false
       bootstrap_env                        = {}
       kubelet_extra_args                   = var.workers_group_defaults["kubelet_extra_args"]
+      device_name                          = coalesce(var.workers_group_defaults["root_block_device_name"], "/dev/xvda")
       disk_size                            = var.workers_group_defaults["root_volume_size"]
       disk_type                            = var.workers_group_defaults["root_volume_type"]
       disk_iops                            = var.workers_group_defaults["root_iops"]
--- a/examples/managed_bottlerocket_node_group/main.tf
+++ b/examples/managed_bottlerocket_node_group/main.tf
@@ -57,6 +57,9 @@ module "eks" {
 [settings.kubernetes.node-labels]
 ingress = "allowed"
 EOT
+
+      device_name = "/dev/xvdb"
+      disk_type   = "gp3"
     }
   }
 
--- a/modules/node_groups/launch_template.tf
+++ b/modules/node_groups/launch_template.tf
@@ -28,7 +28,7 @@ resource "aws_launch_template" "workers" {
   update_default_version = lookup(each.value, "update_default_version", true)
 
   block_device_mappings {
-    device_name = "/dev/xvda"
+    device_name = each.value["device_name"]
 
     ebs {
       volume_size           = lookup(each.value, "disk_size", null)


stevehipwell:
@scalen I was planning on doing this work so I'm very happy to see you've already picked it up. Could you rebase this onto the latest master branch and I'll give it a review?

stevehipwell:
@scalen I'm not sure there is a precedent for complex objects in the node group config so I'd suggest removing the nesting and using `user_data_` as a prefix (e.g. `user_data_mime_type`), there are only 3 variables. This will also make it possible to only override one of the values without having to specifically set the others to their defaults, which wouldn't be as easy to use and could cause outdated defaults to be used if they were ever changed. Other than that this looks really good. How much testing have you done?

stevehipwell:
@scalen if you make the changes to the variables I'll review again and then we can see if @daroga0002 can approve running CI. Once this is merged I might take a look at adding the changes in my previous PRs and the changes here for launch template workers.

scalen:
@stevehipwell Requested changes made in 7135ba9338f52513510152c22d7e3b256f931951

stevehipwell (approved):
Looks good to me, @daroga0002 what do you think?

jaimehrubiks:
Does this modify existing node_groups with user data like these:

      pre_userdata = join("\n\n#----------------\n\n",[
        templatefile("${path.module}/templates/_volume_fg6001.sh", {}),
        templatefile("${path.module}/templates/_volume_fg6002.sh", {}),
      ])

If you don't change anything?

scalen:
> Does this modify existing node_groups with user data like these:
> 
> ```
>   pre_userdata = join("\n\n#----------------\n\n",[
>     templatefile("${path.module}/templates/_volume_fg6001.sh", {}),
>     templatefile("${path.module}/templates/_volume_fg6002.sh", {}),
>   ])
> ```
> 
> If you don't change anything?

No, the default behaviour remains the same.

bryantbiggs:
ah, good ol' userdata again. I think we just need to provide a bring-your-own user_data functionality. something like:

```hcl
	...
	user_data = var.create_user_data ?  base64encode(local.launch_template_userdata_rendered[count.index]) : var.user_data_base64
	...
```

thoughts @daroga0002 ?

stevehipwell:
@bryantbiggs isn't this PR providing a BYO fallback implementation?

bryantbiggs:
yes, but in a way that is getting very complex and convoluted

stevehipwell:
@bryantbiggs it's a complex problem that's being solved. The templated solutions currently provided by the input variables have a pretty good UX and for everyone else this PR will allow them to own the full complexity themselves.

scalen:
> yes, but in a way that is getting very complex and convoluted

It is a little convoluted, but by providing the template and additional args separately, it means we can utilise many variables created within the module. If we required the user data to be passed in whole, we may struggle to get some of the same synchronisation that is possible/trivial with the current approach.

modules/node_groups/locals.tf
```diff
@@ -34,6 +34,9 @@ locals {
       metadata_http_tokens                 = var.workers_group_defaults["metadata_http_tokens"]
       metadata_http_put_response_hop_limit = var.workers_group_defaults["metadata_http_put_response_hop_limit"]
       ami_is_eks_optimized                 = true
+      user_data_mime_type                  = "text/x-shellscript"
+      user_data_template_extra_args        = lookup(var.workers_group_defaults, "userdata_template_file", {})
```
bdwyertech, line 38: I think the lookup reference to `userdata_template_file` here was meant to be `user_data_template_extra_args`

scalen: You're right, this is incorrect, but [it should actually be `user data_template_extra_args`](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/locals.tf#L255-L259).

bdwyertech (approved):

bdwyertech (approved):

djordje-petrovic (approved):

stevehipwell:
@daroga0002 @antonbabenko could we get this reviewed and merged?

stevehipwell:
@scalen do you not need to provide a variable to customise the `block_device_mappings.device_name` to support Bottlerocket (and other OSs)?

See https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1695 & https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1685.

scalen:
> @scalen do you not need to provide a variable to customise the `block_device_mappings.device_name` to support Bottlerocket (and other OSs)?
> 
> See #1695 & #1685.

We haven't seen any issues running of this PR branch, but then we haven't specified any significant configuration.  I assume the suggestion is to make a more flexible version of [this change](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1695#issuecomment-975500387), to let the user define the volume to configure:

> `device_name = length(split("BOTTLEROCKET", each.value["ami_type"])) > 1 ? "/dev/xvdb" : "/dev/xvda"`

This still means only one volume will be configured, but it seems that will be fine for the Bottlerocket case 🤷🏼

stevehipwell:
@scalen that is correct; I assume that if you tried to configure the volume that it'd be applied to the wrong Bottlerocket volume (as is discussed in the other PR). A generic way to pass in the block device name would be a good solution, as would adding it in to the Bottlerocket example.

scalen:
> I made the original volume comment on #1695, saying that we shouldn't really care about the `xvda` volume. For context, `xvda` contains operating system itself: the active & passive partitions, the bootloader, the dm-verity data and etc. However, I can see the use case where you want to customize both `xvda` and `xvdb`. For example, in case you are required to encrypt all your ebs devices.

I thought of that too, but that is really starting to stray far from the point of the PR, which was to add support for alternative user data formats back to node_groups.  Making an extra value configurable in order to slightly extend the existing functionality to behave as one might expect for the specific Bottlerocket case is arguably related; I would also argue, however, that extending the module to support the configuration of multiple distinct volumes is a new feature, which should get its own PR.

stevehipwell:
@hi-artem @scalen the Bottlerocket OS volume is readonly so I think that if this PR supports changing the device name we're good here. Adding support for additional devices should be a new PR, but I think we need to make sure that the right Bottlerocket (or other OS) device is being configured in this PR.

scalen:
> I think we need to make sure that the right Bottlerocket (or other OS) device is being configured in this PR.

Done, @stevehipwell .

FrediWeber:
> @hi-artem @scalen the Bottlerocket OS volume is readonly so I think that if this PR supports changing the device name we're good here. Adding support for additional devices should be a new PR, but I think we need to make sure that the right Bottlerocket (or other OS) device is being configured in this PR.

Is the Bottlerocket volume RO from an EC2 standpoint? I thought the volume just gets mounted as RO.

stevehipwell:
@FrediWeber that's worth exploring. I guess the `block_device` variable could be made `block_devices` and take a list and then the mapping be dynamically looped which would support all potential cases? @scalen thoughts?

FrediWeber:
We just configured a launch template for a classic ASG and needed to configure both block devices to be encrypted and use a custom KMS key to be compliant with our regulations. I think it would be beneficial if we could encrypt all block devices on the Bottlerocket instances.

idan-gur:
What is the blockage on this PR?

stevehipwell:
@daroga0002 @antonbabenko could you approve the workflow even if you don't have time to review yet?

daroga0002:
> @daroga0002 @antonbabenko could you approve the workflow even if you don't have time to review yet?

I dont have permissions and @antonbabenko  is on re:Invent so probably he will be back next week.

I am a little stuck in last 2 weeks by some work, but I will try soon perform a review pending PRs.

scalen:
Tests from before rebase to fix formatting: https://github.com/terraform-aws-modules/terraform-aws-eks/runs/4397905746

lucasreed:
Checking in to see if this is in the pipeline for reviewing. Would love to see this released!

antonbabenko:

Issue #1578, opened 2021-09-09
Global Tags - Note: Objects have changed outside of Terraform every


Ashtonian:
## Description

Currently getting a change every run even though nothing is changing. I think this is an issue with setting the global tags on the provider. 

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions
 - Terraform:
`Terraform v1.0.6`
on darwin_amd64
 - Provider(s):
```
+ provider registry.terraform.io/hashicorp/aws v3.57.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/helm v2.3.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.4.1
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1
```

Set global tags on provider, setup eks, run tf apply.

no

yes

`tf apply`
### Code Snippet to Reproduce

```tf

# Configure the AWS Provider
provider "aws" {
  region = local.region

  // This is necessary so that tags required for eks can be applied to the vpc without changes to the vpc wiping them out.
  // https://stackoverflow.com/questions/57495581/terraform-eks-tagging
  // https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging
  ignore_tags {
    key_prefixes = ["kubernetes.io/", "k8s.io/"]
    keys         = concat(keys(local.global_tags), [])

  }
  default_tags {
    // https://github.com/hashicorp/terraform-provider-aws/issues/18311#issuecomment-860868546
    tags = local.global_tags
  }
}
locals{
  global_tags = {
    Environment  = local.environment
    TF_Timestamp = timestamp()
    TF_state     = "us-east-1:s3:stateblah"
    TF_lock      = "us-east-1:dynamo:keyblah"
    TF_source    = "https://github.com/blah"
  }
}


module "eks" {
[...]

Issue #1455, opened 2021-06-21
Module resources not inheriting `default_tags` from AWS provider.


citrix-ericgo:
## Description

Setting `default_tags` in a provider block does not percolate down to resources created by the module.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions
```
Terraform v0.15.4
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v3.45.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.3.2
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1
 - Module: terraform-aws-eks?
```

## Reproduction

Set a locals with a base_tags variable:
```
locals {
  base_tags = {
    environment = "foo"
    owner       = "bar@baz.qux"
  }
}
```
Apply default_tags to the aws provider:
```
terraform {
  required_version = "~> 0.14"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "3.45.0"
    }
  }
  backend "s3" {
    encrypt = true
  }
}
```
`terraform apply`

Observe no tags applied to the resources (such as AWS EC2 Autoscaling groups

### Code Snippet to Reproduce
```module "eks" {
  source                      = "terraform-aws-modules/eks/aws"
  cluster_name                = "widget-${var.environment}"
  cluster_version             = "1.18"
  subnets                     = var.private_subnets
  vpc_id                      = var.vpc_id
  cluster_enabled_log_types   = local.cluster_enabled_log_types
  manage_aws_auth             = true
  write_kubeconfig            = true
  kubeconfig_output_path      = "./"
  workers_additional_policies = ["arn:aws:iam::123456789012:policy/foobar"]
  map_users                   = var.user_permissions
  tags                        = local.base_tags

  worker_groups = [
    {
      name                 = "worker-group-1"
      instance_type        = var.instance_type
      asg_desired_capacity = 1
      asg_max_size         = 4
      asg_min_size         = 1
      disk_size            = 80
    },
    {
[...]

Issue #1248, opened 2021-02-17
cluster_AmazonEKSClusterPolicy always changing


bilalbayasut:
# I have issues

## I'm submitting a...

* [ ] bug report
* [ ] feature request
* [x] support request - read the [FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md) first!
* [x] kudos, thank you, warm fuzzy

## What is the current behavior?
I have node_group set and everytime I made changes, the policies are always changing
```
# module.k8s.module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSClusterPolicy[0] must be replaced
+/- resource "aws_iam_role_policy_attachment" "cluster_AmazonEKSClusterPolicy" {
      ~ id         = "cluster-name20210216185206327200000001-20210217180909569600000006" -> (known after apply)
      ~ policy_arn = "arn:aws:iam::aws:policy/AmazonEKSClusterPolicy" -> (known after apply) # forces replacement
        # (1 unchanged attribute hidden)
    }

  # module.k8s.module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSServicePolicy[0] must be replaced
+/- resource "aws_iam_role_policy_attachment" "cluster_AmazonEKSServicePolicy" {
      ~ id         = "cluster-name20210216185206327200000001-20210217180909465800000001" -> (known after apply)
      ~ policy_arn = "arn:aws:iam::aws:policy/AmazonEKSServicePolicy" -> (known after apply) # forces replacement
        # (1 unchanged attribute hidden)
    }

  # module.k8s.module.eks.aws_iam_role_policy_attachment.cluster_AmazonEKSVPCResourceControllerPolicy[0] must be replaced
+/- resource "aws_iam_role_policy_attachment" "cluster_AmazonEKSVPCResourceControllerPolicy" {
      ~ id         = "cluster-name20210216185206327200000001-20210217180909476200000002" -> (known after apply)
      ~ policy_arn = "arn:aws:iam::aws:policy/AmazonEKSVPCResourceController" -> (known after apply) # forces replacement
        # (1 unchanged attribute hidden)
    }

  # module.k8s.module.eks.aws_iam_role_policy_attachment.workers_AmazonEC2ContainerRegistryReadOnly[0] must be replaced
+/- resource "aws_iam_role_policy_attachment" "workers_AmazonEC2ContainerRegistryReadOnly" {
[...]

PR #1654, opened 2021-10-20
improvement: Introduce wg_tags to handle default_tags on provider level

kbcz1989:
# PR o'clock
closes: [#1578](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1578), [#1455(?)](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1455), [#1248(?)
](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1248)

## Description

When default_tags are used on aws provider level, and the same tags are defined as input to this module, we end up with this issue:
[default_tags always shows an update](https://github.com/hashicorp/terraform-provider-aws/issues/18311)

If "tags" var is not set, no tags will be applied to resources which defines tag_specification.

So, either we have default_tags and do not set "tags". which leaves us with "empty" tag_specifications or we don't use default_tags.

This PR introduces new variable "wg_tags", which will be used on resources which define dynamic tags/tag_specifications if "tags" is not set. Backward compatibility is therefore fully retained.

There will be changes only if wg_tags is set, otherwise the module works the same as before.

### Checklist


hatemosphere:
@antonbabenko this is actually quite important with the current behaviour of the AWS provider's default tags. we had to disable the usage of default tags in TF state with this module due to not having the ability to set non-default tags only on  tag_specifications and dynamic tag resources

antonbabenko:
If anyone wants to work on this one, please check this PR (around data source `aws_default_tags`) - https://github.com/terraform-aws-modules/terraform-aws-autoscaling/pull/169/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR1

kbcz1989:
The solution proposed by @antonbabenko is much better. I will close this PR and try to implement the proposed solution when I find some time.

PR #1675, opened 2021-11-05
feat: Add autoscaling_group_tag to node_groups

wgj:
Initial commit for ci tests

# PR o'clock

## Description

Please explain the changes you made here and link to any relevant issues.

### Checklist

- [ ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

Issue #1558, opened 2021-08-31
Add Tags to ASG when using node groups


marianobilli:
## Is your request related to a new offering from AWS? 
No

## Is your request related to a problem? Please describe.
When using nodegroups it is necessary to be able to add tags to the autoscaling groups as described in [AWS EKS Cluster Autoscaler Doc](https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html)

in my case I need to add tags related to taints
`Key: k8s.io/cluster-autoscaler/node-template/taint/$TAINT_KEY
Value: NoSchedule`

## Describe the solution you'd like.
Include in the node group module, the positiblity to add `asg_tags` per each nodegroup and in the module use the terraform resource `aws_autoscaling_group_tag` 

## Describe alternatives you've considered.
Manually mantain the ASG tags.  t(-_-t)

marianobilli:
> you have map `node_groups_defaults` where you can specify key `additional_tags` or specify this directly in node group map as `additional_tags` and then those tags can be added.
> 
> https://github.com/terraform-aws-modules/terraform-aws-eks/blob/979d62d9b807db10ed3a6327c3554ee480175c62/modules/node_groups/node_groups.tf#L79-L83

But those tags are going to the node group not to the autoscaling group related to the node group

daroga0002:
autoscaling group for node groups is done thru EKS api, so you cannot influence it. If you want have own autoscaling group then follow worker nodes approach with custom AMI

marianobilli:
> or node groups is done thru EKS api, so you cannot influence it. If you want have own autoscaling group then follow worker nodes approach with custom AMI

I like the nodegroups managed by aws.
Currently I added this code outside the EKS module, but adding it inside the node_group module would be so much easier

```
resource "aws_autoscaling_group_tag" "tag" {
  for_each = local.asg_tags
  autoscaling_group_name = module.eks_cluster.node_groups[split("_",each.key)[0]].resources[0].autoscaling_groups[0].name

  tag {
    key   = each.value.key
    value = each.value.value
    propagate_at_launch = each.value.propagate_at_launch
  }
}
```

I can do a proposed PR but only if the maintainers agree to the idea

daroga0002:
this approach is not supported by AWS, so probably we shouldn't implement it (even if possible). 

EKS node group has own limitations which people must accept choosing it, if they want to hack or use unsupported tricks then I think they make it on own responsibility.

[...]

PR #1680, opened 2021-11-07
BREAKING CHANGE: remove support for launch configuration and replace `count` with `for_each`

bryantbiggs:
# PR o'clock

## Description

### Checklist


daroga0002:
@bryantbiggs I think this feature should be moved to wubmodule
similar to node groups

bryantbiggs:
@daroga0002 / @antonbabenko this is getting close for review if you have time to check it out. I still need to populate the upgrade document and comprehensive list of changes, plus a few other design decisions (i.e. - using the default blocks, security group usage, etc.). I've made some more changes today to align variable usage and descriptions so I need to run through all of the examples yet again tomorrow to re-validate but 🤞🏽 they are working as intended (if not, please let me know). 

Let me know if you see any glaring/obvious errors - I'll be working on the upgrade doc and I would like to avoid having to make multiple iterations on it since I know its going to be quite involved.

cc - @barryib / @max-rocket-internet / @dpiddock / @lisfo4ka for any input you may have as well 🙏🏽

examples/irsa_autoscale_refresh/main.tf
```diff
@@ -0,0 +1,165 @@
...
+      name = "terraform"
+      context = {
+        cluster = "${module.eks.cluster_id}"
+        user    = "terraform"
+      }
+    }]
+    users = [{
+      name = "terraform"
+      user = {
+        token = "${data.aws_eks_cluster_auth.this.token}"
```
antonbabenko, line 105:
```suggestion
        token = data.aws_eks_cluster_auth.this.token
```

Similar fix in other places above.

modules/_user_data/variables.tf
```diff
@@ -0,0 +1,65 @@
+variable "create" {
+  description = "Determines whether to create EKS managed node group or not"
```
antonbabenko, line 2:
```suggestion
  description = "Determines whether to create user-data or not"
```

antonbabenko (approved):
I did the first round of reviews and it looks very good. Minor comments here and there.

antonbabenko:
I think it is a good idea to have an SVG diagram describing bootstrap logic - https://raw.githubusercontent.com/bryantbiggs/terraform-aws-eks/refactor/breaking-changes/.github/images/user_data.svg

examples/irsa_autoscale_refresh/main.tf
bryantbiggs: ah yes, fixed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1680/commits/e9d94763b53ccb21f9bc965cc8cdafd42111e8f3

bryantbiggs:
> I think it is a good idea to have an SVG diagram describing bootstrap logic - [bryantbiggs/terraform-aws-eks@`refactor`/breaking-changes/.github/images/user_data.svg (raw)](https://raw.githubusercontent.com/bryantbiggs/terraform-aws-eks/refactor/breaking-changes/.github/images/user_data.svg)

yes, agreed! the images will be in the final PR but right now I have both a relative link and a fully qualified link. the relative link allows the image to be rendered in this PR just to make sure everything shows up as expected but it won't render on the remote Terraform docs for the module; this is what the fully qualified URL is for. so I was just putting a TODO reminder to remove the relative links before merging so that we have the fully qualified URLs that will show the images on both GitHub and on Terraform's module docs

examples/irsa_autoscale_refresh/charts.tf
```diff
@@ -0,0 +1,292 @@
...
+    value = "true"
+  }
+
+  set {
+    name  = "rbac.create"
+    value = "true"
+  }
+
+  depends_on = [
+    module.eks
```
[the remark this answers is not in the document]
bryantbiggs: yes; updated - thank you!

bryantbiggs:
ok this is finally ready for final review. I tried to work out some `terraform state mv ...` steps but in the end the module wants to replace nearly everything that is moved to the new name due to the extensive use of `name_prefix` (IAM roles, security groups, launch templates, autoscaling group, EKS node group, etc.). Therefore, I have left out any state mv recommendations from the upgrade guide. Let me know if there are any additional changes/feedback/final validation - @antonbabenko / @terraform-aws-modules/triage-supporters

daroga0002:
@bryantbiggs  https://www.terraform.io/docs/language/modules/develop/refactoring.html this can be helpful for migration

Terraform 1.1 introducing this 🎉

antonbabenko:
Are we raising minimum requirements from Terraform 0.13 to 1.1 already? :) It will be a surprise for many users, plus refactoring features released yesterday are not battle-tested yet. I think we can come back to this and start using something like this in half a year but not before.

examples/user_data/templates/windows_custom.tpl
```diff
@@ -0,0 +1,10 @@
+# Custom user data template provided for rendering
+<powershell>
+${pre_bootstrap_user_data ~}
+[string]$EKSBinDir = "$env:ProgramFiles\Amazon\EKS"
+[string]$EKSBootstrapScriptName = 'Start-EKSBootstrap.ps1'
+[string]$EKSBootstrapScriptFile = "$EKSBinDir\$EKSBootstrapScriptName"
+& $EKSBootstrapScriptFile -EKSClusterName ${cluster_name} ${bootstrap_extra_args} 3>&1 4>&1 5>&1 6>&1
```
daroga0002, line 7: can we unify this with linux parameters to also use B64_CLUSTER_CA and API_SERVER_URL as then we wil be able to bootstrap node without asking AWS API (bootstrap scripts in case when parameters are not passed are doing awscli to gather required info)

Example:
```
& $EKSBootstrapScriptFile -EKSClusterName "windohs" -APIServerEndpoint "https://test.com" -Base64ClusterCA "dGVzdA==" -KubeletExtraArgs "--node-labels= --register-with-taints=" 3>&1 4>&1 5>&1 6>&1
```

examples/user_data/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 0.13.1"
```
daroga0002, line 2: do we really need to support so legacy version? a lot of issues or plan diffs are happening on legacy terraform versions (as in some point of time the split refresh phase from terraform)?

examples/irsa_autoscale_refresh/main.tf
```diff
@@ -0,0 +1,211 @@
...
+      }
+    }]
+  })
+}
+
+resource "null_resource" "apply" {
+  triggers = {
+    kubeconfig = base64encode(local.kubeconfig)
+    cmd_patch  = <<-EOT
+      kubectl create configmap aws-auth -n kube-system --kubeconfig <(echo $KUBECONFIG | base64 --decode)
```
daroga0002, line 163: this can be very problematic in case of terraform cloud probably

daroga0002 (commented):
few questions:
- do we remove users/roles feature in aws-auth config? what will be desired way to make it?
- aws-auth seems be now required to be done outside module what is causing that in case of self nodes it will always require handling this externally (so module can be threated as not `sufficient` in  terms of creating cluster
- EKS silently create in background SG which is assigned by default to node groups, can we also assign this by default to self nodes so then we will have unified approach and will dont require adding those SGs for DNS and etc.
- what will happen with legacy module? should it be archived in seperate branch?


Beside some questions and concerns looks pretty well and solving a lot of issues 🎉

.pre-commit-config.yaml
```diff
@@ -26,3 +26,4 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-merge-conflict
```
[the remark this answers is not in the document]
bryantbiggs: I believe the pre-commit setup we have already includes this https://github.com/antonbabenko/pre-commit-terraform/blob/7c6ad7cfb40d92e8f49cb0a3dd412c7509d5aeda/.pre-commit-config.yaml#L14 - I'm not 100% sure how that all works but I believe we do have that check in the hooks as well

examples/user_data/templates/windows_custom.tpl
bryantbiggs: fixed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1680/commits/d380c9389f4ed61da1a51fa65099b9dd31a53438

examples/irsa_autoscale_refresh/main.tf
bryantbiggs: agreed! this is just a quick and dirty way to facilitate testing and demonstrate steps users may need to take/handle on their end. what this change is proposing here is that managing/updating/creating the aws-auth configmap is now up to users and outside the scope of this module. you can do hacky kubectl way like shown (least amount of effort, again just for testing and verification), you can integrate it with your gitops approach if so desired, etc. There are numerous ways, but it greatly simplifies the module by not prescribing and managing the aws-auth configmap going forward (IMO)

bryantbiggs:
thank you for the feedback @daroga0002 ; to answer some of your questions posed: 

> - do we remove users/roles feature in aws-auth config? what will be desired way to make it?

this change set has removed all management of the aws-auth configmap. there is one output that is provided that renders a configmap containing all of the roles created by the module, but how users utilize/manage the configmap is now completely up to them. https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1680#discussion_r765859051

> - aws-auth seems be now required to be done outside module what is causing that in case of self nodes it will always require handling this externally (so module can be threated as not sufficient in terms of creating cluster

I think this is similar to above (apologies, not quite following the question). yes; aws-auth is out of scope for the module in this proposed change

> - EKS silently create in background SG which is assigned by default to node groups, can we also assign this by default to self nodes so then we will have unified approach and will dont require adding those SGs for DNS and etc.

yes, I believe you are correct that EKS managed node groups do create a security group by default that connects the node group to the control plane. could we somehow integrate and tap into that node group in this module - TBH I'm not sure. just off the top of my head I see a few concerns:
- It looks very precarious and ripe for cyclical dependencies which are problematic
- It is only made available when an EKS managed node group is added into the mix (maybe Fargate profiles, I can't remember?), but if you only use self-managed node groups then this is not present and you are back to no network connectivity

I'm not saying its not a good idea, I just don't know what that would look like other than adding in quite a bit of complexity and weird edge cases that will once again introduce more complexity back into the module

> - what will happen with legacy module? should it be archived in seperate branch?

Nothing, just git business as usual. The 17.x (and all prior) are eternally locked into git/github/and the Terraform registry. we always strongly recommend users fully pin their module versions so they don't get any unexpected surprises so even if this does go forward, people using 17.x today will not know anything different. However, if they wish to utilize new features/fixes/changes, then they will have to evaluate the upgrade to 18.x and make that change when they see fit

examples/user_data/versions.tf
antonbabenko: @daroga0002 IIRC 0.15.4 and 1.0 have breaking behavior related to the way "outside changes" are tracked. This is one more reason to leave versions ranging so wide (some users can't simply upgrade so fast).

README.md
```diff
@@ -1,310 +1,779 @@
...
+- AWS EKS Cluster with one or more AWS EKS Managed Node Groups
+- AWS EKS Cluster with one or more Self Managed Node Groups
+- AWS EKS Cluster with one or more Fargate profiles
+- AWS EKS Cluster with one or more AWS EKS Managed Node Groups, one or more Self Managed Node Groups, one or more Fargate profiles
 
-- [fargate](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/fargate) - creates Fargate profiles, see [examples/fargate](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/fargate) for detailed examples.
-<!--
-- [node_groups](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/node_groups) - creates Managed Node Group resources
--->
+It is also possible to have the various node groups of each family configured differently, as well as externally defined (outside of the root `eks` module definition) node groups using the provided sub-modules attached to the cluster created - there are no restrictions on the the various different possibilities provided by the module.
```
johngmyers, line 345:
```suggestion
It is also possible to configure the various node groups of each family differently. Node groups may also be defined outside of the root `eks` module definition by using the provided sub-modules. There are no restrictions on the the various different possibilities provided by the module.
```

README.md
```diff
@@ -1,310 +1,779 @@
...
+      }
+    }
+  }
+```
+
+See the [`examples/complete/` example](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/complete) for a working example of these configurations.
+
+### Default configurations
+
+Each node group type (EKS managed node group, self managed node group, Fargate profile) provides a default configuration setting that allows users to provide their own default configurations instead of the modules default configurations. This allows users to set a common set of defaults for their node groups and still maintain the ability to override these settings within the specific node group definition. The order of precedence for each node group type roughly follows (from highest precedence, to least):
```
johngmyers, line 471:
```suggestion
Each node group type (EKS managed node group, self managed node group, or Fargate profile) provides a default configuration setting that allows users to provide their own default configuration instead of the module's default configuration. This allows users to set a common set of defaults for their node groups and still maintain the ability to override these settings within the specific node group definition. The order of precedence for each node group type roughly follows (from highest to least precedence):
```

README.md
```diff
@@ -1,310 +1,779 @@
...
+
+### General Notes
+
+While the module is designed to be flexible and support as many use cases and configurations as possible, there is a limit to what first class support can be provided without over-burdening the complexity of the module. Below are a list of general notes on the design intent captured by this module which hopefully explains some of the decisions that are, or will be made, in terms of what is added/supported natively by the module:
+
+- Despite the addition of Windows Subsystem for Linux (WSL for short), containerization technology is very much a suite of Linux constructs and therefore Linux is the primary OS supported by this module. In addition, due to the first class support provided by AWS, Bottlerocket OS and Fargate Profiles are also very much fully supported by this module. This module does not make any attempt to NOT support Windows, as in preventing the usage of Windows based nodes, however it is up to users to put in additional effort in order to operate Windows based nodes when using the module. User can refer to the [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html) for further details. What does this mean:
```
johngmyers, line 520:
```suggestion
- Despite the addition of Windows Subsystem for Linux (WSL for short), containerization technology is very much a suite of Linux constructs and therefore Linux is the primary OS supported by this module. In addition, due to the first class support provided by AWS, Bottlerocket OS and Fargate Profiles are also very much fully supported by this module. This module does not make any attempt to NOT support Windows, as in preventing the usage of Windows based nodes, however it is up to users to put in additional effort in order to operate Windows based nodes when using the module. User can refer to the [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html) for further details. What this means is:
```

README.md
```diff
@@ -1,310 +1,779 @@
...
+- AWS EKS Managed Node groups are current the preferred route over Self Managed Node Groups for compute nodes. Both operate very similarly - both are backed by autoscaling groups and launch templates deployed and visible within your account. However, AWS EKS Managed Node groups provide a better user experience and offer a more "managed service" experience and therefore has precedence over Self Managed Node Groups. That said, there are currently inherent limitations as AWS continues to rollout additional feature support similar to the level of customization you can achieve with Self Managed Node Groups. When requesting added feature support for AWS EKS Managed Node groups, please ensure you have verified that the feature(s) are 1) supported by AWS and 2) supported by the Terraform AWS provider before submitting a feature request.
```
johngmyers, line 523:
```suggestion
- AWS EKS Managed Node groups are currently the preferred route over Self Managed Node Groups for compute nodes. Both operate very similarly - both are backed by autoscaling groups and launch templates deployed and visible within your account. However, AWS EKS Managed Node groups provide a better user experience and offer a more "managed service" experience and therefore has precedence over Self Managed Node Groups. That said, there are currently inherent limitations as AWS continues to rollout additional feature support similar to the level of customization you can achieve with Self Managed Node Groups. When requesting added feature support for AWS EKS Managed Node groups, please ensure you have verified that the feature(s) are 1) supported by AWS and 2) supported by the Terraform AWS provider before submitting a feature request.
```

README.md
```diff
@@ -1,310 +1,779 @@
...
+<h4>How can I work with the cluster if I disable the public endpoint?</h4>
 
-Report issues/questions/feature requests on in the [issues](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/new) section.
+You have to interact with the cluster from within the VPC that it is associated with; either through a VPN connection, a bastion EC2 instance, etc.
 
-Full contributing [guidelines are covered here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/CONTRIBUTING.md).
+<h4>How can I stop Terraform from removing the EKS tags from my VPC and subnets?</h4>
 
-## Authors
+You need to add the tags to the VPC and subnets yourself. See the [basic example](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/basic).
```
johngmyers, line 588:
```suggestion
You need to add the tags to the Terraform definition of the VPC and subnets yourself. See the [basic example](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/examples/basic).
```

README.md
```diff
@@ -1,310 +1,779 @@
...
-## License
+<h4>Why are there no changes when a node group's desired count is modified?</h4>
+
+The module is configured to ignore this value. Unfortunately, Terraform does not support variables within the `lifecycle` block. The setting is ignored to allow the cluster autoscaler to work correctly so that `terraform apply` does not accidentally remove running workers. You can change the desired count via the CLI or console if you're not using the cluster autoscaler.
+
+If you are not using autoscaling and want to control the number of nodes via terraform, set the `min_size` and `max_size` for node groups. Before changing those values, you must satisfy AWS `desired_size` constraints (which must be between new min/max values).
+
+<h4>Why are nodes not recreated when the `launch_template` is recreated?</h4>
+
+By default the ASG is not configured to be recreated when the launch configuration or template changes; you will need to use a process to drain and cycle the nodes.
```
johngmyers, line 600: Is this not only relevant for self-managed groups?

johngmyers (changes requested):

README.md
```diff
@@ -1,310 +1,779 @@
...
 
-## Examples
+Cluster private endpoint can also be enabled by setting `cluster_endpoint_private_access = true` on this module. Node communication to the endpoint stays within the VPC. Ensure that VPC DNS resolution and hostnames are also enabled for your VPC when the private endpoint is enabled.
 
-There are detailed examples available for you to see how certain features of this module can be used in a straightforward way. Make sure to check them and run them before opening an issue. [Here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/iam-permissions.md) you can find the list of the minimum IAM Permissions required to create EKS cluster.
+Nodes need to be able to connect to other AWS services plus pull down container images from container registries (ECR). If for some reason you cannot enable public internet access for nodes you can add VPC endpoints to the relevant services: EC2 API, ECR API, ECR DKR and S3.
```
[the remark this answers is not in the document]
bryantbiggs: Thank you for that feedback! While I do agree with the sentiment shared, I am cautiously trying to avoid re-creating the current situation where various bits of documentation are added for "How to troubleshoot EKS". I'll have to think on how best to incorporate the suggestion without going down that rabbit hole. Ideally, the docs here provided something short and to the point of "Here is what you need, and go look at x,y,z for help troubleshooting any issues that you may encounter"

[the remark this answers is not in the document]
johngmyers: I think the minimum is to mention that there is a second potential cause.

README.md
bryantbiggs: I don't believe this is related to AWS EKS managed node groups because there are provisions to [force updates](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#force_update_version) as well as update [configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#update_config-configuration-block)

However, I see a reference to launch configuration here that I need to remove

README.md
```diff
@@ -1,310 +1,779 @@
...
-## License
+<h4>Why are there no changes when a node group's desired count is modified?</h4>
+
+The module is configured to ignore this value. Unfortunately, Terraform does not support variables within the `lifecycle` block. The setting is ignored to allow the cluster autoscaler to work correctly so that `terraform apply` does not accidentally remove running workers. You can change the desired count via the CLI or console if you're not using the cluster autoscaler.
+
+If you are not using autoscaling and want to control the number of nodes via terraform, set the `min_size` and `max_size` for node groups. Before changing those values, you must satisfy AWS `desired_size` constraints (which must be between new min/max values).
+
+<h4>Why are nodes not recreated when the `launch_template` is recreated?</h4>
+
+By default the ASG is not configured to be recreated when the launch configuration or template changes; you will need to use a process to drain and cycle the nodes.
```
johngmyers, line 600:
```suggestion
By default the ASG for a self-managed node group is not configured to be recreated when the launch configuration or template changes; you will need to use a process to drain and cycle the nodes.
```

README.md
```diff
@@ -1,310 +1,779 @@
...
+If you are using the cluster autoscaler:
+
+- Drain an old node `kubectl drain --force --ignore-daemonsets --delete-local-data ip-xxxxxxx.eu-west-1.compute.internal`
+- Wait for pods to be Running
+- Cluster autoscaler will create new nodes when required
+- Repeat until all old nodes are drained
+- Cluster autoscaler will terminate the old nodes after 10-60 minutes automatically
+
+You can also use a third-party tool like Gruntwork's kubergrunt. See the [`eks deploy`](https://github.com/gruntwork-io/kubergrunt#deploy) subcommand.
+
```
johngmyers, line 619:
```suggestion

Alternatively, use a managed node group instead.
```

johngmyers (changes requested):

modules/eks-managed-node-group/variables.tf
```diff
@@ -0,0 +1,461 @@
...
+}
+
+################################################################################
+# Launch template
+################################################################################
+
+variable "create_launch_template" {
+  description = "Determines whether to create a launch template or not. By default, EKS will use its own default launch template"
+  type        = bool
+  default     = false
```
johngmyers, line 78:
```suggestion
  description = "Determines whether to create a launch template or not. If set to `false`, EKS will use its own default launch template"
  type        = bool
  default     = true
```

There are no downsides to creating a launch template when using this module and many features require enabling launch templates. Enabling this by default will simplify use of this module greatly.

bryantbiggs: Agreed - can't believe I forgot to flip this. It actually solves the whole conundrum of tags on ASGs without any hacky efforts 
- #1558
- #1675
- #1688
- #1689
- #1705
- #1711

README.md
```diff
@@ -1,310 +1,780 @@
...
-## Usage example
+1. By default, the `eks-managed-node-group` sub-module will use the default configurations provided by AWS EKS Managed Node Groups; EKS MNG will provide its own launch template and utilize the latest AWS EKS Optimized AMI (Linux) for the given Kubernetes version:
 
 ```hcl
-data "aws_eks_cluster" "eks" {
-  name = module.eks.cluster_id
-}
+  eks_managed_node_groups = {
+    default = {}
+  }
```
johngmyers, line 184:
```suggestion
  eks_managed_node_groups = {
    default_launch_temlate = {
       create_launch_template = false        
    }
  }
```

README.md
```diff
@@ -1,310 +1,780 @@
...
 
-module "eks" {
-  source          = "terraform-aws-modules/eks/aws"
+3. AWS EKS Managed Node Groups allow you to extend configurations by providing your own launch template and user data that is merged with what the service provides. For example, to provide additional user data before the nodes are bootstrapped as well as supply additional arguments to the bootstrap script:
+
+```hcl
+  eks_managed_node_groups = {
+    extend_config = {
+      create_launch_template = true
+
```
johngmyers, line 204:
```suggestion
```

README.md
```diff
@@ -1,310 +1,780 @@
...
+  }
+```
+
+4. The same configurations extension is offered when utilizing Bottlerocket OS AMIs, but the user data is slightly different. Bottlerocket OS uses a TOML user data file and you can provide additional configuration settings via the `bootstrap_extra_args` variable which gets merged into what is provided by the AWS EKS Managed Node Service:
+
+```hcl
+  eks_managed_node_groups = {
+    bottlerocket_extend_config = {
+      create_launch_template = true
+
```
johngmyers, line 225:
```suggestion
 ```

File: README.md [hunk 21]
```diff
@@ -1,310 +1,780 @@
...
+  }
+```
 
+5. Users can also utilize a custom AMI, but doing so means that AWS EKS Managed Node Group will NOT inject the necessary bootstrap script and configurations into the user data supplied to the launch template. When using a custom AMI, users must also opt in to bootstrapping the nodes  via user data and either use the module default user data template or provide your own user data template file:
+
+```hcl
+  eks_managed_node_groups = {
+    custom_ami = {
+      create_launch_template = true
+
```
johngmyers, line 247:
```suggestion
```

README.md
```diff
@@ -1,310 +1,780 @@
...
+6. Similarly, for Bottlerocket there is similar support:
+
+```hcl
+  eks_managed_node_groups = {
+    bottlerocket_custom_ami = {
+      ami_id   = "ami-0ff61e0bcfc81dc94"
+      platform = "bottlerocket"
+
+      create_launch_template = true
+
```
johngmyers, line 280:
```suggestion
 ```

File: README.md [hunk 23]
```diff
@@ -1,310 +1,780 @@
...
+    }
+  }
+
+  # EKS Managed Node Group(s)
+  eks_managed_node_group_defaults = {
+    ami_type               = "AL2_x86_64"
+    disk_size              = 50
+    instance_types         = ["m6i.large", "m5.large", "m5n.large", "m5zn.large"]
+    vpc_security_group_ids = [aws_security_group.additional.id]
+    create_launch_template = true
```
johngmyers, line 403:
```suggestion
```

johngmyers (changes requested):
... and so on.

bryantbiggs:
ok - the last change is in, this should be ready to go at any time @antonbabenko

Vlaaaaaaad:
@bryantbiggs awesome work on writing all this! 👏 ❤️ 

I got excited seeing this, so I tried using this new version of the module! Unfortunately, I hit an issue. Chances are I am doing something wrong, but when I try to pass in dynamic resources to this new version of the module, I get an error: `The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the for_each depends on.` I opened a PR with a minimal example of failing code here: https://github.com/bryantbiggs/terraform-aws-eks/pull/1

I think this may be due to [Terraform's limitations for `for_each`](https://www.terraform.io/docs/language/meta-arguments/for_each.html#limitations-on-values-used-in-for_each)? Or I am doing something wrong. Both are possible 😅

examples/fargate_profile/main.tf
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  fargate_profiles = {
+    default = {
+      name = "default"
+      selectors = [
+        {
+          namespace = "kube-system"
+          labels = {
+            k8s-app = "kube-dns"
```
Vlaaaaaaad, line 75: I'd remove this example, or use a fake name here.

CoreDNS can't run on Fargate without changes( https://github.com/aws/containers-roadmap/issues/944) and leads to this error `1 node(s) had taint {eks.amazonaws.com/compute-type: fargate}, that the pod didn't tolerate.`

bryantbiggs: yes, this is an unfortunate issue with Fargate https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-coredns

thank you for the suggestion, I think changing the name will work best to avoid any confusion

antonbabenko:
@bryantbiggs Sorry for the delay. Is it ready for final review & merge (I remember there were some Q in your fork)?

DanyC97:
> Drop support for managing aws-auth configmap

@bryantbiggs what will be replaced with ?

bryantbiggs:
> > Drop support for managing aws-auth configmap
> 
> @bryantbiggs what will be replaced with ?

@DanyC97 it is left up to users as to how they want to manage the aws-auth configmap. This change set does provided a rendered aws-auth configmap template that can be used (see examples)

DanyC97:
> > > Drop support for managing aws-auth configmap
> > 
> > 
> > @bryantbiggs what will be replaced with ?
> 
> @DanyC97 it is left up to users as to how they want to manage the aws-auth configmap. This change set does provided a rendered aws-auth configmap template that can be used (see examples)

hmm, i see. If before we had "batteries included" now the user will need to know that "hey, you need one more thing to do outside/ in addition to current module".  Hopefully this note will be somehow written in bold ;)

bryantbiggs:
> > > > Drop support for managing aws-auth configmap
> > > 
> > > 
> > > @bryantbiggs what will be replaced with ?
> > 
> > 
> > @DanyC97 it is left up to users as to how they want to manage the aws-auth configmap. This change set does provided a rendered aws-auth configmap template that can be used (see examples)
> 
> hmm, i see. If before we had "batteries included" now the user will need to know that "hey, you need one more thing to do outside/ in addition to current module". Hopefully this note will be somehow written in bold ;)

This is all captured in the extensive [UPGRADE-18.0.md](https://github.com/bryantbiggs/terraform-aws-eks/blob/refactor/breaking-changes/UPGRADE-18.0.md) guide

I will also caution, your statement above rings true for the current "batteries included" configuration. Users are required to have aws-iam-authenticator installed locally to where they are running Terraform; this is one of the many reasons for removing. This module is for provisioning AWS infrastructure resources for EKS, while cluster configuration and management is left out of scope in order to reduce the burden on this module (meaning whatever changes sent to the cluster API, those should not be managed here as that is mixing concerns)

DanyC97:
> > > > > Drop support for managing aws-auth configmap
> > > > 
> > > > 
> > > > @bryantbiggs what will be replaced with ?
> > > 
> > > 
> > > @DanyC97 it is left up to users as to how they want to manage the aws-auth configmap. This change set does provided a rendered aws-auth configmap template that can be used (see examples)
> > 
> > 
> > hmm, i see. If before we had "batteries included" now the user will need to know that "hey, you need one more thing to do outside/ in addition to current module". Hopefully this note will be somehow written in bold ;)
> 
> This is all captured in the extensive [UPGRADE-18.0.md](https://github.com/bryantbiggs/terraform-aws-eks/blob/refactor/breaking-changes/UPGRADE-18.0.md) guide
> 
> I will also caution, your statement above rings true for the current "batteries included" configuration. Users are required to have aws-iam-authenticator installed locally to where they are running Terraform; this is one of the many reasons for removing. This module is for provisioning AWS infrastructure resources for EKS, while cluster configuration and management is left out of scope in order to reduce the burden on this module (meaning whatever changes sent to the cluster API, those should not be managed here as that is mixing concerns)

fair @bryantbiggs  thanks for taking detailed info

marandalucas:
@bryantbiggs Great Job! When the version 18.0 will be released? Thanks so much

martijnvdp:
> @bryantbiggs awesome work on writing all this! 👏 ❤️
> 
> I got excited seeing this, so I tried using this new version of the module! Unfortunately, I hit an issue. Chances are I am doing something wrong, but when I try to pass in dynamic resources to this new version of the module, I get an error: `The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the for_each depends on.` I opened a PR with a minimal example of failing code here: [bryantbiggs#1](https://github.com/bryantbiggs/terraform-aws-eks/pull/1)
> 
> I think this may be due to [Terraform's limitations for `for_each`](https://www.terraform.io/docs/language/meta-arguments/for_each.html#limitations-on-values-used-in-for_each)? Or I am doing something wrong. Both are possible 😅

workaround is to precreate the iam roles

examples/user_data/main.tf
```diff
@@ -0,0 +1,289 @@
...
+  user_data_template_path = "${path.module}/templates/windows_custom.tpl"
+
+  pre_bootstrap_user_data = <<-EOT
+  [string]$Something = 'IDoNotKnowAnyPowerShell ¯\_(ツ)_/¯'
+  EOT
+  # I don't know if this is the right way on WindowsOS, but its just a string check here anyways
+  bootstrap_extra_args = "-KubeletExtraArgs --node-labels=node.kubernetes.io/lifecycle=spot"
+
+  post_bootstrap_user_data = <<-EOT
+  [string]$Something = 'IStillDoNotKnowAnyPowerShell ¯\_(ツ)_/¯'
```
antonbabenko, line 287: :trollface: :trollface: :trollface:

versions.tf
```diff
@@ -4,23 +4,11 @@ terraform {
...
-    kubernetes = {
-      source  = "hashicorp/kubernetes"
-      version = ">= 1.11.1"
-    }
-    cloudinit = {
-      source  = "hashicorp/cloudinit"
-      version = ">= 2.0"
-    }
-    http = {
-      source  = "terraform-aws-modules/http"
```
antonbabenko, line 22: Do you know if there is still a need for this `http` provider we forked some time ago?

antonbabenko (approved):

antonbabenko:
@bryantbiggs I am not sure why this PR didn't trigger a major release action, do you know?

https://github.com/terraform-aws-modules/terraform-aws-eks/runs/4714285326?check_suite_focus=true

antonbabenko:
@bryantbiggs I renamed the PR title to include `!` to make sure that this should be a breaking release but it didn't seem to help. Weird. https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change

mcanevet:
@antonbabenko it's probably because none of the commit (not the PR title) contains a `!`.

antonbabenko:
PR title was "feat!: ...", so it was used as a commit ee9f0c646a45ca9baa6174a036d1e09bcccb87b1, too.

mcanevet:
oh yes, you're right

bryantbiggs:
well thats really annoying! let me see what I can find out @antonbabenko

bryantbiggs:
I've opened a PR to fix this here https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1736 based on some testing I did today to dive into this issue. tl;dr - default of `angular` preset does not respect the `!` and requires `BREAKING CHANGE` in the commit footer to trigger a breaking change (`feat!` is not valid either under this preset which is why nothing was released)

versions.tf
bryantbiggs: No I don't believe we need this anymore - but might be best to just archive it in case others are actually using it even if the EKS module is not

antonbabenko:

lodotek:
I am trying to find out why `docs/iam-permissions.md` is missing from master. Because it is now missing from master, the there is a broken link here: https://learn.hashicorp.com/tutorials/terraform/eks 
image

PR #1695, opened 2021-11-17
Support bottlerocket bootstrapping for node groups

### LINKED ISSUE #1694: Support bottlerocket in node groups

Currently when `ami_type = "BOTTLEROCKET_x86_64` is passed, the bootstrapping process is not successfully completed due to the lack of toml config for the given ami type

Support bottlerocket_* for node groups.

ulm0:
# PR o'clock

## Description

Fixes #1694 

### Checklist

### DIFF
--- a/modules/node_groups/launch_template.tf
+++ b/modules/node_groups/launch_template.tf
@@ -1,5 +1,5 @@
 data "cloudinit_config" "workers_userdata" {
-  for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] && v["ami_type"] != "BOTTLEROCKET_x86_64" }
+  for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] && length(split("BOTTLEROCKET", v["ami_type"])) == 0 }
 
   gzip          = false
   base64_encode = true
@@ -25,7 +25,7 @@ data "cloudinit_config" "workers_userdata" {
 }
 
 data "template_file" "bottlerocket_workers_userdata" {
-  for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] && v["ami_type"] == "BOTTLEROCKET_x86_64" }
+  for_each = { for k, v in local.node_groups_expanded : k => v if v["create_launch_template"] && length(split("BOTTLEROCKET", v["ami_type"])) > 1 }
 
   template = file("${path.module}/templates/userdata.toml.tpl")
   vars = {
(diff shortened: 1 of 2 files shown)


hi-artem:
Doesn't bottlerocket ami ships with 2 ebs volumes? I think you will need to add another `block_device_mappings` in the launch template. Something like that:

```hcl
 block_device_mappings {
    device_name = "/dev/xvdb"

    ebs {
      volume_size           = lookup(each.value, "disk_size", null)
      volume_type           = lookup(each.value, "disk_type", null)
      iops                  = lookup(each.value, "disk_iops", null)
      throughput            = lookup(each.value, "disk_throughput", null)
      encrypted             = lookup(each.value, "disk_encrypted", null)
      kms_key_id            = lookup(each.value, "disk_kms_key_id", null)
      delete_on_termination = true
    }
  }
```

hi-artem (changes requested):
Need to address issue with 2 volumes. See my comment above.

hi-artem:
To simplify logic here, we can only care about `/dev/xvdb` sizing.`/dev/xvda` contains operating system itself: the active & passive partitions, the bootloader, the dm-verity data, and the data store for the Bottlerocket API. We really shouldn't set its sizing explicitly. It should always be the same as in Bottlerocket AMI.

johndietz (approved):

ulm0:
Nice finding @hi-artem will address this topic asap.

ulm0:
Hi @hi-artem just added

`device_name = length(split("BOTTLEROCKET", each.value["ami_type"])) > 1 ? "/dev/xvdb" : "/dev/xvda"`

This sould do the work

modules/node_groups/launch_template.tf
```diff
@@ -24,6 +24,20 @@ data "cloudinit_config" "workers_userdata" {
   }
 }
 
+data "template_file" "bottlerocket_workers_userdata" {
```
jlamande, line 27: template_file is a deprecated provider.

It is better to use the templatefile function.

jlamande: thansks @ulm0 ! So fast ;)

I got an issue with this provider a few days ago as it is not available for arm platforms and won't be (deprecation notice) :D

ulm0: just did a minor fix, should be good to go now @jlamande

ulm0:
Hi @johndietz just did some changes, the PR should be now good to go

stevehipwell:
@ulm0 could you take a look at #1645 and offer feedback; I think it's a more generic solution but some things might be missing?

dusansusic:
@ulm0 AWS provider must be bumped to `>= 3.64`.

https://github.com/hashicorp/terraform-provider-aws/releases/tag/v3.64.0

antonbabenko:

PR #1699, opened 2021-11-23
chore: change default aws authenticator command to aws eks get-token instead of aws-iam-authenticator

### LINKED ISSUE #957: Use "aws eks get-token" instead of "aws-iam-authenticator"
## Use `aws eks get-token` instead of `aws-iam-authenticator`
Since AWS CLI version 1.16.156, `aws-iam-authenticator` is no longer needed to obtain an auth token for the EKS cluster. This can be performed using `awscli` directly with `aws eks get-token`. [(reference)](https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html)

I propose changing the default value of `kubeconfig_aws_authenticator_command` and `kubeconfig_aws_authenticator_command_args` to:
```hcl
  kubeconfig_aws_authenticator_command = "aws"
  kubeconfig_aws_authenticator_command_args = [
    "--region", 
    data.aws_region.this.name,
    "eks",
    "get-token",
    "--cluster-name",
    local.cluster_name
  ]
}
```

## I'm submitting a...

* [ ] bug report
* [x] feature request

External dependency on `aws-iam-authenticator` tool causes `terraform apply` to fail when not installed.
This dependency is not clearly indicated in the readme.

Using this module without having installed `aws-iam-authenticator` will throw an error while provisioning the cluster.

No external dependency on `aws-iam-authenticator`.

Yes

## Environment details

* Affected module version: **v10.0.0**
* OS: **Mac OS Catalina 10.15.4**
* Terraform version: **v0.12.26**

## Any other relevant info
I'm not sure which tool is more common for this purpose, but our team uses `awscli` whenever possible. Since this functionality is now present in `awscli >= 1.16.156` (we are using `v2.0.19`), it might be worthwhile to use it instead of `aws-iam-authenticator`.

In any case, whether depending on `awscli` or `aws-iam-authenticator`, it might be a good idea to clearly indicate external dependencies in the module readme.

lisfo4ka:
## Description
This PR changes default aws authenticator command to `aws eks get-token` instead of `aws-iam-authenticator` since it's already deprecated. Fixes #957.

The next step is to drop `aws-iam-authenticator` support from the terraform-aws-eks module.

## Motivation and Context
Fixes #957.

## Breaking Changes
No breaking changes here, just kubconfig file updates.

## How Has This Been Tested?
EKS cluster was deployed from the master branch and updated with the suggested changes successfully.
Pre-commit hook was run.

### DIFF
--- a/data.tf
+++ b/data.tf
@@ -2,6 +2,8 @@ data "aws_partition" "current" {}
 
 data "aws_caller_identity" "current" {}
 
+data "aws_region" "current" {}
+
 data "aws_iam_policy_document" "workers_assume_role_policy" {
   statement {
     sid = "EKSWorkerAssumeRole"
--- a/locals.tf
+++ b/locals.tf
@@ -162,13 +162,22 @@ locals {
     "t2.xlarge"
   ]
 
+  default_kubeconfig_aws_authenticator_command_args = [
+    "eks",
+    "get-token",
+    "--cluster-name",
+    local.cluster_name,
+    "--region",
+    data.aws_region.current.name
+  ]
+
   kubeconfig = var.create_eks ? templatefile("${path.module}/templates/kubeconfig.tpl", {
     kubeconfig_name                         = coalesce(var.kubeconfig_name, "eks_${var.cluster_name}")
     endpoint                                = local.cluster_endpoint
     cluster_auth_base64                     = local.cluster_auth_base64
     aws_authenticator_kubeconfig_apiversion = var.kubeconfig_api_version
     aws_authenticator_command               = var.kubeconfig_aws_authenticator_command
-    aws_authenticator_command_args          = coalescelist(var.kubeconfig_aws_authenticator_command_args, ["token", "-i", local.cluster_name])
+    aws_authenticator_command_args          = coalescelist(var.kubeconfig_aws_authenticator_command_args, local.default_kubeconfig_aws_authenticator_command_args)
     aws_authenticator_additional_args       = var.kubeconfig_aws_authenticator_additional_args
     aws_authenticator_env_variables         = var.kubeconfig_aws_authenticator_env_variables
   }) : ""
--- a/variables.tf
+++ b/variables.tf
@@ -200,17 +200,17 @@ variable "kubeconfig_api_version" {
 variable "kubeconfig_aws_authenticator_command" {
   description = "Command to use to fetch AWS EKS credentials."
   type        = string
-  default     = "aws-iam-authenticator"
+  default     = "aws"
 }
 
 variable "kubeconfig_aws_authenticator_command_args" {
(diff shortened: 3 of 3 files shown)


schollii:
LGTM

bryantbiggs:
this all looks good and makes sense to me, but unfortunately I think its still a breaking change because we have changed the default behavior. are we able to keep the current behavior (aws-iam-authenticator) as the default, but provide users the ability to use the aws eks get-token if they so choose? I think that would make this change very easy to merge in without disruption

schollii:
I guess it will be a breaking change for those who had customized their iam authenticator args.

lisfo4ka:
At the moment, it's already possible to use aws eks get-token by defining `kubeconfig_aws_authenticator_command` and `kubeconfig_aws_authenticator_command_args` input variables.
So it doesn't require any new changes to release that. Something like that:
```
  kubeconfig_aws_authenticator_command         = var.kubeconfig_aws_authenticator_command
  kubeconfig_aws_authenticator_command_args    = var.kubeconfig_aws_authenticator_command == "aws" ? ["eks", "get-token", "--cluster-name", element(concat(data.aws_eks_cluster_auth.cluster[*].name, [""]), 0)] : []
  kubeconfig_aws_authenticator_additional_args = ["--role-arn", "arn:aws:iam::012345678910:role/EKSDeployerRole"]
  
```

But I believe that everyone who keep `terraform-aws-eks` module up to date and use v.17 have already updated awc cli to the version 1.16.156 or later since it was introduced 3 years ago.
For that users there are two possible cases:

- they will see updates in the `kubeconfig` file if they've used default `aws-iam-authenticator`, but still can apply the changes and authenticate to the cluster with no additional steps.
If they intentionally want to use `use aws-iam-authenticator` there is still possible to re-define defaults. 

- they've already defined specified `kubeconfig_aws_authenticator` variables to use `eks get-token` and it will take precedence over our changes. As a result, terraform wouldn't suggest any changes at all.
I believe that is our target audience for the moment.

In case, the user use aws cli version before 1.16.156 it will really be a breaking change. But I'm sure there are no many people who use it. And we shouldn't focus on this since it's too outdated.

So I agree that it can lead to some changes during terraform module minor version upgrade. But I think it's quite enough to leave some details in the `CHANGELOG.md` to notify everybody about possible changes.
The official AWS documentation proposes to use native `aws eks get-token` auth instead of `aws-iam-authenticator` tool, and our default values for `aws-iam-authenticator` is really confusing.
https://docs.aws.amazon.com/eks/latest/userguide/managing-auth.html
https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html

From my side, I'd prefer to introduce this before v18 since there would be too many breaking changes and not everybody are ready to install them.

antonbabenko:

PR #1700, opened 2021-11-24
add missing "bootstrap_extra_args"

shaunc:
## Description

"bootstrap_extra_args" is present in userdata.sh.tpl, and in locals.tf, but not threaded through.

## Motivation and Context

Specified bootstrap_extra_args not passed in userdata.

## How Has This Been Tested?


shaunc:
Hmm... well not in node_groups/locals.tf either .... I'm not sure if you want to support this? I think I have a workaround so will close for the moment.

PR #1703, opened 2021-11-25
Ignore subnets

tanjarinne:


daroga0002:
what is purpose of this PR? what issue this solves?

tanjarinne:
Closing. Pointed it to the wrong repository. Was supposed to point to our fork.
We're having a situation where changes in the subnets force our clusters to be recreating, so we're ignoring that.

PR #1705, opened 2021-11-25
feat: Add ability for tagging AWS-generated ASGs when using managed node groups

gabops:
## Description
- Add ability for tagging AWS generated autoscaling groups when using managed node groups

## Motivation and Context
- This change comes from the personal need of having the asg tagged in a certain way for allowing cluster-autoscaler to increase managed node groups when they are set initially to 0. However, this change also allows to tag the asgs in any other required case.

## Breaking Changes
None as the module behaves the same when not using this functionality.

## How Has This Been Tested?
- ✅ I have tested and validate these changes by using this module with the proposed changes over my AWS account from a personal Terraform project.

### DIFF
--- a/examples/managed_node_groups/main.tf
+++ b/examples/managed_node_groups/main.tf
@@ -42,6 +42,9 @@ module "eks" {
         GithubRepo = "terraform-aws-eks"
         GithubOrg  = "terraform-aws-modules"
       }
+      asg_tags = {
+        "k8s.io/cluster-autoscaler/node-template/label/monitoring" = true
+      }
       additional_tags = {
         ExtraTag = "example"
       }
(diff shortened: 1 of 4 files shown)


antonbabenko:

weixianearmap:
@antonbabenko  any doc for it? seems still not working in version 18.0.6. When I use the var `tags`, the tags of the autocaling group still doesn't have these tags.

antonbabenko:
@weixianearmap Please see the code in `examples` for more details, and open a separate issue (with complete details provided) if there is a bug or a problem.

weixianearmap:
@antonbabenko  Sorry, I can't find any examples which has asg_tags demo in master branch. I don't think this pr has been merged. That's why I made the comment here.

antonbabenko:
This PR was not merged because the same functionality was introduced in #1680. You can search for `tags` in `examples`.

PR #1712, opened 2021-12-06
Accurate description for workers_additional_policies

schwichti:
## Description
* Description should mention it is a list of ARNs
* It seems that `workers_additional_policies` are also effective for managed nodes (needs confirmation)
* When you pass a custom role for your worker/managed node, it seems that the additional policies are overridden (needs confirmation)

## Motivation and Context
I think the description is not precise.

## Breaking Changes
None.

## How Has This Been Tested?
Haven't tested.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -187,7 +187,7 @@ variable "worker_sg_ingress_from_port" {
 }
 
 variable "workers_additional_policies" {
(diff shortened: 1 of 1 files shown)


bryantbiggs:
closed with #1680

PR #1715, opened 2021-12-12
chore: Added action to validate PR title

antonbabenko:


antonbabenko:
@bryantbiggs [Semantic PR](https://github.com/zeke/semantic-pull-requests#how-it-works) was too simple and often misleading.

The one I am adding here is checking the PR title and printing the reason for the failure. Let me know what do you think about this (check it in action in another existing PR after it is rebased because this check will be using configuration from the `master` branch).

antonbabenko:
Apparently "BREAKING CHANGE" is really [breaking everything](https://github.com/terraform-aws-modules/terraform-aws-eks/runs/4499681560?check_suite_focus=true#step:2:28). I will look into this somewhere else in a separate repo.

bryantbiggs:
this looks great. we can also add one additional step that will show us what the next release version is slated to be by adding

```yaml
	- name: Release check
        uses: cycjimmy/semantic-release-action@v2
        with:
          dry_run: true
          semantic_version: 18.0.0
          extra_plugins: |
            @semantic-release/changelog@6.0.0
            @semantic-release/git@10.0.0
        env:
          GITHUB_TOKEN: ${{ secrets.SEMANTIC_RELEASE_TOKEN }}
```

its essentially the same release step we already use, but by adding the `dry_run: true` it will not make any release changes and instead just output what the next version and changelog entry would be for the PR - Ref https://github.com/cycjimmy/semantic-release-action#dry_run

antonbabenko:
Good, I will update the EKS module.

`dry_run` does not help so much, because the release version will depend on the description of the PR during the merge. If it matches `r/BREAKING CHANGE[s]*/g` then it will be a major release independently of a type (feat, fix, etc...).

antonbabenko:

PR #1717, opened 2021-12-12
fix: Correct DNS Server of kubelet service when using custom AMI

pjrm:
Correct DNS Server of kubelet service when using custom AMI that is EKS optimized and has a custom ipv4 CIDR

## Description
The PR (#1580) is passing the "apiserver-endpoint" and "b64-cluster-ca", which causes the SERVICE_IPV4_CIDR empty (https://github.com/awslabs/amazon-eks-ami/blob/v20211206/files/bootstrap.sh#L366). Because of that, the script fallbacks always to 10.100.0.10 or 172.20.0.10.

Defining the ipv4 cidr ensures that the bootstrap script configures the DNS server correctly on the kubelet service, allowing pods to resolve DNS names.

## Breaking Changes
No

## How Has This Been Tested?
By running terraform module in the AWS


daroga0002:
@bryantbiggs  can we add this into your breaking change PR?

bryantbiggs:
@daroga0002 I believe this is already covered here https://github.com/bryantbiggs/terraform-aws-eks/blob/refactor/breaking-changes/examples/user_data/main.tf#L41 (please correct me if I am wrong)

users should be able to provide any necessary env vars/configs/etc. prior to bootstrapping via the generic `pre_bootstrap_user_data` variable which saves us from having to add multiple, specific variables where we have to check whether or not they are used/provided/etc. 

so for example, with a config such as:

```hcl
	...
	pre_bootstrap_user_data = <<-EOT
	export SERVICE_IPV4_CIDR=172.16.0.0/16
	EOT
	...
```

I believe this will suffice for this issue (and likewise for any future request where `SERVICE_IPV6_CIDR` would be desired as well)

bryantbiggs:
ah I see - I missed that `cluster_service_ipv4_cidr` is a variable that we have in the module root; I thought it was a net new addition in this PR. I see what you mean, I think that makes sense and yes, we want to have a good "out of box" experience as much as possible (without introducing cumbersome complexities). I can add this change now

@antonbabenko - lets hold on the EKS breaking change a sec, I'll add this modification to the PR now

daroga0002:
> users should be able to provide any necessary env vars/configs/etc. prior to bootstrapping via the generic `pre_bootstrap_user_data` variable which saves us from having to add multiple

Yup, it will work but now is the question do we should make the same consistent approach with:
https://github.com/bryantbiggs/terraform-aws-eks/blob/ef0e59859cd4eada948dbe7bbf73b36682322ddb/templates/bottlerocket_user_data.tpl#L3-L5

https://github.com/bryantbiggs/terraform-aws-eks/blob/ef0e59859cd4eada948dbe7bbf73b36682322ddb/templates/linux_user_data.tpl#L7-L8

??


Just thinking about as we passing some variables and some not 🤔

bryantbiggs:
I say we make them consistent. Because its a setting that is set at the cluster level, I think all routes need to support this if a value is present (both EKS and Self managed node groups, as well as all OS flavors) 

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#service_ipv4_cidr

bryantbiggs:
doesn't look like Bottlerocket supports it yet https://github.com/bottlerocket-os/bottlerocket/blob/3a2e65fde1f16d56d0d0e681231f7d7f0a8acb05/sources/api/pluto/src/main.rs#L216

bryantbiggs:
Ref https://github.com/bottlerocket-os/bottlerocket/issues/1866

bryantbiggs:
and it looks like Windows based nodes do not support this either https://github.com/awslabs/amazon-eks-ami/issues/805 

so I will just add it for the AWS EKS Optimized AMI for now. This change isn't breaking so if support for Bottlerocket and/or Windows is added in the future, it should be an easy feature change without disruption

daroga0002:
> I say we make them consistent.

the second solution is that we can also make it consistent by removing them (as they are optional and bootstrap eks script can gather that info itself).

Of course both has pros and cons

bryantbiggs:
spinning up a cluster now to verify but this should do it. I added some notes on the user data module in case others come and request similar, we can quickly check if those issues are resolved to support Bottlerocket and Windows

https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1680/commits/0c2a270abc4599639743ee888da4310b982250fc

bryantbiggs:
ok this all looks good that was added/changed [here](https://github.com/terraform-aws-modules/terraform-aws-eks/commit/0c2a270abc4599639743ee888da4310b982250fc)

image

pjrm:
Thanks, @bryantbiggs, and @daroga0002, for applying the fix!

Another issue that I think we have is that the map variable "bootstrap_env" assigns the environment variables consumed by the bootstrap.sh script, but since we are not exporting all variables by using `set -a', we should also export those variables.

So we should do something like:
```
%{ for k, v in bootstrap_env ~}
export ${k}="${v}"
%{ endfor ~}
```

Another option is to use at the beginning of the script, and we can export all variables out-of-the-box:
`set -a'

Or I'm doing something wrong?

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/node_groups/templates/userdata.sh.tpl#L22-L24

bryantbiggs:
hi @pjrm  - I am not sure I quite follow. what is the issue you are seeing?

pjrm:
Hi @bryantbiggs,

My point is, imagine this use case of SERVICE_IPV4_CIDR, assuming that we didn't apply this fix. If we use the bootstrap_env variable, the bootstrap script will not consume that variable.sh, because it's another bash run. But, on the other hand, if we export the variable, we are telling that variable is available inside the user-data script and in all programs that we invoke inside it.

In other words, if the objective of `bootstrap_env` is to pass environment variables to the bootstrap script, we should export those variables or tell bash to export all variables automatically (set -a option). I prefer the first option tho.

bryantbiggs:
hi @pjrm I think there is a bit of a misunderstanding. we have three possible cases happening in the current user data
1. We do [NOT export variables that are being passed as arguments directly to the bootstrap.sh script](https://github.com/bryantbiggs/terraform-aws-eks/blob/d2fa4d2d7aa80a8c34c5396e9786ea39c14cd6fc/templates/linux_user_data.tpl#L10-L12). These are therefore not available after this script ends, but they are passed to the bootstrap script since they are passed as arguments
2. We [DO explicitly export `SERVICE_IPV4_CIDR`](https://github.com/bryantbiggs/terraform-aws-eks/blob/d2fa4d2d7aa80a8c34c5396e9786ea39c14cd6fc/templates/linux_user_data.tpl#L7) because it is not passed as a script argument and instead it needs to be populated into the `env` so that it can be referenced within the scope of the "child" bootstrap.sh script
3. Lastly, we have a [pre and post user data setting](https://github.com/bryantbiggs/terraform-aws-eks/blob/d2fa4d2d7aa80a8c34c5396e9786ea39c14cd6fc/templates/linux_user_data.tpl#L5) that allows users to inject their own settings as desired. if users want to inject `set -a` or export their own variables, they can do so through these general purpose configuration settings

Not sure if that helps clear it up. not exporting is, in my opinion, the preferred route since it will avoid polluting the environment with unnecessary variables. Because `SERVICE_IPV4_CIDR` is optional, and if not provided then it is not rendered in the user data, and because there isn't a specific script parameter for this (only `--dns-cluster-ip`), we instead elect to export it IF the value has been set

kkapoor1987:
@pjrm @bryantbiggs We are hitting the same issue. I have created issue #1726 to track it. 
Do you guys plan to merge this anytime soon ?Please let me know if you need help on this I can test or push the fix if needed

pjrm:
Hi @bryantbiggs,

Yes, it makes sense. If we export the variables, they are only available in the user_data script. However, from what I understood, it's expected that bootstrap_env is only for passing variables as arguments.

I was thinking of edge cases, like SERVICE_IPV4_CIDR and other variables only exposed via environment variables.

@kkapoor1987 I think we will not apply the change because the idea is to pass those variables into bootstrap arguments, right @bryantbiggs ?

And thanks for the feedback :)

antonbabenko:

Issue #1714, opened 2021-12-08
node_groups/launch_template.tf needs to be more flexible on the device_name within block_device_mappings section


kongdewen:
## Description
First, thank you so much to enable custom AMI image in the node_groups launch templates. But it also can cause a problem when custom AMI image root volume name isn't `/dev/xvda` as hardcoded in https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/node_groups/launch_template.tf#L41. it will cause consumers can't setup correct root devices block_device_mappings when we launch the EC2 instance. 

## Reproduction
To reproduce the issue, we can launch an custom AMI image with root device name differ than `/dev/xvda`

## Expected behavior
we should be able to setup the ami root device name based on the AMI image itself

## Actual behavior
the root device name is fixed to `/dev/xvda`

bryantbiggs:
closed with #1680

PR #1721, opened 2021-12-17
improvement: match block_device_mappings with workers_launch_template in node group 

kongdewen:
## Description
In https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/workers_launch_template.tf, the module support `root_block_device_name ` as variable, and also added support for `additional_ebs_volumes` and `additional_instance_store_volumes`, It will be nice if we can match the same flexibility as we support node groups as now we are supporting custom AMI image. 

## Motivation and Context
It starts with issue fix: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1714.  Thought why don't we want to do more

## Breaking Changes
N/A

## How Has This Been Tested?
We have a batch version fixed the issue(https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1714) deployed in our prod account.
for `additional_ebs_volumes`, `additional_instance_store_volumes`, we don't have any usage yet, we only did `terraform plan` for test scenario.


antonbabenko:

PR #1728, opened 2021-12-21
docs: Update complete example to include aliased providers.

wesleyorama-grindr:
## Description
Added some documentation around using aliased providers to the complete example. This took a while for me to find / figure out so I thought it would be helpful to put it in the example. 

## Motivation and Context

This solved the issue where this solution was poorly documented. I found this fix by digging though old closed issues. 

## Breaking Changes

No

## How Has This Been Tested?

This is currently deployed in this manner in our dev environment.

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -14,6 +14,11 @@ locals {
 
 module "eks" {
   source = "../.."
+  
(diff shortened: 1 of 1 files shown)


bryantbiggs:
Thank you but I think this is more generic to terraform usage and not adding a lot of value directly to this module. Thank you though!

PR #1736, opened 2022-01-05
fix: Update preset rule on semantic-release to use conventional commits

bryantbiggs:
## Description
- update preset rule on semantic-release to use conventional commits

Note: this should re-scan the commit since the last release and pick up the breaking `feat!` change causing v18.x to be released

## Motivation and Context
- current release process is not triggering a breaking change release due to using default `angular` preset

## Breaking Changes
- no

## How Has This Been Tested?
- https://github.com/clowdhaus/terraform-release-test/releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ on:
       - '**/*.tpl'
       - '**/*.py'
       - '**/*.tf'
+      - '.github/workflows/release.yml'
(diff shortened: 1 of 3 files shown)


antonbabenko:

antonbabenko:
Fantastic! It is so magical but it works when it works! Thanks for this fix. 👍

PR #1743, opened 2022-01-06
fix: Not to iterate over remote_access object in dynamic block

knight42:
## Motivation and Context

Currently if user specifies `remote_access` in `eks_managed_node_groups`, terraform will throw an error since we iterate the `remote_access` object directly, which is not expected.

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -295,10 +295,10 @@ resource "aws_eks_node_group" "this" {
   }
 
   dynamic "remote_access" {
-    for_each = var.remote_access
+    for_each = length(var.remote_access) > 0 ? [var.remote_access] : []
     content {
-      ec2_ssh_key               = lookup(remote_access.value, "ec2_ssh_key", null)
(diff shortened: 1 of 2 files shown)


modules/eks-managed-node-group/main.tf
```diff
@@ -295,7 +295,7 @@ resource "aws_eks_node_group" "this" {
   }
 
   dynamic "remote_access" {
-    for_each = var.remote_access
+    for_each = length(var.remote_access) > 0 ? [var.remote_access] : []
```
bryantbiggs, line 298: just testing locally looks like we also need to change lines 300-301 to:

```hcl
	ec2_ssh_key               = try(remote_access.value.ec2_ssh_key, null)
	source_security_group_ids = try(remote_access.value.source_security_group_ids, [])
```

knight42: Nice catch! I have applied your fix. Thanks!

bryantbiggs (approved):
@antonbabenko 👍🏽 

thanks for the patch @knight42 🎉

antonbabenko:
This PR is included in version 18.0.4 :tada:

PR #1745, opened 2022-01-06
fix: Remove trailing hyphen from cluster security group and iam role name prefix

imdevin567:
## Description
A small naming change to prevent downstream recreation of resources.

## Motivation and Context
Clusters created with previous versions of this module that relied on the generated cluster security group and/or cluster iam role are forced to be recreated when upgrading to version 18. The previous version used the cluster name as a prefix for both the SG and role. While you can override these now with `cluster_security_group_name` and `iam_role_name`, the module appends a hyphen to these values, which makes it impossible to set to the previous value.

Renaming a security group or an IAM role causes a recreation of the resource; _changing_ the security group or IAM role for an EKS cluster forces a recreation of the cluster.

Removing this hyphen still allows the user to append one if they want to, but doesn't force it, making the upgrade smoother.

## Breaking Changes
No breaking change

## How Has This Been Tested?


bryantbiggs:
I will let others chime in, but this is the defacto standard that we have adopted across the modules and the breaking nature of the change is why we have introduced it in the v18.x breaking change

antonbabenko:
Well, just a day has passed since we released 18.0. I think we can bump 18.1.0 with this PR merged. We just need to ask people to use the latest version.

Are there any other similar changes we should be doing?

imdevin567:
@antonbabenko Not a breaking change per se, but the default description of the security group was changed from: `EKS cluster security group.` to `EKS cluster security group` (period removed). This causes a recreation of the resource (which causes a downstream recreation of the cluster), but this was easily circumvented by just adding: `cluster_security_group_description = "EKS cluster security group."`.

In addition, the resource rename of `aws_iam_role.cluster` to `aws_iam_role.this` causes a recreation of the resource, but this can be worked around by moving the resource prior to applying:

```
tf state mv 'module.eks.aws_iam_role.cluster[0]' 'module.eks.aws_iam_role.this[0]'
```

IMO as long as there is a workaround to prevent cluster recreation it's fine. While I understand this is a breaking change by nature of the v18.x bump, I suspect most people will expect to be able to upgrade this module without a _requirement_ of needing to tear down their existing cluster.

bryantbiggs:
@imdevin567 users can elect to manage/create those roles and security groups outside of the EKS module and then pass them in - therefore, users still have the ability to avoid re-creation disruptions

jseiser:
If you are defaulting to things that cause a cluster re-creation, their probably needs to be really good documentation on what is required to prevent that from happening.

imdevin567:
@bryantbiggs yes, but roles/security groups created by the module previously have dates appended to the names: i.e. `develop-eks-cluster2022010514434579540000000b`. The auto-generated names don't lend themselves well to being "migrated" to a managed resource.

As @jseiser said above, if we're going to leave this as the default then we should add more to the upgrade guide to prevent this from happening in the first place.

imdevin567:
If we want to leave the standard in place, would it be better to just add two variables?

- `iam_role_name_prefix`
- `cluster_security_group_name_prefix`

This would decouple the name from the name prefix, allowing users to override either.

bryantbiggs:
@jseiser we have documentation, if there are pieces missing then PRs are welcome https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/UPGRADE-18.0.md

Again, a lot of the changes made in v18.x are intended to hopefully simplify and stabilize the module. This module has been notoriously complex leading to very difficult maintenance and troubleshooting (https://github.com/terraform-aws-modules/terraform-aws-eks/issues/635). 

There are always ways around mitigating resource disruption depending on the level of user experience. The external role and security group creation that I am referring to @imdevin567  is not via imports but creating fresh new roles/security groups and then swapping those in to the EKS module. This does not result in any downtime or disruption and can potentially save you any additional headaches in the future

imdevin567:
@bryantbiggs correct me if I'm wrong, but you can't swap out a cluster role or security group without recreating the cluster. This was the original reason for the PR.

bryantbiggs:
the security group you should be able to since its an additional security group. I can't for certain on the IAM role swap as I have not attempted it myself

jseiser:
> @jseiser we have documentation, if there are pieces missing then PRs are welcome https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/UPGRADE-18.0.md


Considering I have yet to be able to follow you upgrade guide and end up with a working cluster, i do not think ill be much help adding to the documentation.  So our enterprise is hard stuck on <18 for the time being.

bryantbiggs:
> > @jseiser we have documentation, if there are pieces missing then PRs are welcome [`master`/UPGRADE-18.0.md](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/UPGRADE-18.0.md?rgh-link-date=2022-01-06T19%3A29%3A09Z)
> 
> Considering I have yet to be able to follow your upgrade guide and end up with a working cluster, i do not think ill be much help adding to the documentation. So our enterprise is hard stuck on <18 for the time being.

You or your enterprise can reach out and contribute a sponsorship if you would like, otherwise please feel free to enjoy the free OSS

antonbabenko:
I think we should add a couple of new variables as per this comment by @imdevin567 - https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1745#issuecomment-1006838822 

So that people in the same situation as @imdevin567 can provide the values to prevent an unnecessary recreation of resources.

@bryantbiggs agree?

imdevin567:
I can make those changes to this PR pending a thumbs up from @bryantbiggs 

I'll submit a separate PR with additions to the upgrade guide for steps I needed to work through to avoid cluster recreation.

jseiser:
> > > @jseiser we have documentation, if there are pieces missing then PRs are welcome [`master`/UPGRADE-18.0.md](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/UPGRADE-18.0.md?rgh-link-date=2022-01-06T19%3A29%3A09Z)
> > 
> > 
> > Considering I have yet to be able to follow your upgrade guide and end up with a working cluster, i do not think ill be much help adding to the documentation. So our enterprise is hard stuck on <18 for the time being.
> 
> You or your enterprise can reach out and contribute a sponsorship if you would like, otherwise please feel free to enjoy the free OSS

You really think that is an appropriate response?

bryantbiggs:
> I think we should add a couple of new variables as per this comment by @imdevin567 - [#1745 (comment)](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1745#issuecomment-1006838822)
> 
> So that people in the same situation as @imdevin567 can provide the values to prevent an unnecessary recreation of resources.
> 
> @bryantbiggs agree?

are we doing this for all occurrences of both IAM role name and security group name?
- `eks-managed-node-group` sub-module
- `self-managed_node-group` sub-module
- `fargate-profile` sub-module
- root module

antonbabenko:
@bryantbiggs I am thinking only about places where it actually changed between v17 and v18.

bryantbiggs:
if we are only looking at the cluster IAM role and security group then we could probably get away with a patch variable like `prefix_separator` that defaults to `"-"` which is the desired state but users can use `""` which reflects the v17.x state. does this work? shouldn't make things messy and would be quite clear why its used. We'll also have to update:
- `"${local.cluster_sg_name}-"` -> `"${local.cluster_sg_name}${var.prefix_separator}"`
- `"${local.iam_role_name}-"` -> `"${local.iam_role_name}${var.prefix_separator}"`

imdevin567:
Works for me. I'll add the change.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+### [18.0.1](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v18.0.0...v18.0.1) (2022-01-06)
```
bryantbiggs, line 5: we can leave out any changes here, the release workflow will automatically update this

imdevin567: My bad, needed to pull from upstream.

UPGRADE-18.0.md
```diff
@@ -28,6 +28,8 @@ Please consult the `examples` directory for reference example configurations. If
...
+- The `aws_eks_cluster` resource has been removed and replaced with the `aws_eks_cluster_role_policy_attachment` resource. This change is to align with the [AWS EKS API changes](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html) and [AWS EKS documentation](https://docs.aws.amazon.com/eks/latest/userguide/create-cluster.html)
```
bryantbiggs, line 31: is this line 31 change intentional?

CHANGELOG.md
bryantbiggs: ah, I see - no worries!

UPGRADE-18.0.md
imdevin567: hahaha nope, that would be Copilot chiming in. Don't hit tab too quickly.

bryantbiggs (approved):
@antonbabenko 👍🏽 

thanks for your effort on this fix @imdevin567 🎉

antonbabenko:
This PR is included in version 18.0.3 :tada:

PR #1754, opened 2022-01-08
fix: Set default cluster sg description eq 17.24

martijnvdp:
## Description
Set the default value of the cluster security group description eq to the default value of 17.24

## Motivation and Context
a change in the description of the cluster security group forces recreation of the security group as well as the eks cluster
as a change of the cluster security group triggers the replacement of the eks cluster
#1744 
## How Has This Been Tested?

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -154,7 +154,7 @@ variable "cluster_security_group_use_name_prefix" {
 variable "cluster_security_group_description" {
   description = "Description of the cluster security group created"
   type        = string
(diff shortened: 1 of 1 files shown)


bryantbiggs:
This is already a variable so users are free to add in the period if desired. Descriptions are not completely, grammatically correct sentences and so periods are not useful here. This was introduced at a breaking change so we are not trying to maintain the legacy compatibility

PR #1759, opened 2022-01-10
feat: Add support for networking `ip_family` which enables support for IPV6

### LINKED ISSUE #1746: Support creating IPv6 EKS clusters

Yes. I doubt it's in the provider as the CLI arguments aren't even documented yet. But we should have an issue to track this.

I'd like to be able to create IPv6 enabled clusters as described in the [AWS news announcement](https://aws.amazon.com/blogs/aws/amazon-elastic-kubernetes-service-adds-ipv6-networking/).

I'd like this option supported as soon as the provider exposes it.

N/A

## Additional context

N/A

bryantbiggs:
## Description
- add support for networking `ip_family` which enables support for IPV6
- add policy for `AmazonEKS_CNI_IPv6_Policy` for nodes using VPC CNI to connect when IPV6 is enabled
	- propagate `cluster_ip_family` down to sub-modules to be able to determine which CNI policy should be used
	- all sub-modules now utilize the `iam_role_attach_cni_policy` variable to control whether the CNI policy should be attached or not
- bump Terraform AWS provider min version required to support `ip_family` attribute
	- https://github.com/hashicorp/terraform-provider-aws/pull/22485
- update `eks-managed-node-group` example to demonstrate/validate use of IPV6
	
Note: IPV6 can only be enabled on new clusters, not existing. So the addition of the IPV6 rule on the security groups is non-breaking (does not affect current clusters either)

## Motivation and Context
- Closes #1746 

## Breaking Changes
- No

## How Has This Been Tested?
	- Validated using `eks-managed-node-group` example


stevehipwell:
@bryantbiggs this looks great to me, now that the AWS provider v3.72.0 do you anticipate this can be released soon?

bryantbiggs:
@stevehipwell yes! I just wanted to have at least one deploy to ensure it works as advertised. Just launched a cluster this morning and everythings looking good 
image

stevehipwell:
@bryantbiggs I assume you're using the latest cordns version (`v1.8.4`), are you using the add-in or manually managing it? I don't think there was anything in the IPv6 announcement about coredns specifically.

Have you looked at the bootstrap options to see if there is an issue with the old logic?

bryantbiggs:
> @bryantbiggs I assume you're using the latest cordns version (`v1.8.4`), are you using the add-in or manually managing it? I don't think there was anything in the IPv6 announcement about coredns specifically.
> 
> Have you looked at the bootstrap options to see if there is an issue with the old logic?

Ya, the addons are all pulling the latest versions that are specified in the announcement. I assume there are no changes in the bootstrap logic (I hope) - I would presume this is all handled in the `bootstrap.sh` script, I didn't see any mention in the docs/announcements

bryantbiggs:
> @bryantbiggs We observed a while back that when launching a new cluster, compute nodes need to be available before the CoreDNS plugin will launch successfully. Since we manage the cluster addons outside of the module we solved this by simply putting a `depends_on = [module.eks]` in our case, as that will force it to run after the nodegroup in the module finished creating. Is it possible that you're hitting the same wall? You could try launching the cluster without the addon first and then add it back to see if that works.
> 
> Note that this happens on IPv4 clusters as well, so it's not related to your changes here.

Yes, this works today in the `eks-managed-node-group` - the addons are the last bit and the node groups in that example all come up; the CoreDNS plugin takes the longest but it eventually turns healthy all in one `terraform apply`. I may tear down the cluster and have another go because I did have to update the VPC settings a bit (add egress only gateway and assign IPV6 on launch) and update the node security group for IPV6 egress

bryantbiggs:
Between this module and MSK, talk about longest turn around time to test out changes 😅 🕐

bryantbiggs:
ah, might be a permissions issue https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html#cni-iam-role-create-ipv6-policy

bryantbiggs:
Policy addition fixed it - should be all set for review now
image

antonbabenko:
This PR is included in version 18.1.0 :tada:

PR #1762, opened 2022-01-11
fix: Use split function to allow setting multiple source security group ids for remote_access

cbirajdar:
## Description

Creating a new managed node group in EKS using the `eks-managed-node-group` submodule when remote_access.{ec2_ssh_key, source_security_group_ids} values are set, fails with following error.

```
Error: Incorrect attribute value type

  on .terraform/modules/eks_devops_dev.eks_node_group.eks_managed_node_group/modules/eks-managed-node-group/main.tf line 301, in resource "aws_eks_node_group" "this":
 301:       source_security_group_ids = try(remote_access.value.source_security_group_ids, []) 

Inappropriate value for attribute "source_security_group_ids": set of string
required.
```

Based on AWS API documentation [here](https://docs.aws.amazon.com/eks/latest/APIReference/API_RemoteAccessConfig.html) and eks_node_group documentation using Terraform AWS provider [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#remote_access-configuration-block), `remote_access.source_security_group_ids` should be a set of security group IDs. Although, `remote_access` defined in [variables.tf](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/eks-managed-node-group/variables.tf#L335) type map(string) which only allows managing source_security_group_ids as a string.

## Motivation and Context

Managing `remote_access` as type map(string) is probably fine, we can define multiple security group IDs with values in comma separated string as below. Change added as part of pull request simply splits the string to a set of string IDs that gets assigned as required to `eks_node_group` resource.

```
  remote_access = {
     ec2_ssh_key = "ec2-ssh-key"
     source_security_group_ids = "security-group-id-1,security-group-id-2"
  }
```

## Breaking Changes

No, this fixes an issue with setting source_security_group_ids as part of remote_access map.

## How Has This Been Tested?

Produces output for remote_access block as,

```
+ remote_access {
          + ec2_ssh_key               = "ec2-ssh-key"
          + source_security_group_ids = [
              + "security-group-id-1",
              + "security-group-id-2",
            ]
```

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -298,7 +298,7 @@ resource "aws_eks_node_group" "this" {
     for_each = length(var.remote_access) > 0 ? [var.remote_access] : []
     content {
       ec2_ssh_key               = try(remote_access.value.ec2_ssh_key, null)
-      source_security_group_ids = try(remote_access.value.source_security_group_ids, [])
+      source_security_group_ids = try(split(",", remote_access.value.source_security_group_ids), [])
     }
   }
 


bryantbiggs:
Ah shoot, we should change the variable type to `map(any)` or just `any`

cbirajdar:
@bryantbiggs yeah, I thought of changing type to `any` as well but was hoping to avoid it, in favor of use strongly typed variable instead. That's why thought of leveraging comma-separated values instead but it might not be an ideal way to handle it.

May be we use an object type to allow explicitly defining that, as set of strings? Default values would be what's already being substituted in main.tf for both attributes i.e. `null and []`.

```
variable "remote_access" {
  description = "Configuration block with remote access settings"
  type = object({ ec2_ssh_key = string, source_security_group_ids = list(string) })
  default = { ec2_ssh_key=null, source_security_group_ids=[] }
}
```

bryantbiggs:
I think in the future this sort of thing will be better handled by https://www.terraform.io/language/expressions/type-constraints#experimental-optional-object-type-attributes but while its still in experimental, I think the best course of action is to just change the type to be `map(any)` or `any`.

antonbabenko:
This issue has been resolved in version 18.0.6 :tada:

Issue #1753, opened 2022-01-07
Data source for current AWS partition breaks modules


watsonjm:
## Description

Please provide a clear and concise description of the issue you are encountering, your current setup, and what steps led up to the issue. If you can provide a reproduction, that will help tremendously.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions
Terraform v1.1.2
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v3.71.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/helm v2.2.0
+ provider registry.terraform.io/hashicorp/http v2.1.0
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/hashicorp/tls v3.1.0

 - Module: 18.0.4

## Reproduction

No Workspaces

I have cleared the local cache

steps to reproduce:
terraform init
terraform plan

### Code Snippet to Reproduce
```
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "18.0.4"

  cluster_name                    = local.eks_cluster_name
  cluster_version                 = "1.21"
  cluster_endpoint_private_access = true
  cluster_endpoint_public_access  = true
  vpc_id                          = module.vpc.0.vpc_id
  subnet_ids                      = local.private_subnets_ids
  cluster_enabled_log_types       = ["api", "audit", "authenticator", "controllerManager", "scheduler"]
  cluster_security_group_name     = "${local.name_tag}-eks-cluster"
  cluster_service_ipv4_cidr       = "10.100.0.0/16"
  cluster_addons = {
    coredns = {
      resolve_conflicts = "OVERWRITE"
    }
    kube-proxy = {}
    vpc-cni = {
      resolve_conflicts = "OVERWRITE"
    }
  }

  # cluster_encryption_config = [{
[...]

PR #1766, opened 2022-01-11
fix: Correct remote access variable for security groups and add example for additional IAM policies

bryantbiggs:
## Description
- fix vpc remote access variable for security groups to allow for list of security groups; closes #1762 
- update `eks-managed-node-group` to demonstrate shortcomings of additional IAM policies and security groups, providing an explanation and example for how best to utilize these due to https://github.com/hashicorp/terraform/issues/4149
	- note added to main README to alert and educate users on this issue

## Motivation and Context
- Ref https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1753#issuecomment-1010146823
- Closes #1762 
- documentation for users on how to use additional resources within the module

## Breaking Changes
- No

## How Has This Been Tested?
	- Validate with `eks-managed-node-group` example

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,39 @@ Terraform module which creates AWS EKS (Kubernetes) resources
 - Support for providing maps of node groups/Fargate profiles to the cluster module definition or use separate node group/Fargate profile sub-modules
 - Provisions to provide node group/Fargate profile "default" settings - useful for when creating multiple node groups/Fargate profiles where you want to set a common set of configurations once, and then individual control only select features
 
+### ℹ️ `Error: Invalid for_each argument ...`
+
+Users may encounter an error such as `Error: Invalid for_each argument - The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply ...`
+
+This error is due to an upstream issue with [Terraform core](https://github.com/hashicorp/terraform/issues/4149). There are two potential options you can take to help mitigate this issue:
+
+1. Create the dependent resources before the cluster  => `terraform apply --target <your policy or your security group>` and then `terraform apply` for the cluster (or other similar means to just ensure the referenced resources exist before creating the cluster)
+  - Note: this is the route users will have to take for adding additonal security groups to nodes since there isn't a separate "security group attachment" resource
+2. For addtional IAM policies, users can attach the policies outside of the cluster definition as demonstrated below
+
+```hcl
+resource "aws_iam_role_policy_attachment" "additional" {
+  for_each = module.eks.eks_managed_node_groups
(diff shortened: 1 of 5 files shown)


bryantbiggs:
I think I can improve this a bit for users - one sec

bryantbiggs:
I was thinking for a second of adding this policy attachment to the root of the module and handling this for users but that only handles one scenario - where users want to attach a policy to ALL roles created. It would get out of hand rather quickly if we take into consideration where users want a policy attached to specific roles or specific node groups, etc. and decided against it. for now, I think the note added is the best course of action for users at this time

README.md
```diff
@@ -18,6 +18,39 @@ Terraform module which creates AWS EKS (Kubernetes) resources
 - Support for providing maps of node groups/Fargate profiles to the cluster module definition or use separate node group/Fargate profile sub-modules
 - Provisions to provide node group/Fargate profile "default" settings - useful for when creating multiple node groups/Fargate profiles where you want to set a common set of configurations once, and then individual control only select features
 
+### ℹ️ `Error: Invalid for_each argument ...`
+
+Users may encounter an error such as `Error: Invalid for_each argument - The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply ...`
+
+This error is due to an upstream issue with [Terraform core](https://github.com/hashicorp/terraform/issues/4149). There are two potential options you can take to help mitigate this issue:
+
+1. Create the dependent resources before the cluster  => `terraform apply --target <your policy or your security group>` and then `terraform apply` for the cluster (or other similar means to just ensure the referenced resources exist before creating the cluster)
```
antonbabenko, line 27:
```suggestion
1. Create the dependent resources before the cluster  => `terraform apply -target <your policy or your security group>` and then `terraform apply` for the cluster (or other similar means to just ensure the referenced resources exist before creating the cluster)
```

Maybe it also works with `--target` but I am not sure.

examples/eks_managed_node_group/main.tf
```diff
@@ -269,6 +270,18 @@ module "eks" {
   tags = local.tags
 }
 
+# References to resources that do not exist yet when creating a cluster will cause a plan failure due to https://github.com/hashicorp/terraform/issues/4149
+# There are two options users can take
+# 1. Create the dependent resources before the cluster => `terraform apply --target <your policy or your security group> and then `terraform apply`
```
antonbabenko, line 275:
```suggestion
# 1. Create the dependent resources before the cluster => `terraform apply -target <your policy or your security group> and then `terraform apply`
```

antonbabenko (approved):

antonbabenko:
This PR is included in version 18.0.6 :tada:

PR #1768, opened 2022-01-12
fix: truncate iam_role_name to 38 characters

mcanevet:
IAM roles name can't be longer than 38 characters so we have to truncate it.


bryantbiggs:
I believe this is only true for `name_prefix`, correct? We tend to avoid this manipulation since it's not what users expect unless they dig into the source code and see it's being truncated

mcanevet:
Yes, indeed it's only true for `name_prefix`.
I'll do it on my side if you think that doing it in this module does not make sense.
Thanks for your quick feedback.

PR #1774, opened 2022-01-12
fix: Don't tag self managed node security group with kubernetes.io/cluster tag

imdevin567:
## Description
This removes the forced kubernetes.io/cluster tag in self managed node security groups in favor of the user supplying it themselves.

## Motivation and Context
We currently tag the created node security group with the `kubernetes.io/cluster/clustername = owned` tag. We are also tagging the self managed node security group with the same tag, which causes a conflict launching load balancers:

```
Error syncing load balancer: failed to ensure load balancer: Multiple tagged security groups found for instance i-aaaaaab0ab17adb5d; ensure only the k8s security group is tagged;
```

Since we don't add this tag to the EKS managed node security group, this will ensure consistency among node groups.

## Breaking Changes
No breaking change

## How Has This Been Tested?

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -456,10 +456,6 @@ resource "aws_security_group" "this" {
 
   tags = merge(
     var.tags,
-    {
-      "Name"                                      = local.security_group_name
-      "kubernetes.io/cluster/${var.cluster_name}" = "owned"
-    },
     var.security_group_tags
   )
 }


bryantbiggs:
I think the alternative here would be to not let the self managed node group create a security group, and instead supply your own security group

bryantbiggs:
@daroga0002 thoughts?

daroga0002:
@imdevin567  could you pass here problematic module configuration which is solved by this PR?

modules/self-managed-node-group/main.tf
```diff
@@ -459,10 +459,6 @@ resource "aws_security_group" "this" {
 
   tags = merge(
     var.tags,
-    {
-      "Name"                                      = local.security_group_name
```
bryantbiggs, line 463: please add back in the `"Name"` tag, then we should be good to go with this one

imdevin567: Should be all set now.

bryantbiggs (approved):
@antonbabenko good to go 👍🏽  - we might see an issue pop up since its removing a tag, but it is the right move to remove it and let user opt in to when and where they want to use that tag

antonbabenko:
This PR is included in version 18.2.7 :tada:

PR #1789, opened 2022-01-18
feat: Update userdata logic for consistency

stevehipwell:
## Description
This PR improves the consistency of the userdata module and it's usage by the node group modules. The following improvements have been added.

- Provide a consistent default template shebang and `set` directive
- Simplify user data module implementation
- Added `user_data_env` variable for consistency and to enable default template re-use
- Improve user data module templating documentation
- Support setting `enable_bootstrap_user_data` to `false` for self managed node groups
- Support setting `SERVICE_IPV4_CIDR` via the `cluster_service_ipv4_cidr` variable for self-managed node groups
- Add a framework for managed node groups using the default AMI to pass environment variables to the bootstrap script
- Support setting `SERVICE_IPV4_CIDR` via the `cluster_service_ipv4_cidr` variable for managed node groups using the default AMI
- Make managed node group examples providing additional bootstrap variables work
- Improve userdata examples to cover more supported use cases
- Improve documentation on managed node group edge cases

Once this has been released the `bootstrap_extra_args` should be removed and replaced with a Bottlerocket specific input as the new `user_data_env` can do everything this can do in a simpler way.

## Motivation and Context
The v18 release broke patterns that worked in v17 and provided an inconsistent "API" for the featured it supported, this PR adds back some of the basic functionality and fixes inconsistencies in the "API" without making any breaking changes.

## Breaking Changes
No

## How Has This Been Tested?


stevehipwell:
The code here has some changes I missed staging so isn't expected to work. And after sleeping on it I've got a couple of changes to make to the implementation.

ArchiFleKs:
Looks good, I'll try to test it

ArchiFleKs:
@stevehipwell sorry did not have time to test it yet but happy to see that you have

stevehipwell:
@antonbabenko @bryantbiggs I've tested this PR and it works as expected and allows me to use exactly the same bootstrapping inputs for EKS and self managed node groups. Could you please review the code?

bryantbiggs:
1. Theres a lot of whitespace noise that is incorrect (an indent inside the heredocs blocks will be an indent in the userdata which is not desired)
2. Self-managed node groups and EKS managed node groups do not operate the same. Therefore, I don't see the point to try to introduce some "magic" to make the user data the same. EKS managed node groups append their own logic in the user data - thats just how that service works (unless a user opts in to providing a custom AMI). 

I will be honest - I don't see us accepting this change. I don't know what problem it is trying to solve and I don't want to slip back into the "magic" logic that was in v17 and below

stevehipwell:
@bryantbiggs to answer your points.

1. The heredoc blocks have no additional whitespace as they're using [indented heredocs](https://www.terraform.io/language/expressions/strings#indented-heredocs)
2. It's not magic, it just allows you to do the common userdata tasks without having to fully understand how launch template merging works. Yes you do need to make some decisions based on managed or un-managed, but managed node groups just work with this code and can be made as complex as required which they can't with the current version due to limited template inputs.

I'm really confused why you wouldn't want this PR? I'm happy to address feedback but it contains missing functionality which is foundational and was removed in the v17 refactor.

bryantbiggs:
Please see the 2nd example at the heredocs link - we don't want a tab space at the start of each line of the user data. It was correct as it was before.

> I'm really confused why you wouldn't want this PR?

I just don't see the benefits of this added complexity - I would suspect if there was a problem that was worth solving here that we would see more issues raised for it. However, thats not the case. Once K8s 1.24 lands and CoreDNS is the default, we won't even show the work around (that you provided) in the examples and things will just be plain, simple, boring (from the module perspective, not K8s perspective)

bryantbiggs:
> Which change are you saying is incorrect? When using the <<-EOT marker adding an indent to all lines in the heredoc is a no-op and IMHO makes it easier to read. I could easily undo it but functionally it wouldn't make any difference.

Run this locally and you will see once you inspect the generated file:

```hcl
locals {
  heredoc = <<-EOT
  Aligned
    Not aligned - preceded by whitespace
  EOT
}

resource "local_file" "test" {
  content  = local.heredoc
  filename = "${path.module}/test.txt"
}
```

We want the text aligned with the trailing `EOT` to left align the text (current format). The indented `<<-EOT` (vs `<<EOT`) format just allows us to indent within HCL for readability, but we still want to align with the trailing EOT so the text is not rendered with a leading space

> I assume you mean containerd?

Ah yes, my bad - contianerd not CoreDNS

I'm sorry but I'm still going to have to disagree - I don't see the benefits of this change other than a preference about how data is passed to the user data. I'm afraid I don't have any other comments on this topic

stevehipwell:
@bryantbiggs

> Run this locally and you will see once you inspect the generated file:

Why? If I've done that it's a typo, all the heredoc indentations I did were intended to be to all lines so a no-op. I still can't see any that this happened to?

Run this locally:

```hcl
locals {
  heredoc = <<-EOT
    Aligned - no whitespace
      Not aligned - preceded by whitespace but only 2
  EOT
}

resource "local_file" "test" {
  content  = local.heredoc
  filename = "${path.module}/test.txt"
}
```

> I'm sorry but I'm still going to have to disagree - I don't see the benefits of this change other than a preference about how data is passed to the user data.

It's not **how** data is passed to the module, it's **the data** passed to the module; which is a significant difference. A change in how would be pretty inconsequential but a change in the data makes lots of currently impossible things possible.

>  I'm afraid I don't have any other comments on this topic

I thought this was a community module?

bryantbiggs:
again, thank you for the PR but unfortunately we won't be going forward with these changes

PR #1797, opened 2022-01-19
docs: Adding Information on reusing Cluster IAM Role

jseiser:
## Description
Documenting how to keep the existing cluster iam role post upgrade.

## Motivation and Context
This information allowed me to upgrad my existing cluster without recreating it.

## Breaking Changes
None

## How Has This Been Tested?
This is the process I used in my own upgrade

### DIFF
--- a/UPGRADE-18.0.md
+++ b/UPGRADE-18.0.md
@@ -584,3 +584,22 @@ resource "aws_iam_role_policy_attachment" "default" {
   policy_arn = aws_iam_policy.default.arn
 }
 ```
+### Reusing existing Cluster IAM Role
+
+If you allowed the module to create your Cluster IAM Role
+and want to reuse it moving forward you must
+migrate it into the state.
+
+#### Pre-Upgrade
+
+```hcl
+terraform state mv 'module.eks.aws_iam_role.cluster[0]' 'module.eks.aws_iam_role.this[0]'
+```
+
+```hcl
+  prefix_separator         = ""
+  create_iam_role          = true
+  iam_role_use_name_prefix = false
+  iam_role_name            = "EXISTING ROLE NAME"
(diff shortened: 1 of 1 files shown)


UPGRADE-18.0.md
```diff
@@ -584,3 +584,22 @@ resource "aws_iam_role_policy_attachment" "default" {
   policy_arn = aws_iam_policy.default.arn
 }
 ```
+### Reusing existing Cluster IAM Role
+
+If you allowed the module to create your Cluster IAM Role
+and want to reuse it moving forward you must
+migrate it into the state.
```
bryantbiggs, line 591:
```suggestion
When upgrading from v17.x to v18.x using the previously generated cluster IAM role, you will need to make a few changes to avoid replacing/re-creating the role which will cause the cluster to be replaced.

First, you will need to perform a Terraform state mv to "move" the previous resource name into the new resource name. See state move command below:

```bash
terraform state mv 'module.eks.aws_iam_role.cluster[0]' 'module.eks.aws_iam_role.this[0]'
```


Next, you will need to configure a few settings that will align your role with the v17.x version and avoid replacing/re-creating. See the associated settings below that should be set:

```hcl
  iam_role_name            = <name of current role created by v17.x module>
	iam_role_use_name_prefix = false
```
```

We don't need to tell users to specify all those settings because they do not make sense. If you set the prefix separator to an empty string, but then follow it up with `iam_role_use_name_prefix = false`, it doesn't make sense because we've ignored its use.

Also, we don't need to specify the role ARN because that is only for externally managed roles.

jseiser:
Then what would you consider the correct information to migrate an existing cluster without re-creating the IAM role?

I created a pull matching the information that was presented in #1744

PR #1804, opened 2022-01-21
fix: Attributes in timeouts are erroneously reversed

qingvincentyin:
## Description
The 2 lines in my proposed change were erroneously reversed in your original code.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -43,8 +43,8 @@ resource "aws_eks_cluster" "this" {
 
   timeouts {
     create = lookup(var.cluster_timeouts, "create", null)
-    delete = lookup(var.cluster_timeouts, "delete", null)
     update = lookup(var.cluster_timeouts, "update", null)
+    delete = lookup(var.cluster_timeouts, "delete", null)
   }
 
   depends_on = [


bryantbiggs:
is this change necessary?

qingvincentyin:
Are you asking me? If so, why not? The original code obviously had the 2 attributes transposed.

bryantbiggs:
ooooh, I see it now. Sorry, the information provided in the PR was not clear

main.tf
```diff
@@ -43,8 +43,8 @@ resource "aws_eks_cluster" "this" {
 
   timeouts {
     create = lookup(var.cluster_timeouts, "create", null)
-    delete = lookup(var.cluster_timeouts, "update", null)
```
bryantbiggs, line 46: minor nit-pick - could we do the order of `create`, `update`, `delete` (CRUD)

bryantbiggs (approved):
@antonbabenko should be go with or without the order change

thanks for the catch and fix @qingvincentyin

antonbabenko:
I agree that it should be in order like CRUD. Merging now.

antonbabenko:
This PR is included in version 18.2.2 :tada:

PR #1805, opened 2022-01-21
chore: Add example usage of extending cluster and node security group access rules

### LINKED ISSUE #1790: communication between pods is too narrow.
## Description

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/9f85dc8cf5028ed2bab49f495f58e68aa870e7d4/node_groups.tf#L86

Communication with clusters in node_security_group_rules is fine, but communication between pods is too narrow.
For example, if you deploy an ArgoCD cluster, it will not work because the communication between pods does not work.
Since the communication between pods is basically limited by k8s, it is better to keep the default limitation as follows.
This is also how `eksctl` works.

```
{
  ingress_self_any = {
    description = "Node to node All"
    protocol    = "all"
    from_port   = 0
    to_port     = 65535
    type        = "ingress"
    self        = true
  }
  egress_self_any = {
    description = "Node to node All"
    protocol    = "all"
    from_port   = 0
    to_port     = 65535
    type        = "egress"
    self        = true
  }
}
```

bryantbiggs:
## Description
- add example usage of extending cluster and node security group access rules

## Motivation and Context
- Closes #1790 
- This should provide more example usage to users to demonstrate how to extend the security group access for the baseline that is provided

## Breaking Changes
- No

## How Has This Been Tested?
	  - Update examples and tested on `complete` example

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,6 +578,44 @@ Module provided default templates can be found under the [templates directory](h
   - Each node group (EKS Managed Node Group and Self Managed Node Group) by default creates its own security group. By default, this security group does not contain any additional security group rules. It is merely an "empty container" that offers users the ability to opt into any addition inbound our outbound rules as necessary
   - Users also have the option to supply their own, and/or additional, externally created security group(s) to the node group as well via the `vpc_security_group_ids` variable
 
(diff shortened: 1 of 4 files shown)


antonbabenko:
This PR is included in version 18.2.3 :tada:

PR #1818, opened 2022-01-29
fix: Add missing `launch_template_use_name_prefix` parameter to the root module

efaif:
## Description
* Add missing `launch_template_use_name_prefix` parameter to the root module

## Motivation and Context
* Currently `self-managed-node-group`'s `launch_template_use_name_prefix` variable is not passed to the module, this PR is intended to solve this

## How Has This Been Tested?
* Tested by using the `self-managed-node-group` sub-module with `launch_template_use_name_prefix` set to `True`/`False` and verifying that `aws_launch_template` resource's `name_prefix`/`name` arguments are being set accordingly and the launch template is created as expected

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -385,10 +385,11 @@ module "self_managed_node_group" {
   user_data_template_path  = try(each.value.user_data_template_path, var.self_managed_node_group_defaults.user_data_template_path, "")
 
   # Launch Template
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
thank you for catching this @efaif !

good to go @antonbabenko 👍🏽

antonbabenko:
This PR is included in version 18.2.4 :tada:

Issue #1819, opened 2022-01-30
Wrong rolearn in aws_auth_configmap_yaml


Anorlondo448:
## Description

"data.maproles[].rolearn" set in aws_auth_configmap_yaml is not “Pod Execution Role ARN”. It is “Fargate Profile ARN”.
The correct policy is not attached to the user mapped in the aws-auth ConfigMap.

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/16f46db94b7158fd762d9133119206aaa7cf6d63/templates/aws_auth_cm.tpl#L30-L37

## Versions
 - Terraform: v1.1.4
 - Provider(s):
```
+ provider registry.terraform.io/hashicorp/aws v3.74.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v1.13.4
+ provider registry.terraform.io/hashicorp/tls v3.1.0
```

## Reproduction
- use `terraform-aws-modules/eks/aws`
- Use only fargate_profile, not nodegroups.
- output `module.eks.aws_auth_configmap_yaml`

### Code Snippet to Reproduce
- output value
```
Changes to Outputs:
  + aws_auth_configmap_yaml = <<-EOT
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: aws-auth
          namespace: kube-system
        data:
          mapRoles: |
            - rolearn: arn:aws:eks:<region>:<AWS account id>:fargateprofile/<cluster name>/<fargate profile name>/xxxxxxxx
              username: system:node:{{SessionName}}
              groups:
                - system:bootstrappers
                - system:nodes
                - system:node-proxier
    EOT
```

## Expected behavior
"data.maproles[].rolearn" should be set to “Pod Execution Role ARN”.

## Actual behavior
set to "Fargate Profile ARN"

## Additional context
- The output of aws_auth_configmap_yaml does not have a Pod Execution Role ARN
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/be86c0b898c34943e898e2ecd4994bb7904663ff/outputs.tf#L164-L172

- templates/aws_auth_cm.tpl should also use the "Pod Execution Role ARN", not "Fargate Profile ARN"
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/16f46db94b7158fd762d9133119206aaa7cf6d63/templates/aws_auth_cm.tpl#L30-L32

[...]

PR #1820, opened 2022-01-30
fix: Wrong rolearn in aws_auth_configmap_yaml 

Anorlondo448:
## Description
Change the rolearn set in aws_auth_configmap_yaml to the correct one.
Currently, the “Fargate Profile ARN” is set instead of the “Pod Execution Role ARN”, so the correct IAM policy cannot be retrieved.

## Motivation and Context
- #1819

## Breaking Changes
Yes.
Because currently, “Fargate Profile ARN” is set in rolearn instead of “Pod Execution Role ARN”, so the correct IAM policy is It cannot be retrieved.

## How Has This Been Tested?
Output aws_auth_configmap_yaml and confirmed that “Pod Execution Role ARN” is set in rolearn.


bryantbiggs (approved):
thanks for catching this @Anorlondo448 - verified that this change is the correct output this morning. please run `pre-commit run -a` on your PR so the docs are updated, etc.

should be good to go with the above @antonbabenko 👍🏽

antonbabenko:
This PR is included in version 18.2.6 :tada:

PR #1824, opened 2022-02-01
fix: Correct issue where custom launch template is not used when EKS managed node group is used externally

### LINKED ISSUE #1816: Cannot use custom launch template when launch_template_name is not specified
## Description
Hi team, I found a little confusing behavior about `eks-managed-node-group` module.

When I don't specify `launch_template_name` and leave it blank, the launch template will not be used by the corresponding managed node group. I've already found why it's happening (details are below), so could you judge if it's is an intended behavior? 

## Versions

 - Terraform: v1.1.4

 - Provider(s): registry.terraform.io/hashicorp/aws v3.73.0

 - Module: v18.2.3

## Reproduction

Deploy a VPC, an EKS cluster, and a managed node group. To reproduce this, we must explicitly use `eks-managed-node-group` module, not using `eks_managed_node_groups` input of a eks cluster module.

### Code Snippet to Reproduce

```tf

module "vpc" {
  source = "terraform-aws-modules/vpc/aws"

  name = var.cluster_name
  cidr = "10.0.0.0/16"

  azs             = data.aws_availability_zones.available.names
  private_subnets = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24"]
  public_subnets  = ["10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24"]

  enable_nat_gateway     = true
  single_nat_gateway     = true
  one_nat_gateway_per_az = false
}

module "cluster" {
  version = "~> 18.0"

  cluster_version = "1.21"
  cluster_name    = var.cluster_name
  vpc_id          = module.vpc.vpc_id
  subnet_ids      = module.vpc.private_subnets
}

module "managed_node_group" {
  source = "terraform-aws-modules/eks/aws//modules/eks-managed-node-group"

  name                 = "group"
  cluster_name         = var.cluster_name
  # this line is practically required
  # launch_template_name = "template_name"

  subnet_ids     = var.vpc.public_subnets
  vpc_id         = var.vpc.vpc_id
  vpc_security_group_ids = [var.node_security_group_id]
}
```

## Expected behavior
the resulting ASG uses the custom launch template.

## Actual behavior
the resulting ASG uses the default launch template.

## Additional context
The behavior is due to the code below. When `launch_template_name` is undefined, it won'd register the launch custom template name to the managed node group.

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/16f46db94b7158fd762d9133119206aaa7cf6d63/modules/eks-managed-node-group/main.tf#L32-L33
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/16f46db94b7158fd762d9133119206aaa7cf6d63/modules/eks-managed-node-group/main.tf#L292-L298

We might need another condition to define `use_custom_launch_template` variable.

BTW if we use `eks_managed_node_groups` input to deploy managed node groups, the problem won't happen because `launch_template_name` is always defined internally here.

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/16f46db94b7158fd762d9133119206aaa7cf6d63/node_groups.tf#L272

Thanks!

bryantbiggs:
## Description
- correct issue where custom launch template is not used when EKS managed node group is used externally
- update documentation and provide example of utilizing `containerd` runtime on EKS managed node group
- update static AMI IDs in examples to use data source; ensures the current AMI is always pulled when testing
- minor formatting for correct use of EOT and left aligning text

## Motivation and Context
- Closes #1816
- There have been a number of questions regarding the EKS managed node group user data and configuring the bootstrap process (specifically around enabling the containerd runtime) and hopefully this helps clear things up for users

## Breaking Changes
- No

## How Has This Been Tested?
	  - Using the `eks-managed-node-group` example
	  	- `eks-manged-node-group` example was deployed as its defined today
	  	- Added in sub-module snippet provided in issue here https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1816
	  	- 2nd apply created new node group using external sub-module - confirmed that the node group was not using the custom launch template 
	  	- Found issue and made correction on line 33 (see below) - re-ran plan and apply - only the external sub-module node group was affected and it was an update to use the custom launch template (desired affect)

### DIFF
--- a/examples/self_managed_node_group/main.tf
+++ b/examples/self_managed_node_group/main.tf
@@ -375,23 +375,23 @@ resource "aws_kms_key" "eks" {
 }
 
 data "aws_ami" "eks_default" {
+  most_recent = true
+  owners      = ["amazon"]
+
   filter {
     name   = "name"
     values = ["amazon-eks-node-${local.cluster_version}-v*"]
   }
-
-  most_recent = true
-  owners      = ["amazon"]
 }
 
 data "aws_ami" "eks_default_bottlerocket" {
+  most_recent = true
+  owners      = ["amazon"]
+
   filter {
     name   = "name"
     values = ["bottlerocket-aws-k8s-${local.cluster_version}-x86_64-*"]
   }
-
-  most_recent = true
-  owners      = ["amazon"]
 }
 
 resource "tls_private_key" "this" {
--- a/examples/self_managed_node_group/main.tf
+++ b/examples/self_managed_node_group/main.tf
@@ -163,12 +163,12 @@ module "eks" {
       bootstrap_extra_args = "--kubelet-extra-args '--max-pods=110'"
 
       pre_bootstrap_user_data = <<-EOT
-        export CONTAINER_RUNTIME="containerd"
-        export USE_MAX_PODS=false
+      export CONTAINER_RUNTIME="containerd"
+      export USE_MAX_PODS=false
       EOT
 
       post_bootstrap_user_data = <<-EOT
-        echo "you are free little kubelet!"
+      echo "you are free little kubelet!"
       EOT
 
       disk_size     = 256
--- a/examples/self_managed_node_group/main.tf
+++ b/examples/self_managed_node_group/main.tf
@@ -94,7 +94,7 @@ module "eks" {
       name = "bottlerocket-self-mng"
 
       platform      = "bottlerocket"
-      ami_id        = data.aws_ami.bottlerocket_ami.id
+      ami_id        = data.aws_ami.eks_default_bottlerocket.id
       instance_type = "m5.large"
       desired_size  = 2
       key_name      = aws_key_pair.this.key_name
@@ -159,7 +159,7 @@ module "eks" {
       max_size     = 7
       desired_size = 1
 
-      ami_id               = "ami-0caf35bc73450c396"
+      ami_id               = data.aws_ami.eks_default.id
       bootstrap_extra_args = "--kubelet-extra-args '--max-pods=110'"
 
       pre_bootstrap_user_data = <<-EOT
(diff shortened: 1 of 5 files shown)


modules/eks-managed-node-group/main.tf
```diff
@@ -30,7 +30,7 @@ module "user_data" {
 ################################################################################
 
 locals {
-  use_custom_launch_template = var.launch_template_name != ""
+  use_custom_launch_template = var.create_launch_template || var.launch_template_name != ""
```
bryantbiggs, line 33: this is the fix for ensuring the custom launch template is used when the sub-module is used externally

there are 4 scenarios here which is why it can be a bit weird to understand:
1. `var.create_launch_template = false && var.launch_template_name == ""` => EKS MNG will use its own default LT
2. `var.create_launch_template = false && var.launch_template_name == "something"` => User provided custom LT will be used
3. `var.create_launch_template = true && var.launch_template_name == ""` => Custom LT will be used, module will provide a default name
4. `var.create_launch_template = true && var.launch_template_name == "something"` => Custom LT will be used, LT name is provided by user

antonbabenko: Maybe put your comment right above this line as a description for this "light magic"? It will be helpful to know the reason in some time.

examples/self_managed_node_group/main.tf
```diff
@@ -374,14 +374,24 @@ resource "aws_kms_key" "eks" {
   tags = local.tags
 }
 
-data "aws_ami" "bottlerocket_ami" {
+data "aws_ami" "eks_default" {
+  filter {
```
antonbabenko, line 378: Let's move `filter` block after `owners` as it was.

antonbabenko (commented):
Minor comment. LGTM otherwise.

examples/self_managed_node_group/main.tf
bryantbiggs: fixed in d147a691f4d5f7a0bba9080976d46a9b61e8a23c

antonbabenko:
This PR is included in version 18.2.5 :tada:

Issue #1825, opened 2022-02-01
Add optional IRSA provisioning for most common EKS components ? 


philicious:
This module helps and eases EKS provisioning tremendously, especially for people less profound with EKS and/or Kubernetes.
However there are still more steps many users likely need to take to have a truely basic EKS setup:

I have created and managed over 10 EKS clusters productively with this module and easily 20 more for testing and engineering. Which also lead to adding LT support a yr ago ( https://github.com/terraform-aws-modules/terraform-aws-eks/pull/997 ), and the example LT by now made it into v18.x. Which is a huge leap forward in general.

Now, besides LTs, I also always need IRSA for cluster-autoscaler, external-dns, aws-loadbalancer-control and often also aws-ebs-csi-driver. So I basically have an extra .tf file for each in my wrapper module, utilizing https://registry.terraform.io/modules/terraform-aws-modules/iam/aws/latest/submodules/iam-assumable-role-with-oidc and a fitting policy.

Similarly, but more rare case, I also have a policy for a KMS key, so the ASGs are able to mount the encrypted disks

(Some of) these might be a good addition to this upstream module, disabled by default. It would ease provisioning of EKS clusters even more.

What do you guys think? I would volunteer preparing PRs and such

philicious:
cc @bryantbiggs @ArchiFleKs @barryib @stevehipwell (just highlighting some of you I know have contributed in discussions and PRs before )

ArchiFleKs:
I'd be happy to see this implemented, first for VPC-CNI addon which by default runs with node IAM roles I think. 

For the other "standard" EKS addons like ebs-csi-drivers (which will be in official addons soon I think), AWS load balancer controller etc why not. But I'm not sure it is the goal of this module to deploy this addons directly ? It would just be the IRSA roles then ?

I personally maintains this https://github.com/particuleio/terraform-kubernetes-addons/tree/main/modules/aws which I installed on top of all my EKS clusters that takes care of all the "addons" that make use of IRSA.

philicious:
@ArchiFleKs I am solely speaking of the IRSA part, as those require the EKS OIDC issuer URL and are thereby quite close to this module, logically.

the way of provisioning the components themselves, I would consider a users choice as there are multiple ways that can make sense depending on scenario

philicious:
[...]

PR #1827, opened 2022-02-01
feat: Add new sub-module for creating IAM roles for service accounts

bryantbiggs:
## Description
- Add new sub-module for creating IAM roles for service accounts

One one hand I can see this making some sense, but here are my current reservations:
- We already have this -> https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-assumable-role-with-oidc
- Which extended to then also add this -> https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-eks-role

Why do I like these ☝🏽? 
- They only have one provider which is the AWS provider. This change sort of sneakily adds in the Kubernetes provider which if you don't set the proper configs it just errors out with an obscure `dial tcp 127.0.0.1:80: connect: connection refused`

Why am I not (yet) drawn to this 👇🏽 change?
- What do we really gain? Service account and namespace creation? I do not like managing Kubernetes manifests in Terraform - I think they should be handled by whatever tool you use to deploy your applications. For IRSA, all we need to do is provide an IAM role really and if you see above, we have two module that already do that

Why this could be useful:
- Folks come here for EKS guidance and maybe they don't see the IRSA roles. Maybe moving it to a first-class supported sub-module here increases that visibility for them

## Motivation and Context

- #1825 

## How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+### [18.2.6](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/v18.2.5...v18.2.6) (2022-02-01)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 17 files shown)


bryantbiggs:
direction is to enable opt-in IAM policies under the https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-eks-role module

PR #1835, opened 2022-02-03
feat: Add launch_template_tags variable for additional launch template tags

kahirokunn:
## Description
add "launch_template_ignore_changes" option

## Motivation and Context
If I add a tag to `var.tags`, it will be added to all of them, but there was a situation where I wanted to add an arbitrary tag only to the specifications_tags of launch_template_tag, and I want to make it supported.


modules/eks-managed-node-group/variables.tf
```diff
@@ -408,6 +408,12 @@ variable "security_group_tags" {
   default     = {}
 }
 
+variable "launch_template_tag_specifications_tags" {
```
bryantbiggs, line 411:
```suggestion
variable "launch_template_tags" {
```

bryantbiggs: you'll need to update in both places and fix the static checks (i.e. - `pre-commit run -a`)

bryantbiggs:
I almost forgot - we'll need to update self-managed node groups module as well, and then propagate those changes  up to the root `node_groups.tf` as well - and the title should be updated to reflect - "feat: Add `launch_template_tags` variable for additional launch template tags"

kahirokunn:
@bryantbiggs
I think I did everything you told me to do.
But `pre-commit run -a` command does not complete no matter how long I wait 😢

bryantbiggs:
do you have pre-commit installed? in this repo it can take quite some time - you can get around this with (executed  from project root):
- `terraform fmt -recursive .`
- `pre-commit run terraform_docs -a`

kahirokunn:
@bryantbiggs 
It was just like you told me.
I think I've accomplished all the work I needed to!
Thx!

node_groups.tf
```diff
@@ -296,6 +296,7 @@ module "eks_managed_node_group" {
   enable_monitoring                  = try(each.value.enable_monitoring, var.eks_managed_node_group_defaults.enable_monitoring, true)
   network_interfaces                 = try(each.value.network_interfaces, var.eks_managed_node_group_defaults.network_interfaces, [])
   placement                          = try(each.value.placement, var.eks_managed_node_group_defaults.placement, null)
+  launch_template_tags               = try(each.value.launch_template_tags, var.eks_managed_node_group_defaults.launch_template_tags, null)
```
bryantbiggs, line 299: woops, one more - the very last default should be `{}` like the other tags are set here, and we can move this up to line 276

node_groups.tf
```diff
@@ -420,6 +421,7 @@ module "self_managed_node_group" {
   enable_monitoring                  = try(each.value.enable_monitoring, var.self_managed_node_group_defaults.enable_monitoring, true)
   network_interfaces                 = try(each.value.network_interfaces, var.self_managed_node_group_defaults.network_interfaces, [])
   placement                          = try(each.value.placement, var.self_managed_node_group_defaults.placement, null)
+  launch_template_tags               = try(each.value.launch_template_tags, var.self_managed_node_group_defaults.launch_template_tags, null)
```
bryantbiggs, line 424: likewise here, default of `{}` and move up to line 394

node_groups.tf
kahirokunn: done

node_groups.tf
kahirokunn: done

kahirokunn:
@bryantbiggs how about now?

node_groups.tf
```diff
@@ -273,6 +273,7 @@ module "eks_managed_node_group" {
   launch_template_use_name_prefix = try(each.value.launch_template_use_name_prefix, var.eks_managed_node_group_defaults.launch_template_use_name_prefix, true)
   launch_template_version         = try(each.value.launch_template_version, var.eks_managed_node_group_defaults.launch_template_version, null)
   launch_template_description     = try(each.value.launch_template_description, var.eks_managed_node_group_defaults.launch_template_description, "Custom launch template for ${try(each.value.name, each.key)} EKS managed node group")
+  launch_template_tags            = try(each.value.launch_template_tags, var.eks_managed_node_group_defaults.launch_template_tags, null)
```
bryantbiggs, line 276: Change the defaults please (on both)
```suggestion
  launch_template_tags            = try(each.value.launch_template_tags, var.eks_managed_node_group_defaults.launch_template_tags, {})
```

node_groups.tf
```diff
@@ -390,6 +391,7 @@ module "self_managed_node_group" {
   launch_template_use_name_prefix = try(each.value.launch_template_use_name_prefix, var.self_managed_node_group_defaults.launch_template_use_name_prefix, true)
   launch_template_version         = try(each.value.launch_template_version, var.self_managed_node_group_defaults.launch_template_version, null)
   launch_template_description     = try(each.value.launch_template_description, var.self_managed_node_group_defaults.launch_template_description, "Custom launch template for ${try(each.value.name, each.key)} self managed node group")
+  launch_template_tags            = try(each.value.launch_template_tags, var.self_managed_node_group_defaults.launch_template_tags, null)
```
bryantbiggs, line 394:
```suggestion
  launch_template_tags            = try(each.value.launch_template_tags, var.self_managed_node_group_defaults.launch_template_tags, {})
```

bryantbiggs (changes requested):

kahirokunn:
The correction was omitted.
Sorry about that.

bryantbiggs (approved):
thanks for your efforts @kahirokunn 🎉 

good to go 👍🏽 @antonbabenko

kahirokunn:
Thanks for all the advice!
I've learned so much!

antonbabenko:
This PR is included in version 18.3.0 :tada:

PR #1836, opened 2022-02-03
feat: Add "launch_template_ignore_changes" option

kahirokunn:
## Description
add "launch_template_ignore_changes" option

## Motivation and Context
default_tags always shows an update

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -238,6 +238,7 @@ resource "aws_launch_template" "this" {
 
   lifecycle {
     create_before_destroy = true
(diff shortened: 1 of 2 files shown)


modules/eks-managed-node-group/main.tf
```diff
@@ -238,6 +238,7 @@ resource "aws_launch_template" "this" {
 
   lifecycle {
     create_before_destroy = true
+    ignore_changes = var.launch_template_ignore_changes
```
bryantbiggs, line 241: lifecycle blocks do not accept variables so this won't work

kahirokunn: oh my god 😢

PR #1840, opened 2022-02-03
fix: Add node group dependency for EKS addons resource creation

### LINKED ISSUE #1801: Add-ons management Race condition with Node Groups
## Description
During module creation with Add-on "coredns" and a managed node group, there is the inevitable possibility of a race condition that I have hit a few times. Based on the random order of returned resources of spinning up the new the coredns add-on can potentially be attempted to be created before a manage node group which inherently creates a "wait for creation condition" which will fail out after 20 min. On a subsequent plan/apply given the ordering of resource creation has shifted, the node group will be created and then finally core-dns will destroy and recreate successfully. 

## Versions

 - Terraform: latest

 - Provider(s): latest

 - Module: latest

## Reproduction

 N/A
 N/A
 N/A

## Expected behavior
CoreDNS addon should probably have a depends-on block for the node profile. 

## Additional context

Happy to submit a PR, I just have not had an opportunity to get familiar with the actual module code at the moment.  I will happily attempt to get familiar with the module to PR it this weekend. However, wanted to open an issue so if someone more familiar with it knows it could easily be a 5 minute fix.

bryantbiggs:
## Description
- Add node group dependency for EKS addons resource creation
- Added note regarding future addons that may need to be provisioned ahead of node groups; we can add a separate resource for those addons or re-visit if the lifecycle of addons has been fixed upstream. However, for now I think we can live with this change without much issue

## Motivation and Context
- Closes #1801

## Breaking Changes
- No

## How Has This Been Tested?
	- Tested with `eks-managed-node-group` example

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -234,6 +234,16 @@ resource "aws_eks_addon" "this" {
     ]
   }
 
+  # Note: if an addon needs to be provisioned ahead of a node group users will
+  # need to create the addon outside of this module until a 2nd addon resource is added
+  # to the module (here) that is not dependent on node groups
+  # Or if addon management improves, this dependency can be removed https://github.com/aws/containers-roadmap/issues/1389
+  depends_on = [
+    module.fargate_profile,
+    module.eks_managed_node_group,
+    module.self_managed_node_group,
+  ]
+
   tags = var.tags
 }
 


bryantbiggs:
to follow this up, I think the next "breaking" change will be to move addons out to a sub-module (not yet, but getting very close):
- these dependencies are resolved once addons are externally managed/created from the cluster 
- this sub-module will contain other features related to addons (see #1841)

bryantbiggs:
after looking into this further, I do think this is a valid change to consider:
1. From the [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html)
> Amazon EKS add-ons run on the nodes that you provision or configure for your cluster. Node types include Amazon EC2 instances and Fargate.

This means that the hard dependency is valid and my prior thinking conflicts could arise from control plane addons was incorrect

2. https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1841 will be supported with the policy additions in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/184

bryantbiggs:
@antonbabenko this is ready for review if you get some time

antonbabenko:
This PR is included in version 18.4.1 :tada:

drodrigues3:
I`m  facing an issue that I belive is related to this feature, I did create a new cluster and after 20 minutes all the node groups was failed to start. In the description of the node I did found this error message:

Ready            False   Fri, 11 Feb 2022 15:46:44 -0300   Fri, 11 Feb 2022 15:46:34 -0300   KubeletNotReady              container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized

To solve this I just enabled manually the ADDON AWS CNI and KUBE-PROXY, so all the node groups  was started.

May these two addons should be created at same time that the node group ?

FernandoMiguel:
my understanding is that at least AWS VPC CNI needs to be available before ec2 are spun up, or they wont join the cluster
coredns needs to be added only after the node groups are healthy and joined the cluster.

bryantbiggs:
@FernandoMiguel  per the docs, addons come first. However, i do agree with your sentiment and have submitted a proposal for one possible resolution (see linked issues - they all related to the CNI quite a bit) https://github.com/aws/containers-roadmap/issues/1666

FernandoMiguel:
On our module that depends on this, I've split up what goes to the module (at creation time of the workers) and another set of vars and addons that depend on the workers be already created....it's smelly code but works most of the times.
I still have issues with fargate profiles where coredns gets unhealthy a bunch of times... I have to manually delete it and reapply...
Need to spend time figuring out what's preventing it from get healthy at creation time.

bryantbiggs:
> On our module that depends on this, I've split up what goes to the module (at creation time of the workers) and another set of vars and addons that depend on the workers be already created....it's smelly code but works most of the times. I still have issues with fargate profiles where coredns gets unhealthy a bunch of times... I have to manually delete it and reapply... Need to spend time figuring out what's preventing it from get healthy at creation time.

I am assuming you saw this https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-coredns - but just FYI in case you didn't

FernandoMiguel:
> > On our module that depends on this, I've split up what goes to the module (at creation time of the workers) and another set of vars and addons that depend on the workers be already created....it's smelly code but works most of the times. I still have issues with fargate profiles where coredns gets unhealthy a bunch of times... I have to manually delete it and reapply... Need to spend time figuring out what's preventing it from get healthy at creation time.
> 
> I am assuming you saw this https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-coredns - but just FYI in case you didn't 

I did.. And something tells me the doc is outdated.
Recently seems you no longer need to perform the EDIT..
At least I've been able to deploy clusters just fine... But sometimes it fails... Not every time.
This started working with the newer version of the coredns addon.
If you don't specify the version, it will pick and older one, and I believe that one fails.
I've pinned ours to the latest and is mostly OK.

Issue #1844, opened 2022-02-04
"Instance market options" with the launch template


refucktor:
## Description

EKS Managed Group should support the "Instance market options" with the launch template.

## Versions

 - Terraform:
 
 ```shell
 Terraform v1.1.4                                                                                   
on linux_amd64
 ```

 - Provider(s):
 
 ```shell
 + provider registry.terraform.io/hashicorp/aws v3.73.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/helm v2.4.1
+ provider registry.terraform.io/hashicorp/kubernetes v2.7.1
+ provider registry.terraform.io/hashicorp/tls v3.1.0
 ```

 - Module:
 ```
 terraform-aws-modules/eks/aws: 18.2.1 
 ```

## Reproduction

1. Configure this module using the examples
2. Add an "instance_market_options" under the "Managed Node Groups"
3. Plan and apply

### Code Snippet to Reproduce

For simplicity I referenced some values by using variables, local or other modules, because the root of the error is from the Launch Template creation

```hcl
module "dev_eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "18.2.1"

  # cluster configuration
  cluster_name              = local.cluster_name
  cluster_version           = local.cluster_version
  cluster_enabled_log_types = ["audit", "api", "authenticator"]
  cluster_addons = {
    coredns = {
      resolve_conflicts = "OVERWRITE"
    }
    vpc-cni = {
      resolve_conflicts = "OVERWRITE"
    }
  }
  cluster_encryption_config = [{
    provider_key_arn = var.aws_kms_key_eks_arn
    resources        = ["secrets"]
  }]

  # iam config
  iam_role_name            = local.cluster_iam_role_name
  iam_role_path            = local.cluster_iam_path
  iam_role_use_name_prefix = false
  enable_irsa              = true

  # network
  vpc_id     = var.aws_vpc_id
  subnet_ids = var.vpc_private_subnets_ids

  # security groups
  cluster_security_group_name = format("%s-sg", local.cluster_name)
  node_security_group_name    = format("%s-node-sg", local.cluster_name)

  # node groups
  eks_managed_node_group_defaults = {
    disk_size = 50
    platform  = "bottlerocket"
    ami_id    = data.aws_ami.bottlerocket_ami.id
  }
 eks_managed_node_groups = {
    spot_worker = {
[...]

PR #1847, opened 2022-02-04
fix: The `block_duration_minutes` attribute under launch template `spot_options` is not a required

bryantbiggs:
## Description
- `block_duration_minutes` under launch template `spot_options` is not a required attribute
- Fix `compelte` example, external EKS managed node group module which was missing the security group IDs and failing to join the cluster

## Motivation and Context
- This was a bug found when investigating #1844

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -158,7 +158,7 @@ resource "aws_launch_template" "this" {
       dynamic "spot_options" {
         for_each = lookup(instance_market_options.value, "spot_options", null) != null ? [instance_market_options.value.spot_options] : []
         content {
-          block_duration_minutes         = spot_options.value.block_duration_minutes
+          block_duration_minutes         = lookup(spot_options.value, block_duration_minutes, null)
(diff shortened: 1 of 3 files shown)


antonbabenko:
This PR is included in version 18.3.1 :tada:

modules/self-managed-node-group/main.tf
```diff
@@ -158,7 +158,7 @@ resource "aws_launch_template" "this" {
       dynamic "spot_options" {
         for_each = lookup(instance_market_options.value, "spot_options", null) != null ? [instance_market_options.value.spot_options] : []
         content {
-          block_duration_minutes         = spot_options.value.block_duration_minutes
+          block_duration_minutes         = lookup(spot_options.value, block_duration_minutes, null)
```
agconti, line 161: block_duration_minutes should be quoted right? 

```
block_duration_minutes = lookup(spot_options.value, "block_duration_minutes", null)
```

agconti: Ok, opened a pr: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1881

PR #1848, opened 2022-02-06
feat: Allow conditional creation of node groups to be set within node group definitions

PhilippMT:
## Description
- Allow conditional creation of self_managed_node_group by passing the variable "create" to the module
- If "create" is not present the default value "true" is used
- If "create" is set to "false" the module is not created

## Motivation and Context
We are using the same manifest for different stages. Some resources are created conditionally based on variables set for a specific stage. We defined multiple self managed node groups in the map self_managed_node_group, but not all of them are relevant for all stages. We would like to be able to create some of the entries conditionally on the "create" variable defined by the self_managed_node_group submodule.

## Breaking Changes
- No

## How Has This Been Tested?
I tested three different scenarios
1.  create = true -> The self_managed_node_group was created with all its resources
2. create = false -> The self_managed_node_group was ignored
3. create missing -> The self_managed_node_group was create with all its resources

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -331,6 +331,8 @@ module "self_managed_node_group" {
 
   for_each = { for k, v in var.self_managed_node_groups : k => v if var.create }
 
+  create = try(each.value.create, true)
+
   cluster_name      = aws_eks_cluster.this[0].name
   cluster_ip_family = var.cluster_ip_family
 
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -194,6 +194,8 @@ module "fargate_profile" {
 
   for_each = { for k, v in var.fargate_profiles : k => v if var.create }
 
+  create = try(each.value.create, true)
+
   # Fargate Profile
   cluster_name      = aws_eks_cluster.this[0].name
   cluster_ip_family = var.cluster_ip_family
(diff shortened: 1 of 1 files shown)


PhilippMT:
Sorry for my confusion, this is my first PR. As of the contributing guidelines [https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/.github/CONTRIBUTING.md](url) "improvement" is a valid PR title but the CI job does not allow this value. Should I alternatively use "feat"?

node_groups.tf
```diff
@@ -331,6 +331,8 @@ module "self_managed_node_group" {
 
   for_each = { for k, v in var.self_managed_node_groups : k => v if var.create }
 
+  create = try(each.value.create, true)
```
bryantbiggs, line 334: I haven't tested this yet, but any changes we make like this should also be propagated to the EKS managed node group and Fargate profile definitions

PhilippMT: Good point, I added the feature to the managed node groups and fargate profiles but did not test it, because I am not using them anywhere

bryantbiggs:
and regarding your title, yes lets just use `feat: ...` and it should pass the check then

bryantbiggs (approved):
looks good, one last change - lets change the title to `feat: Allow conditional creation of node groups to be set within node group definitions` - thanks for the PR @PhilippMT !

should be good to go with title change @antonbabenko 👍🏽

antonbabenko:
This PR is included in version 18.5.0 :tada:

PR #1861, opened 2022-02-09
fix: Use existing node security group when one is provided

### LINKED ISSUE #1859: Confusing security group assignment for self managed nodes
## Description
When configuring a custom security group for self managed nodes, you have two variables in the root module:
```
create_node_security_group       = false
node_security_group_id           = aws_security_group.self_managed_nodes.id
```
With this set, I would expect that security group to be attached to the nodes. However, if i follow the variable in the module, I see it gets assigned to the local in `node-groups.tf`
```
node_security_group_id = local.create_node_sg ? aws_security_group.node[0].id : var.node_security_group_id
```
but that local is not used anywhere. More specifically, I would expect that it is passed in [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/43822aff06ffa7ab38f281422a3c5663e4e0b061/node_groups.tf#L408) in `vpc_security_group_ids`.

Currently, to make it work I need to set the two root variables like above and  also
```
  self_managed_node_group_defaults = {
    vpc_security_group_ids                   = [aws_security_group.self_managed_nodes.id]
    ....
    }
```

Is this intended? I think, it would be nice if I didn't also have to set it in the `defaults` of the node groups.

bryantbiggs:
## Description
- use existing node security group when one is provided

## Motivation and Context
- Closes #1859 

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -281,7 +281,7 @@ module "eks_managed_node_group" {
 
   ebs_optimized                          = try(each.value.ebs_optimized, var.eks_managed_node_group_defaults.ebs_optimized, null)
   key_name                               = try(each.value.key_name, var.eks_managed_node_group_defaults.key_name, null)
-  vpc_security_group_ids                 = compact(concat([try(aws_security_group.node[0].id, "")], try(each.value.vpc_security_group_ids, var.eks_managed_node_group_defaults.vpc_security_group_ids, [])))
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 18.5.1 :tada:

PR #1870, opened 2022-02-11
feat: Add additional output for OIDC provider (issuer URL without leading `https://`)

bryantbiggs:
## Description
- Add additional output for OIDC provider (issuer URL without leading `https://`)

## Motivation and Context
- Avoid duplicating logic of stripping leading `https://` in multiple places and match what is shown in the console. If you log into the console and navigate to IAM -> Identity providers -> select an OIDC provider, in the top left under `Summary` you will see this same value stored under the field `Provider`; hence the output name here is simply `oidc_provider`

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -74,6 +74,11 @@ output "node_security_group_id" {
 # IRSA
 ################################################################################
 
+output "oidc_provider" {
+  description = "The OpenID Connect identity provider (issuer URL without leading `https://`)"
+  value       = replace(try(aws_eks_cluster.this[0].identity[0].oidc[0].issuer, ""), "https://", "")
+}
+
 output "oidc_provider_arn" {
   description = "The ARN of the OIDC Provider if `enable_irsa = true`"
   value       = try(aws_iam_openid_connect_provider.oidc_provider[0].arn, "")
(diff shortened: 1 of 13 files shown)


outputs.tf
```diff
@@ -74,6 +74,11 @@ output "node_security_group_id" {
 # IRSA
 ################################################################################
 
+output "oidc_provider" {
+  description = "The OpenID Connect identity provider (issuer URL without leading `https://`)"
+  value       = replace(try(aws_eks_cluster.this[0].identity[0].oidc[0].issuer, ""), "https://", "")
```
antonbabenko, line 79:
```suggestion
  value       = try(replace(aws_eks_cluster.this[0].identity[0].oidc[0].issuer, "https://", ""), "")
```

Since we love `try` so much we should put it as the first function.

antonbabenko: Agree?

bryantbiggs: fixed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1870/commits/c5b3a772c14f49fc278b6761575424c9791bdecb and also added to example outputs

antonbabenko:
This PR is included in version 18.6.0 :tada:

PR #1896, opened 2022-02-22
feat: Implementing a separated nodegroup creation flow

fascenvi:
## Description

To allow a flow of creating nodegroups independently of creating clusters flow

## Motivation and Context
Recently I needed to separate the flow to provisioning EKS Clusters of the flow to provisioning EKS Nodegroups.
So, the idea was improve the use of the "create" variable and implementing the "create_nodegroups" variable to separate the flow to use when I need just create one additional nodegroups without passing on EKS cluster creation flow

Was needed to add two new variables (var.cluster_endpoint and var.cluster_auth_base64) to pass to the terraform-aws-eks module the information about the cluster that was previously created.

## Breaking Changes
There is no breaking

## How Has This Been Tested?
The tests were done in two steps where I provisioning an EKS Cluster and on a separate way, I provisioning the EKS NodeGroups passing data about the cluster. I tested the Managed NodeGroups, Fargate and Self Managed NodeGroup scenarios


bryantbiggs:
I don't follow this change - users can already create node groups/profiles separate from the control plane https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/complete/main.tf#L180-L234

fascenvi:
@bryantbiggs,
I thinking on this change to do not change the source on the root module because on my tests, when I set the var.create to false, the flow is finished with a bypass on the nodegroup modules.
If I understood, on your example, I need to pointing the path of submodules, not just set the node/profile groups

I would like to use the same root module on different pipelines, one to provisioning an EKS Cluster and another to provisioning EKS NodeGroups without any relationship between them and using different tfstates.

bryantbiggs:
I don't follow, however I am confident that all use cases are currently covered so we won't be proceeding with this proposed change

Issue #1897, opened 2022-02-23
capacity_reservation_specification not mapped properly


shupcode:
Using capacity_reservation_specification in the eks managed node group or the self managed node group throws an invalid function argument:
```
      capacity_reservation_specification = {
        capacity_reservation_target = {
          capacity_reservation_id = aws_ec2_capacity_reservation.foobar.id
        }
      }
```

```
 Error: Invalid function argument
│ 
│   on .terraform/modules/eks/modules/eks-managed-node-group/main.tf line 99, in resource "aws_launch_template" "this":
│   99:           capacity_reservation_id = lookup(capacity_reservation_target.value, "capacity_reservation_id", null)
│     ├────────────────
│     │ capacity_reservation_target.value is "cr-xxxxxxxx"
│ 
│ Invalid value for "inputMap" parameter: lookup() requires a map as the first argument.
```

Fix:
A pull request has been made to flatten the nested dynamic 'capacity_reservation_target'

antonbabenko:
This issue has been resolved in version 18.17.1 :tada:

PR #1898, opened 2022-02-23
fix: Input capacity_reservation_target should be flattened as a nested dynamic

shupcode:
Fixes the invalid function error when specifying a capacity_reservation_target in both the aws managed node groups and self managed node groups

Link to issue: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1897

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -95,7 +95,7 @@ resource "aws_launch_template" "this" {
       capacity_reservation_preference = lookup(capacity_reservation_specification.value, "capacity_reservation_preference", null)
 
       dynamic "capacity_reservation_target" {
-        for_each = lookup(capacity_reservation_specification.value, "capacity_reservation_target", [])
+        for_each = flatten([lookup(capacity_reservation_specification.value, "capacity_reservation_target", [])])
         content {
           capacity_reservation_id = lookup(capacity_reservation_target.value, "capacity_reservation_id", null)
         }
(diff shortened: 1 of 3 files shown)


bryantbiggs:
you'll need to update one of the examples to show/validate that this change works as intended

shupcode:
> you'll need to update one of the examples to show/validate that this change works as intended

I avoided updating the eks managed node group  'complete' example because this only works with on-demand instances and the 'complete' example uses a spot instance.

Therefore, I added it to the self managed node group example and terraform validate does not complain.

On my own production deployment I can confirm this works for eks managed node groups when I am specifying a node group with on-demand instances.

modules/self-managed-node-group/main.tf
```diff
@@ -95,7 +95,7 @@ resource "aws_launch_template" "this" {
       capacity_reservation_preference = lookup(capacity_reservation_specification.value, "capacity_reservation_preference", null)
 
       dynamic "capacity_reservation_target" {
-        for_each = lookup(capacity_reservation_specification.value, "capacity_reservation_target", [])
+        for_each = flatten([lookup(capacity_reservation_specification.value, "capacity_reservation_target", [])])
```
bryantbiggs, line 98:
```suggestion
        for_each = try([capacity_reservation_specification.value.capacity_reservation_target], [])
```

antonbabenko:
This issue has been resolved in version 18.17.1 :tada:

PR #1911, opened 2022-03-01
feat: Allow EKS cluster security group to be included in node group launch template

### LINKED ISSUE #1892: eks module tries to force cluster recreation if no cluster_security_group_id specified
## Description

There seems to be a regression of a very old issue https://github.com/terraform-aws-modules/terraform-aws-eks/pull/798 where if 
`cluster_security_group_id` is not specified and creation is disabled, the EKS module will try and replace the cluster every time.

Technically, the EKS module documentation does indicate `cluster_security_group_id` is a required parameter if `create_cluster_security_group` is set to `false`, however it is an optional parameter in the dependent `aws_eks_cluster` module and it should not be required since EKS creates a baseline cluster security group on cluster creation no matter what.

I think this should be as simple as wrapping this line in `compact()` -- 
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L16 to restore the previous behavior of removing any empty string values

## Versions
- Terraform:
   - Terraform v1.1.4 on darwin_amd64
 - Provider(s):
   + provider registry.terraform.io/hashicorp/aws v3.74.3
   + provider registry.terraform.io/hashicorp/tls v3.1.0
 - Module: 
   - 18.2.7

## Reproduction

- Run terraform plan/terraform apply
- After first apply, re-run terraform plan

```
      ~ vpc_config {
          ~ cluster_security_group_id = "sg-XXXXXX" -> (known after apply)
          ~ security_group_ids        = [ # forces replacement
              + "",
            ]
          ~ vpc_id                    = "vpc-YYYYY" -> (known after apply)
            # (4 unchanged attributes hidden)
        }
```

### Code Snippet to Reproduce
```
module "eks" {
  version = "18.7.2"

  cluster_name                    = "test-cluster"
...

  create_cluster_security_group = false
  create_node_security_group    = false
...
}
```

## Expected behavior

Modules should not detect any changes

## Actual behavior

Module thinks something has changed and tries to destroy

davidaah:
## Description
Allow including EKS cluster security group in module created launch templates

~In order to effectively use this feature, additional cluster security groups may not be specified at cluster creation
which does not work right now. Fix is added for that (Fixes #1892)~ Rebased on top of #1934

## Motivation and Context
The behavior of EKS's launch template support today is that if any node security groups are
specified in the launch template configuration, EKS will not automatically add the cluster security group.
If no security groups are specified, the cluster security group is added by default.
(ref: https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html)

For users that want to preserve similar behavior, when including additional node security groups,
optionally allow the cluster security group to be added to the launch template generated
by the EKS Cluster terraform module

## Breaking Changes
* existing behavior preserved with default value (`include_cluster_security_group = false`)

## How Has This Been Tested?

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -181,6 +181,12 @@ variable "cluster_security_group_tags" {
   default     = {}
 }
 
+variable "include_cluster_security_group" {
+  description = "Determines if cluster security group should be included in node launch templates"
+  type        = bool
+  default     = false
+}
+
 ################################################################################
 # EKS IPV6 CNI Policy
 ################################################################################
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -107,6 +107,10 @@ module "eks" {
   }
 
   # EKS Managed Node Group(s)
+
+  # Extend list of configured security groups on nodes to include the AWS-created cluster security group
+  include_cluster_security_group = true
+
   eks_managed_node_group_defaults = {
     ami_type               = "AL2_x86_64"
     disk_size              = 50
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -223,6 +223,9 @@ module "fargate_profile" {
 # EKS Managed Node Group
 ################################################################################
 
+locals {
+  cluster_base_security_group_id = var.include_cluster_security_group ? aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id : null
+}
 module "eks_managed_node_group" {
   source = "./modules/eks-managed-node-group"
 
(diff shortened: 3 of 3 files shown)


variables.tf
```diff
@@ -175,6 +175,12 @@ variable "cluster_security_group_tags" {
   default     = {}
 }
 
+variable "include_cluster_security_group" {
```
davidaah, line 178: note: open to suggestions on naming here. was trying to come up with something that wasn't overly verbose

bryantbiggs:
I've opened #1934 which seems to be necessary and I've validated that it is required in some instances. 

However, I don't follow the "node group" security groups. Is this for EKS managed node groups, self managed node groups, or both? Are you creating a launch template or no?

bryantbiggs:
I just tried this in the examples and the node gets the EKS supplied security group `EKS created security group applied to ENI that is attached to EKS Control Plane master nodes, as well as any managed workloads.`

```hcl
module "eks" {
  source = "../.."

  cluster_name    = local.name
  cluster_version = local.cluster_version

  enable_irsa = false

  # https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#launch-template-security-groups
  create_cluster_security_group = false
  create_node_security_group    = false

  eks_managed_node_groups = {
    # Default node group - as provided by AWS EKS
    default_node_group = {
      # By default, the module creates a launch template to ensure tags are propagated to instances, etc.,
      # so we need to disable it to use the default template provided by the AWS EKS managed node group service
      create_launch_template = false
      launch_template_name   = ""

      create_security_group = false
    }
  }

  tags = local.tags
}
```

davidaah:
@bryantbiggs just wanted to follow up to see if you could take another look at this when you have some time? currently i'm blocked from using this module with a custom node SG and cluster SG

davidaah:
@bryantbiggs hopefully this helps. i think i found the right place to put in the examples... but let me also explain pre/post behavior:

Pre my change (or with change, variable set to false):
- Example would vend managed node group with just `aws_security_group.additional.id`. In this scenario in order for the nodes to properly join the cluster this security group must contain all the rules needed to access the control plane OR you need to use `cluster_security_group_additional_rules` and `node_security_group_additional_rules`

Post change:
- Example will vend nodes with BOTH `aws_security_group.additional.id` and the AWS created cluster SG attached. The reason we need this (as opposed to using the module to manipulate the SGs) is we currently manage these custom SGs with separate automation in our organization which necessitates us not modifying them in-line via Terraform

bryantbiggs:
I've opened #1952 which should satisfy this - there are a few differences I've made so that we continue the patterns established in terms of the flexibility users have in how they managed their node groups

davidaah:
thanks @bryantbiggs that change looks perfectly reasonable to me. much appreciated for helping to bring this over the line!

PR #1915, opened 2022-03-02
feat: Add additional IAM policy to allow cluster role to use KMS key provided for cluster encryption

bryantbiggs:
## Description
- Add additional IAM policy to allow cluster role to use KMS key provided for cluster encryption
- Fix bug in EKS managed node group example where the IPv6 permissions need to be created on new cluster creation despite the use of IRSA for VPC CNI addon

## Motivation and Context
- Closes #1894
- Closes #1913 

## Breaking Changes
- No

## How Has This Been Tested?
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects
	- ✅ `complete` example
	- ✅ `eks_managed_node_group` example
	- ✅ `irsa_autoscale_refresh` example

PR #1917, opened 2022-03-02
fix: Ensure that cluster encryption policy resources are only relevant when creating the IAM role

bryantbiggs:
## Description
- Ensure that cluster encryption policy resources are only relevant when creating the IAM role

## Motivation and Context
- Bug introduced on #1915 where the create role check was not added to the new policy & attachment so v18.8.0 is trying to attach a policy to a non-existent role when users are providing their own cluster role

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -170,6 +170,7 @@ resource "aws_iam_openid_connect_provider" "oidc_provider" {
 ################################################################################
 
 locals {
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 18.8.1 :tada:

Issue #1925, opened 2022-03-08
prefix_separator for nodegroups


christophemorio:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.
When upgrading prior v18 deployments, I cannot use prefix naming mechanism, because of `"-"` set at the end of the string.
This affects resources inside modules `eks-managed-node-group`, `fargate-profile` and `self-managed-node-group`:

- aws_launch_template
- aws_eks_node_group
- aws_security_group
- aws_iam_role
- aws_autoscaling_group
- aws_iam_instance_profile

## Describe the solution you'd like.
Repeat as done for main module, with variable `prefix_separator`.
Inside modules, replace hardcoded `"-"` by `${var.prefix_separator}`

## Describe alternatives you've considered.
Disable prefix and explicitly set full resources name node groups input var.

```
module "eks" {
  eks_managed_node_groups = {
    group1 = {
      iam_role_use_name_prefix       = false
      iam_role_name                  = "foo2020051415392924430000000a"
      security_group_use_name_prefix = false
      security_group_name            = "foo2020051415392924830000000b"
    }
  }
...
}
```

## Additional context

Existing implementation for main module :
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/27f99f03a7004407625f61eb030833f98326645d/variables.tf#L13-L17
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/27f99f03a7004407625f61eb030833f98326645d/main.tf#L111

bryantbiggs:
as discussed in #1792, we will not be implementing this change

PR #1926, opened 2022-03-08
feat: Input prefix_separator for nodegroups resources

christophemorio:
## Description
Repeat as done for main module, with variable prefix_separator.
Inside modules, replace hardcoded `"-"` by `${var.prefix_separator}`.

## Motivation and Context
When upgrading prior v18 deployments, I cannot use prefix naming mechanism, because of "-" set at the end of the string.
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1925

## Breaking Changes
No

## How Has This Been Tested?


bryantbiggs:
as discussed in #1792, we will not be implementing this change

PR #1936, opened 2022-03-14
fix: Fargate profile execution role output

IvanDechovsky:
Signed-off-by: Ivan Dechovski <ivan.dechovski@kpn.com>

## Description
Update the fargate profile execution role ARN output to be fetched from the IAM role attribute or the IAM role ARN variable, instead of referencing the EKS fargate profile resource attribute. 

## Motivation and Context
This change allows us to get the output in the root module `aws_auth_configmap_yaml` as soon as the fargate IAM role ARN is available. Currently, the output `aws_auth_configmap_yaml` is only made available when the profile is created, delaying us with updating the aws-auth configmap. In some cases, creating a fargate profile can take upwards of 10minutes during which non of the node groups would have permissions to join the cluster.

### DIFF
--- a/modules/fargate-profile/outputs.tf
+++ b/modules/fargate-profile/outputs.tf
@@ -38,5 +38,5 @@ output "fargate_profile_status" {
 
 output "fargate_profile_pod_execution_role_arn" {
   description = "Amazon Resource Name (ARN) of the EKS Fargate Profile Pod execution role ARN"
-  value       = try(aws_eks_fargate_profile.this[0].pod_execution_role_arn, "")
+  value       = try(var.create_iam_role ? aws_iam_role.this[0].arn : var.iam_role_arn, "")
 }


modules/fargate-profile/outputs.tf
```diff
@@ -38,5 +38,5 @@ output "fargate_profile_status" {
 
 output "fargate_profile_pod_execution_role_arn" {
   description = "Amazon Resource Name (ARN) of the EKS Fargate Profile Pod execution role ARN"
-  value       = try(aws_eks_fargate_profile.this[0].pod_execution_role_arn, "")
+  value       = try(var.create_iam_role ? aws_iam_role.this[0].arn : var.iam_role_arn, "")
```
bryantbiggs, line 41: are you confusing this with line 12? we don't want to make a change like this here

bryantbiggs (changes requested):

modules/fargate-profile/outputs.tf
IvanDechovsky: I did not confuse it but upon further testing, the suggested change does not fix the issue. The problem at the moment is that if you are creating/updating a self managed group and fargate profile within the same apply, `aws_auth_configmap_yaml` is not made available until the fargate profile is created. Creating a profile can take 10+ minutes, during which the self managed nodes cannot join the cluster as the aws-auth config map is not updated. 

However, the way the output loops over the module of fargate_profile `[for group in module.fargate_profile : group.fargate_profile_pod_execution_role_arn]` means that no outputs are made available until all changes in that module are complete.

IvanDechovsky: I'm going to close the PR as it does not accomplish the indented result, however it would be nice if there was way around it but I cannot see how.

bryantbiggs: AWS handles updating the aws-auth configmap for Fargate profiles and EKS managed node groups.

IvanDechovsky: AWS does create the config map, if it does not exist but I still think you are missing the problem that exists here. The output in the root module `aws_auth_configmap_yaml` waits for all fargate profiles to be ready before the output is made available.  If you rely on `aws_auth_configmap_yaml` to generate your aws-auth map, that means that changes to any self_managed groups would only be reflected in the output, once fargate profile has finished.

What would resolve the problem is if the IAM ARN of each fargate profile is made available before the profile itself has finished creating.

bryantbiggs: Thats down to Terraform's resolution of the dependency graph as to when that output is made available - nothing that we can control here. You can construct your own ARN to work around this

PR #1942, opened 2022-03-17
feat: Add prefix_separator to self-managed-node-group for migration

psibi:
## Description

This does the relevant changes to help in migration to the latest EKS module from v17.x  for self managed node group.

## Motivation and Context

This change is done for migrating from v17.x module. The resources created by the older module didn't have the `-` prefix. We already have [prefix_separator](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/66200f1c1d6dca432c2c87b2a657bb59aaed9c68/variables.tf#L13) variable in the main module and this propagates that information to the `self-managed-node-group` and uses that information in various places like ASG name, IAM role name. Without this change, the module tries to recreate the ASG etc which isn't good.

## Breaking Changes

No, this doesn't break backward compatibility.

## How Has This Been Tested?

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -48,7 +48,7 @@ resource "aws_launch_template" "this" {
   count = var.create && var.create_launch_template ? 1 : 0
 
   name        = var.launch_template_use_name_prefix ? null : local.launch_template_name_int
-  name_prefix = var.launch_template_use_name_prefix ? "${local.launch_template_name_int}-" : null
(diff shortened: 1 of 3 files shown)


bryantbiggs:
thanks for the pull request @psibi - however, this change was intentionally made during a breaking change (v18.0.0). Node groups are generally ephemeral in nature and can easily be replaced so we will not be merging this change

psibi:
@bryantbiggs Thanks for the feedback! But this PR doesn't change anything in the module, but only makes it easy for me to use the upstream module and also use the existing names for the ASG, IAM role etc without it getting replaced. By introducing the `prefix_separator` that's already present in the main module to the sub module, it just propagates it..

PR #1946, opened 2022-03-18
feat: Allow users to specify default launch template name in node groups

### LINKED ISSUE #1945: Node group get launch_template_name from node_group_defaults

No

We found that `launch_template_name` in node group can not inherit the value from `eks_managed_node_group_defaults` like other launch_template related fields like `create_launch_template` or `launch_template_version`: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v18.10.2/node_groups.tf#L276. Instead, it uses the `key` of the node groups map. Is this intentional? 

In our cace, we have multiple node groups per cluster using the same template, this means we need to add the same field for each node groups every time.

Make `launch_template_name` to inherit value from `eks_managed_node_group_defaults`.

Define `launch_template_name` in every node groups.

bryantbiggs:
## Description
- Allow users to specify default launch template name in node groups

## Motivation and Context
- Closes #1945 

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -273,7 +273,7 @@ module "eks_managed_node_group" {
 
   # Launch Template
   create_launch_template          = try(each.value.create_launch_template, var.eks_managed_node_group_defaults.create_launch_template, true)
-  launch_template_name            = try(each.value.launch_template_name, each.key)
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 18.11.0 :tada:

PR #1952, opened 2022-03-22
feat: Allow users to selectively attach the EKS created cluster primary security group to nodes

bryantbiggs:
## Description
- Add new attribute `attach_cluster_primary_security_group` to EKS managed node group and self managed node group root module definitions. When set to `true`, the security group that is automatically created by the EKS control plane is passed to the node group sub-module where it is included in the launch template for nodes to utilize when launched via the added `cluster_primary_security_group_id` variable (note - no changes to Fargate profiles since they use this by default and the API currently doesn't allow for managing profile security groups)
- Updated `complete` example to demonstrate as well as security group section of docs

## Motivation and Context
- Closes #1911

## Breaking Changes
- No

## How Has This Been Tested?
	- Validated with `complete` example


bryantbiggs:
@antonbabenko if you get some time, this one should be ready to go

Issue #1959, opened 2022-03-24
self managed node group fails with default `cluster_version = null`


aidanmelen:
## Description

The self managed node group uses the [cluster_version input](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/self-managed-node-group#input_cluster_version) used to lookup default AMI ID if one is not provided. But the default for `cluster_version` is `null`.

The `data.aws_ami.eks_default` data resource will fail to [filter](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/self-managed-node-group/main.tf#L10) with the null value. e.g. `amazon-eks-node-null-v*`

## Versions

 - Terraform: v1.1.7

 - Provider(s): `terraform-aws-modules/eks/aws` (0.18.3)

 - Sub-Module: `self-managed-node-group`

## Reproduction

1. run `terraform-aws-modules/eks/aws` with `self_managed_node_groups` sub-module and don't set `cluster_version`

### Code Snippet to Reproduce

```hcl
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = ">= 18.0.0"

  cluster_name    = local.name
  # cluster_version = "1.21"
  vpc_id          = module.vpc.vpc_id
  subnet_ids      = module.vpc.private_subnets

  self_managed_node_groups = {
    boo = {
      instance_type = "t3.medium"
      instance_market_options = {
        market_type = "spot"
      }
    }
  }
}
```

## Expected behavior
defaults to the cluster version.

## Actual behavior
defaults to null

## Additional context
one possible solution is to retrieve the `cluster_version` from a data resource and set that as the default

```hcl
data "aws_eks_cluster" "example" {
  name = "example"
}

output "dynamic_cluster_version" {
  value = data.aws_eks_cluster.example.version
}
```

bryantbiggs:
I think its better if we just rely on users specifying the version - it aligns with the other sub-modules as well

bryantbiggs:
> The behavior of the self managed node group is inconsistent with the other sub-modules. for example

Self managed node groups by their very nature are different from EKS managed or Fargate profiles because all management is up to users

PR #1960, opened 2022-03-24
self managed node group fails with default cluster_version = null

aidanmelen:
self managed node group fails with default cluster_version = null

## Description

The self managed node group uses the [cluster_version input](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/self-managed-node-group#input_cluster_version) used to lookup default AMI ID if one is not provided. But the default for `cluster_version` is `null`.

The `data.aws_ami.eks_default` data resource will fail to [filter](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/self-managed-node-group/main.tf#L10) with the null value. e.g. `amazon-eks-node-null-v*`

## Motivation and Context

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1959

## Breaking Changes

No

## How Has This Been Tested?
- I tested locally.

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -2,12 +2,15 @@ data "aws_partition" "current" {}
 
 data "aws_caller_identity" "current" {}
 
+data "aws_eks_cluster" "cluster" {
+  name = var.cluster_name
+}
 data "aws_ami" "eks_default" {
   count = var.create ? 1 : 0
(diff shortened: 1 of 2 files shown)


bryantbiggs:
see comment https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1959#issuecomment-1077543614

PR #1962, opened 2022-03-24
fix: The data.aws_ami.eks_default filter requires sting and got null

aidanmelen:
## Description
The `data.aws_ami.eks_default` filter requires sting and get null by default.

## Motivation and Context

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1959#issuecomment-1077670471

## Breaking Changes
N/A

## How Has This Been Tested?
- comment out the `cluster_version` in the [self_managed_node_group_example](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/self_managed_node_group/main.tf#L27) and let it default to null
- run a plan
```bash
╷
│ Error: Invalid template interpolation value
│ 
│   on ../../modules/self-managed-node-group/main.tf line 10, in data "aws_ami" "eks_default":
│   10:     values = ["amazon-eks-node-${var.cluster_version}-v*"]
│     ├────────────────
│     │ var.cluster_version is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
╷
│ Error: Invalid template interpolation value
│ 
│   on ../../modules/self-managed-node-group/main.tf line 10, in data "aws_ami" "eks_default":
│   10:     values = ["amazon-eks-node-${var.cluster_version}-v*"]
│     ├────────────────
│     │ var.cluster_version is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
╷
│ Error: Invalid template interpolation value
│ 
│   on ../../modules/self-managed-node-group/main.tf line 10, in data "aws_ami" "eks_default":
│   10:     values = ["amazon-eks-node-${var.cluster_version}-v*"]
│     ├────────────────
│     │ var.cluster_version is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
╷
│ Error: Invalid template interpolation value
│ 
│   on ../../modules/self-managed-node-group/main.tf line 10, in data "aws_ami" "eks_default":
│   10:     values = ["amazon-eks-node-${var.cluster_version}-v*"]
│     ├────────────────
│     │ var.cluster_version is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
```

PR #1963, opened 2022-03-24
fix: Default to cluster version for EKS and self managed node groups when a `cluster_version` is not specified

bryantbiggs:
## Description
- Default to cluster version for EKS and self managed node groups when a `cluster_version` is not specified
- Add `cluster_version` output for users to use when using separate node group sub-modules (see updated `complete` example)
- Correct doc change for Karpenter IRSA example

## Motivation and Context
- Relates to #1959
- Relates to #1960
- Closes #1962
- The `version` on the EKS cluster resource is optional and will default to the latest Kubernetes version supported by EKS. However, if a version is not specified in the root module via `var.cluster_version`, then a version is not propagated to the sub-modules which can result in an error. By changing from the `var.cluster_version` to the EKS cluster resource output of `version` and using that as the default, we are able to achieve the same functionality when users do not specify a `var.cluster_version`

## Breaking Changes
- No

## How Has This Been Tested?
	- Updated `complete` example and verified functionality

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -27,6 +27,11 @@ output "cluster_oidc_issuer_url" {
   value       = try(aws_eks_cluster.this[0].identity[0].oidc[0].issuer, "")
 }
 
+output "cluster_version" {
+  description = "The Kubernetes version for the cluster"
+  value       = try(aws_eks_cluster.this[0].version, "")
+}
+
 output "cluster_platform_version" {
   description = "Platform version for the cluster"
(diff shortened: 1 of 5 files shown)


outputs.tf
```diff
@@ -27,6 +27,11 @@ output "cluster_oidc_issuer_url" {
   value       = try(aws_eks_cluster.this[0].identity[0].oidc[0].issuer, "")
 }
 
+output "cluster_version" {
```
aidanmelen, line 30: nice!

PR #1966, opened 2022-03-24
feat: Update TLS provider and remove unnecessary cloud init version requirements

bryantbiggs:
## Description
- Update TLS provider to require at least v3.0; there are no code changes required for this
- Remove unnecessary cloud init version requirements in the EKS managed node group and self managed node group modules; the version is constrained in the user data module which both of those module use, rendering their requirements redundant

## Motivation and Context
- Keep providers up to date and clean up redundant requirements

## Breaking Changes
- No

## How Has This Been Tested?
	- Tested on `eks-managed-node-group` example

### DIFF
--- a/examples/eks_managed_node_group/main.tf
+++ b/examples/eks_managed_node_group/main.tf
@@ -568,6 +568,12 @@ resource "aws_launch_template" "external" {
     enabled = true
   }
 
+  # Disabling due to https://github.com/hashicorp/terraform-provider-aws/issues/23766
+  # network_interfaces {
+  #   associate_public_ip_address = false
+  #   delete_on_termination       = true
+  # }
+
   # if you want to use a custom AMI
   # image_id      = var.ami_id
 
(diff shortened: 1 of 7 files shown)


examples/eks_managed_node_group/main.tf
```diff
@@ -568,11 +568,6 @@ resource "aws_launch_template" "external" {
     enabled = true
   }
 
-  network_interfaces {
```
bryantbiggs, line 571: I'm not sure why this example started complaining about setting a device index so just removing since its not required for the example

antonbabenko: There was a bug in the AWS provider about it - https://github.com/hashicorp/terraform-provider-aws/issues/23766

We hit it in one of the projects but I don't remember the resolution (probably downgraded version). Maybe leave a comment around this section with the link to the issue, so we know what was it in the future?

bryantbiggs: ah, good to know - thought I was losing it 😅 

Added back the block but commented it out with a link to the issue - https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1966/commits/bf45e14950475d3c527618f48a20c940ab621c1e

PR #1968, opened 2022-03-25
feat: Support default_tags in aws_autoscaling_group and launch_template

### LINKED ISSUE #1916: Support default_tags in aws_autoscaling_group
Tags created in the `default_tags` block in the AWS provider are not applied to `aws_autoscaling_group`.

Apply `default_tags` to the `aws_autoscaling_group` in the `self-managed-node-group` module.

## Additional context
 https://learn.hashicorp.com/tutorials/terraform/aws-default-tags?in=terraform/aws#propagate-default-tags-to-auto-scaling-group

raizyr:
## Description
 * Add new attribute `launch_template_use_default_tags` to EKS managed node group and self managed node group module definitions. When set to true, the default tags configured on the provider will be added to the launch_template `tag_specifications`.  
 * Add new attribute `asg_use_default_tags` to EKS self managed node group module definition.  When set to `true`, the default tags configured on the provider will be added to the autoscaling_group tags.
 * Update the root module to look at both `var.eks_managed_node_groups or var.eks_managed_node_group_defaults` and `var.self_managed_node_groups or var.self_managed_node_group_defaults` and pass the result to the apprioriate node group module.

## Motivation and Context

Fixes #1916 

## Breaking Changes

No

## How Has This Been Tested?
- [✅ ] I have tested and validated these changes using one or more of the provided `examples/*` projects


modules/eks-managed-node-group/main.tf
```diff
@@ -238,7 +242,7 @@ resource "aws_launch_template" "this" {
     for_each = toset(["instance", "volume", "network-interface"])
     content {
       resource_type = tag_specifications.key
-      tags          = merge(var.tags, { Name = var.name }, var.launch_template_tags)
+      tags          = merge(local.launch_template_default_tags, { Name = var.name }, var.launch_template_tags)
```
bryantbiggs, line 245: why do we need to set this here? from my understanding, only autoscaling groups behave differently when it comes to default tags

modules/self-managed-node-group/main.tf
```diff
@@ -229,7 +233,7 @@ resource "aws_launch_template" "this" {
     for_each = toset(["instance", "volume", "network-interface"])
     content {
       resource_type = tag_specifications.key
-      tags          = merge(var.tags, { Name = var.name }, var.launch_template_tags)
+      tags          = merge(local.launch_template_default_tags, { Name = var.name }, var.launch_template_tags)
```
bryantbiggs, line 236: same comment as above - why set on launch template explicitly?

modules/self-managed-node-group/main.tf
```diff
@@ -255,6 +259,8 @@ locals {
   launch_template_name = try(aws_launch_template.this[0].name, var.launch_template_name)
   # Change order to allow users to set version priority before using defaults
   launch_template_version = coalesce(var.launch_template_version, try(aws_launch_template.this[0].default_version, "$Default"))
+
+  asg_default_tags = try(var.asg_use_default_tags, false) ? merge(data.aws_default_tags.current.tags, var.tags) : var.tags
```
bryantbiggs, line 263: This won't work like this, it will need to be:
```suggestion
  default_tags = var.use_default_tags ? merge(data.aws_default_tags.current.tags, var.tags) : var.tags
```

raizyr: I disagree that it won't work, as it seemed to work fine when I tested it.  Also I copied the `try(var.my_var, false) ? "Do true" : "Do false"` pattern from the other Boolean vars in the `node_groups.tf` file.

bryantbiggs: `try()` works on "errors" and returns the first result that does not throw an error https://www.terraform.io/language/functions/try

There is no error here so no matter what happens, it will always take the variable listed first in the sequence so there is no point in listing it this way

raizyr: Yep, you're 100% correct, and as I was grabbing my laptop to look into the `launch_template` tags, I realized the mistake I made.   In `node_groups.tf` the `try()` works because it's attempting to access a key on another var,  but you're right that it won't work in the child modules.  I'll push a fix shortly.

modules/eks-managed-node-group/main.tf
raizyr: I double checked what I'd been seeing on Friday, and while the provider default tags do get automatically included in the `launch_template` resource's tags,  they do not get included in the [tag_specifications](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#tag-specifications) of the `launch_template`.   Those instead act much like the ASG tag specifications.  Adding the ability to include the default tags to the `launch_template` tag specifications ensures that any instances created from the `launch_template` will also have the default tags set on the instance itself.

modules/self-managed-node-group/main.tf
raizyr: see above

modules/self-managed-node-group/main.tf
raizyr: I fixed the ternary logic.

modules/eks-managed-node-group/main.tf
raizyr: Would a different variable name make sense to make it clearer?  Maybe `tag_specification_use_default_tags` or something similar?

bryantbiggs: I think for now we leave off the launch template changes - this matches what we have in the autoscaling module (which at some point will get folded into this module) and just add to the autoscaling group

raizyr: I'm very confused.  The module as it exists right now is not able to correctly set tags for resources created by neither an autoscaling group nor a launch template.  Are you saying that only the autoscaling group use case should be fixed?

bryantbiggs: 1. We cannot control the autoscaling groups created by the EKS managed node group service so there is no solution for default tags on that resource currently other than using the external autoscaling group tags resource (has to be external to the module https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1558#issuecomment-1030640207)
2. We should mirror what we did in the autoscaling group module for the self-managed node group module https://github.com/terraform-aws-modules/terraform-aws-autoscaling/blob/3c928b8cfb2be46ff756be350fb26c33ee1a640a/main.tf#L10-L15 which is the same as what the Terraform docs specify. We only add the default tags to the autoscaling group resource

bryantbiggs: I understand there are many issues related to default tags right now which is why I am cautious regarding what gets introduced. The Terraform docs only offer the caveat for autoscaling groups https://www.hashicorp.com/blog/default-tags-in-the-terraform-aws-provider and not launch templates and this is what we have mimicked in the autoscaling group module without issue. So for now, I think that is a safe path if we want to introduce that change

raizyr: Ok, unfortunately I need to maintain my company fork with the launch template changes, since that is the issue I needed to solve when I stumbled across the open issue about the asg.   I'll try to get a moment to pull just the asg changes over to a fork on my personal github and submit a new PR.

PR #1985, opened 2022-04-04
feat: Add `create_before_destroy` lifecycle hook to security groups created

bryantbiggs:
## Description
- Add `create_before_destroy` lifecycle hook to security groups created

## Motivation and Context
- This came up in testing how to remove the node group specific security group that is mostly unused by users. When trying to set `create_security_group` to `false` after it defaults to `true` and has been created, the security group is unable to detach and delete. Reading through the numerous issues, I am hopeful this might work for the autoscaling groups. If not, it should be in place regardless for when users re-name their security groups and they need to be re-created

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -129,6 +129,10 @@ resource "aws_security_group" "cluster" {
     { "Name" = local.cluster_sg_name },
     var.cluster_security_group_tags
   )
+
+  lifecycle {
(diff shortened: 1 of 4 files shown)


antonbabenko:
This PR is included in version 18.19.0 :tada:

PR #1998, opened 2022-04-08
feat: Resolve terraform module issues for nginx and kube2iam

brunzefb:
## Description
 - Allowed disabling of automatic tags for the node security group.  This is to
  overcome an issue with nginx ingress controller. The ingress controller
  refuses to ensure the load balancer if the tags are present.

- Improved the complete example, by adding some comments, and showing
  how to add metadata_options to allow metadata introspection (IMDSv1)
  on ec2 worker nodes, which is required if you use kube2iam.

## Motivation and Context
To help other user overcome issues that I encountered, specifically with nginx ingress controller and kube2iam deployments

## Breaking Changes
No.

## How Has This Been Tested?
- [ X ] I have tested and validated these changes using one or more of the provided `examples/*` projects

This will need documentation changes -- I have just used this code locally, not actually tested it with the pre-commit.


bryantbiggs:
hi @brunzefb - thanks for the PR. We will have some updated documentation coming soon for some of the common addons and how they relate to this module (what settings users most likely will need to make to ensure the addon works as intended). However, for now, I don't think we'll be taking on these changes proposed here

bryantbiggs:
hi @brunzefb - thanks for the PR. We will have some updated documentation coming soon for some of the common addons and how they relate to this module (what settings users most likely will need to make to ensure the addon works as intended). However, for now, I don't think we'll be taking on these changes proposed here

brunzefb:
@bryantbiggs -- Its always for the project to decide what PR's to take or not.  I was just using the module, found some issues using NGINX reverse proxy and kube2iam, and wanted to share, so that other users would not encounter these same, and somewhat difficult-to-resolve issues. If you have this covered somewhere else, that's great.

bryantbiggs:
yes, and I do appreciate the PR. However, we have this weird situation with `examples/` where the contents there are fulfilling two purposes that tend to be in conflict:
1. Demonstrate to users how to use the module(s)
2. Test all the various levels of functionality of the module(s)

If you look here, I doubt you would find any of the examples in a production environment as they are written. This is because of the extensive testing necessary for a module of this complexity. So what I am working on is trying to split that out so that we have a place to guide users for usage, best practices, common questions - this is where the details you have highlighted would go. I don't have something yet, but hopefully soon.

Issue #2006, opened 2022-04-11
Security group tags created by EKS could be overridden


pcheung-sfdc:
## Description

Starting with 18.14.0, this introduced a [resource](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/7369f8a19e3966b1edafa9cead972fbad6b56f80/main.tf#L62) to override tags of the cluster primary SG. This could be used to override a tag used by EKS to delete the SG when destroying the cluster, leaving the SG lingering and unmanaged, which can prevent destroying the VPC it's in.

- [ X] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: 18.14.0 and above

- Terraform version: v1.1.7

- Provider version(s): ~v4.0

## Reproduction Code [Required]

Pass in a "Name" tag to the module like for example:
```  
module "eks" {
...
  tags = {
    Name = "foo"
  }
}
```

## Expected behavior

The cluster primary SG is automatically generated when creating an EKS cluster is also automatically removed when destroying the cluster.

## Actual behavior

The cluster primary SG is not being destroyed if its tags are overridden.

bryantbiggs:
you should be able to re-override using `cluster_tags`

pcheung-sfdc:
Any good reason not to set the cluster_tags? 
If not, I can create a PR to specify the `cluster_tags`

bryantbiggs:
No - we try to provide an "additional" tag for select resources where users might want to provide tags specific to that resource. `var.tags` is generic across all resources and `var.cluster_tags` is applied to just the cluster (i.e. - control plane)

antonbabenko:
This issue has been resolved in version 18.20.2 :tada:

PR #2008, opened 2022-04-11
fix: Possible fix for issue #2006

pcheung-sfdc:
## Description
see issue #2006 

## Motivation and Context
see issue #2006 

## Breaking Changes
no breaking changes

## How Has This Been Tested?
- [x ] I have tested and validated these changes using one or more of the provided `examples/*` projects
- Also tested on my own module using this eks module


bryantbiggs:
hi @pcheung-sfdc - I think there was a little misunderstanding. What I was suggesting in the issue was to use `cluster_tags` to solve your issue. So if you are setting a global name tag like:

```hcl
module "eks" {
  name = var.cluster_name
	
  tags = {
    Name = "foo"
  }
}
```

You should be able to resolve your issue with:
```hcl
module "eks" {
  name = var.cluster_name
	
  tags = {
    Name = "foo"
  }
  
  cluster_tags = {
    Name = var.cluster_name
  }
}
```

pcheung-sfdc:
@bryantbiggs but the name tag is not just the cluster name, it's the security group name. If `var.cluster_tags` is only to be used by the security group then what you propose would make sense, but if `cluster_tags` is used for any other resource, having `Name` be a security group name wouldn't make sense for resources that are not security groups

pcheung-sfdc:
it's not the same name as the cluster name. It's the `cluster-name` and some prefix automatically generated (eks-cluster-sg), as well as a suffix (maybe a timestamp). So if we do what you propose, it would make a generic tag `Name` to represent only a very specific resource, also we have no way of determining what the suffix would be beforehand.

bryantbiggs:
ah, not sure why I thought I saw the name as the cluster name earlier but you are correct. one sec I think I know how to fix/avoid this 
image

PR #2034, opened 2022-04-21
fix: Add conditional variable to allow users to opt out of tagging cluster primary security group

### LINKED ISSUE #2013: conditionally tag cluster_primary_security_group
Request is for terraform-aws-eks module aws_ec2_tag resource logic and not specifically to provider feature.

Module successfully creates eks cluster and cluster primary security group.  The `aws_ec2_tag` resource fails to tag the cluster_primary_security_group_id.  

`Error: error creating EC2 resource (sg-xxx) tag (TagName): error tagging resource (sg-xx): UnauthorizedOperation: You are not authorized to perform this operation.`

Reason - Our IAM policy allows tagging only during creation time, and hence fails when `aws_ec2_tag` is executed post creation.

trigger `aws_ec2_tag` resource  on a need basis?  

```
variable "create_additional_tags_for_cluster_security_group" {
default = true
}

resource "aws_ec2_tag" "cluster_primary_security_group" {
  # This should not affect the name of the cluster primary security group
  # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2006
  # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2008
  for_each = var.create_additional_tags_for_cluster_security_group ? : { for k, v in merge(var.tags, var.cluster_tags) : k => v if local.create && k != "Name" } : {}

  resource_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
  key         = each.key
  value       = each.value
}
```

Alternate options will be expanding our security policy or removing the tags.  Removing tags is not an option, and security group actions will be a challenge since the cluster primary security group name suffix is dynamic..

## Additional context
With the option to opt tagging, we can use the default 3 tags provisioned as part of cluster_security_group creation, and can start using the `aws_ec2_tag` once we are able to specify a static cluster_primary_security_group name.

bryantbiggs:
## Description
- Add conditional variable to allow users to opt out of tagging cluster primary security group
- Correct usage of `disk_size` which has been shown incorrectly in various examples and come up in prior issues causing confusion
  - `disk_size` is only used when using the default launch template created by the EKS managed node group service. This module defaults to creating a custom launch template since this provide a more feature complete experience for users and enables missing features like instance tagging.

## Motivation and Context
- Closes #2013

## Breaking Changes
- No

## How Has This Been Tested?


variables.tf
```diff
@@ -101,6 +101,12 @@ variable "cluster_tags" {
   default     = {}
 }
 
+variable "tag_cluster_primary_security_group" {
```
antonbabenko, line 104: I propose that instead of using `tag_...`, it is better to prefix with the verb we normally use. For example, `create_tag_cluster_primary_security_group` or `enable_tag_cluster_primary_security_group` or `manage_tag_cluster_primary_security_group`?

bryantbiggs: good point, let me update

bryantbiggs: updated in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2034/commits/f2dda4002bab0bb0e852bf567a76673ee2d08e73

antonbabenko:
This PR is included in version 18.20.5 :tada:

PR #2046, opened 2022-04-28
fix: Use null as default for prefix separator and cluster security group description

valentindeaconu:
## Description
Change default values for input variables `prefix_separator` and `cluster_security_group_descripton` to null and then use locals to provide the default values.

## Motivation and Context
This change will allow users that use a wrapper module to encapsulate the v17 compatibility with a single input variable without requiring them to check if the default values changed from one version to another.

Let's assume that the wrapper module has an input variable defined as the following:
```
variable "use_v17_compatibility" {
  type     = bool
  default  = false
}
```

Then the module block:
```
module "cluster" {
  
  prefix_separator                   = var.use_v17_compatibility ? "" : "-"
  iam_role_name                      = var.use_v17_compatibility ? "my-cluster" : null
  cluster_security_group_name        = var.use_v17_compatibility ? "my-cluster" : null
  cluster_security_group_description = var.use_v17_compatibility ? "EKS cluster security group." : "EKS cluster security group"

  # ...
}
```

This change will allow users to pass the `null` value and trust the module to take care of the default values:
```
module "cluster" {
  
  prefix_separator                   = var.use_v17_compatibility ? "" : null
  iam_role_name                      = var.use_v17_compatibility ? "my-cluster" : null
  cluster_security_group_name        = var.use_v17_compatibility ? "my-cluster" : null
  cluster_security_group_description = var.use_v17_compatibility ? "EKS cluster security group." : null

  # ...
}
```

Without this change, every time the module gets updated, the users have to check for the default values and replicate them into their wrapper.

## Breaking Changes
No.

## How Has This Been Tested?
- [x] I have tested and validated these changes using my own cluster


bryantbiggs:
Thank you for the PR but unfortunately this would be categorized as a breaking change due to reverting the variable default values which are known to cause resource replacement. 

However, either way - we won't be making any further changes to these variables for prior version support as they are sufficient as is for now

PR #2065, opened 2022-05-10
fix: Use prefix_separator in self-managed node groups launch templates

gferon:
## Description

`prefix_separator` was introduced in https://github.com/terraform-aws-modules/terraform-aws-eks/commit/7089c71e64dbae281435629e19d647ae6952f9ac to help with migration from v17 to v18. When using self-managed node groups, this problem also appeared at the parameter wasn't propagated.

## Motivation and Context

Propagating this change helps avoiding node groups resource to be forcefully recreated.

## Breaking Changes

None expected, since it uses the existing parameter which defaults to `-`.

## How Has This Been Tested?

I didn't change any example, as the `prefix_separator` is already documenting in the https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/UPGRADE-18.0.md guide and not shown under examples.


gferon:
> Thank you for the PR but this change will not be propagated to node groups. Node groups are meant to be replaced - the changes were implemented to protect the control plane from replacement only

I understand, but I feel like having node groups replaced as part of a Terraform module upgrade is asking for a disaster. I don't see how one can expect an actual change of infrastructure while only performing such change. I guess we'll keep our fork alive for now, and will move back to using the upstream version when provisioning managed node groups.

bryantbiggs:
Yes, unfortunately with the current limitations of Terraform, handling breaking code changes are particularly painful

sgametrio:
@bryantbiggs We are struggling with the update because of the forced renaming of the resources. This PR would allow us to have no downtime easily without spending a lot of effort on understanding whether each resource re-creation is critical or not.

Would you consider it? It's a very small change which still respect your new naming convention by default, but being able to configure the prefix_separator for self_managed_node_groups is the only easy way forward.

I'm sorry to say this and I know we're benefitting for free from your work (and we're grateful for it), but with this breaking change I guess I will gradually move away from the module (or fork it and maintain my own one) because it makes my life really hard to migrate with no downtime.

PR #2076, opened 2022-05-16
feat: Add AmazonEKSServicePolicy to clusters for compatibility

TheRealDwright:
https://docs.aws.amazon.com/eks/latest/userguide/security-iam-awsmanpol.html

You can attach AmazonEKSServicePolicy to your IAM entities. Clusters that
were created before April 16, 2020, required you to create an IAM role
and attach this policy to it. Clusters that were created on or after
April 16, 2020, don't require you to create a role and don't require
you to assign this policy.

## Motivation and Context
Provide backwards compatibility for EKS clusters created prior to April 16, 2020

Without this change, older clusters will not be able to stream logs to Cloudwatch
among other lost functionality.

## Breaking Changes
N/A

## How Has This Been Tested?
Tested on our EKS cluster

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -256,6 +256,7 @@ resource "aws_iam_role" "this" {
 resource "aws_iam_role_policy_attachment" "this" {
   for_each = local.create_iam_role ? toset(compact(distinct(concat([
     "${local.policy_arn_prefix}/AmazonEKSClusterPolicy",
+    "${local.policy_arn_prefix}/AmazonEKSServicePolicy",
     "${local.policy_arn_prefix}/AmazonEKSVPCResourceController",
   ], var.iam_role_additional_policies)))) : toset([])
 


bryantbiggs:
Thank you for the PR! However, I think this change is only required for a very small segment of users and  we don't want to enforce this on all users who do not require this additional permission. Instead of adding this to the module, I think the best course of action would be for those users who require the additional managed policy to add it using `iam_role_additional_policies` like:

```hcl
	...
	iam_role_additional_policies = ["arn:aws:iam::aws:policy/AmazonEKSServicePolicy"]
	...

TheRealDwright:
> I think this change is only required for a very small segment of users and we don't want to enforce this on all users

My thinking is that it provides compatibility between both cluster models and definitely caused us some troubleshooting headaches after upgrading the module. However, if you feel it's not the direction you wish to go, feel free to close.

bryantbiggs:
Again, thank you for the PR but this appears to affect a very small segment of our user base and can be easily resolved by attaching this through the provided `iam_role_additional_policies` variable. We don't want to inflict this additional permissions on all users so unfortunately we won't be accepting this change

PR #2077, opened 2022-05-16
Set ami_type to CUSTOM when providing ami_id

gferon:
## Description

This fix lets you change the `ami_type` of a node-group from managed to `CUSTOM`.

## Motivation and Context

Even though this is an immutable parameter, the provider will destroy and recreate the node-group which is the desired behavior.

## How Has This Been Tested?
- [x] Tested on existing cluster, by switching a node-group from `AL2_x86-64` to `CUSTOM`


gferon:
> Thank you for the PR! However, users can use the CUSTOM AMI type today by simply setting:

That's not true with my current understanding: when you set _both_ `ami_id` and `ami_type`, then `ami_type` is set to an empty string, which means the provider will not mark the resource as "to be replaced", and the API request will fail.

bryantbiggs:
I am just following what is stated in the documentation that I linked:
image

PR #2083, opened 2022-05-20
fix: Bugfix kubernetes configmap update

matisilva:
## Description
Just Bugfixing an issue with kubernetes provider

## Motivation and Context
I need to add auth role for admin control 

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -411,6 +411,16 @@ locals {
   }
 }
 
+provider "kubernetes" {
+  host                   = aws_eks_cluster.this[0].endpoint
+  cluster_ca_certificate = base64decode(aws_eks_cluster.this[0].certificate_authority[0].data)
+  exec {
+    api_version = "client.authentication.k8s.io/v1alpha1"
+    args        = ["eks", "get-token", "--cluster-name", aws_eks_cluster.this[0].name]
+    command     = "aws"
+  }
+}
+
 resource "kubernetes_config_map" "aws_auth" {
   count = var.create && var.create_aws_auth_configmap ? 1 : 0
 


bryantbiggs:
Thank you for the PR @matisilva - however, we do not embed providers within the modules. Providers are left up to users to configure

matisilva:
@bryantbiggs you are currently using that provider in an implicit way.  Adding this piece of code lets you use the resource. Otherwise the system tries to use the kubectl config in your system that would be not the same as the one needed for the cluster created.  It is needed to reconfigure the provider with the newly created cluster. Please re consider the fix otherwise the aws-auth customization is going to fail.

bryantbiggs:
@matisilva that is how providers are handled in Terraform https://www.terraform.io/language/providers/configuration#provider-configuration-1

matisilva:
Yeap but the porblem here is that you are using a composite module  with two steps
1) eks creation
2) kubernetes role customization

For 2) you need to reconfigure your provider with the values of the recently created cluster. 
There is no other way to make this work.  I cant setup the provider with future values of the eks cluster nor either leave it blank (it wont be changing role of the respective cluster it is going to be using my local config) Please re consider this deply or give me a way to  asign an extra role to manage the cluster from code..

matisilva:
https://stackoverflow.com/questions/69655605/configmaps-aws-auth-not-found/72313033#72313033 Lot of people is having the same issue..  If you have something in mind is ok for me but make it explcit how it should work

bryantbiggs:
Have you looked at the examples provided in the repo?

matisilva:
Ok let me try that way

matisilva:
My apologies for the misunderstanding. You were right. Thanks for the answers.!

PR #2101, opened 2022-06-04
docs: Mention that security groups must be manually referenced on externally attached EKS managed groups

lucasff:
## Description
Improving the docs a little bit, I was bitten hard by the lack of guidance on those two variables.

## Motivation and Context
Gives the information to anyone that tries to decouple the main `eks` module to different managed or self-managed groups.

## Breaking Changes
None. Docs only.

### DIFF
--- a/modules/eks-managed-node-group/README.md
+++ b/modules/eks-managed-node-group/README.md
@@ -15,6 +15,11 @@ module "eks_managed_node_group" {
   vpc_id     = "vpc-1234556abcdef"
   subnet_ids = ["subnet-abcde012", "subnet-bcde012a", "subnet-fghi345a"]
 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
isn't this covered by the example already? https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/complete/main.tf#L234-L290

lucasff:
@bryantbiggs yes but it's not clear that is mandatory. more specifically, a complete example doesn't tell what's required and what's optional. I'm changing the non-complete example, because one would assume the Managed Group would well, manage itself.
Extra docs never hurt.

bryantbiggs (approved):

antonbabenko:
This PR is included in version 18.24.0 :tada:

PR #2119, opened 2022-06-23
docs: Added note explaining requirement for kubernetes provider.

rmkellogg:
## Description
Document the need for kubernetes provider in this example.

## Motivation and Context

The inclusion of the aws_auth_users resource in this example requires this and manual addition of this 
resource can lead to confusing errors tracking down the root cause.

Related to issue #852.

## Breaking Changes
None

## How Has This Been Tested?
Documentation only.

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -8,6 +8,8 @@ provider "aws" {
   }
 }
 
+# The Kubernetes provider is included in this file so the EKS module can complete successfully. Otherwise, it throws an error when 
+# creating `kubernetes_config_map.aws_auth`.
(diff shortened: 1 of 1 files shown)


rmkellogg:
Agreed but I spent a better part of a day without the full snippet of the Kubernetes provider configuration in the complete example.  Thought that a note explaining why it is needed would be helpful.

rmkellogg:
Entirely your call.  The text I added was originally in another sample I believe came from here but could be mistaken.  Feel free to close the PR if you need not necessary or inconsistent.  Thanks.

PR #2121, opened 2022-06-26
feat: Add support for creating KMS key for cluster secret encryption

bryantbiggs:
## Description
- Add support for creating KMS key for cluster secret encryption

## Motivation and Context
- Adding the cluster IAM role to a KMS key policy creates a cyclical dependency when the key/policy are created externally unless users also create the role externally. Provisioning a cluster with secrets encrypted using KMS requires the KMS key to be created first. Creating the KMS key first with a key policy causes a conflict because the cluster IAM role has not been created yet and cannot be referenced. The only resolution currently is to either create the key, key policy, and role externally to ensure proper ordering and name resolution, or apply secret encryption AFTER the cluster and cluster IAM role have been created (poor user experience). Embedding the key and key policy creation into the module resolves this scenario while allowing users to extend the policy as well

## Breaking Changes
- No

## How Has This Been Tested?


antonbabenko (approved):
Lgtm. I was thinking for some time while reading this if we should prefix variables with 'key_' or 'kms_key_' to avoid potential confusion with "private key". Current naming is good regardless :)

bryantbiggs:
> Lgtm. I was thinking for some time while reading this if we should prefix variables with 'key_' or 'kms_key_' to avoid potential confusion with "private key". Current naming is good regardless :)

thats a good point, I've updated to include `kms_` to make it more clear

antonbabenko:
This PR is included in version 18.25.0 :tada:

PR #2133, opened 2022-06-30
fix: Correct variable types to improve dynamic check correctness

### LINKED ISSUE #2132: Fix: Allow setting metadata endpoint options so users can configure or disable
Can we get this fix back-ported to this repo: https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/657 ? It would be useful for the same reasons it was implemented for the AWS EKS Blueprints repo.

Allow setting metadata endpoint options so users can configure or disable.

**Motivation**

Allow users to configure the metadata endpoint options to suit their workloads, or even disable if not supported in their region

bryantbiggs:
## Description
- Correct variable types to improve dynamic check correctness. Comparing against `null` doesn't appear to work as expected, instead we can compare against number of elements 

## Motivation and Context
- On current module version, if users want to disable `metadata_options`, the initial approach would be to set `metadata_options` = `null` but this still results in metadata options being set, possibly due to [this](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L303) series within the `try()` block. However, if we instead change the type to be a default of `{}` and check how many elements are within the map, we can now disable the creation of `metadata_options` by setting it equal to `{}` (has to be explicitly set to `{}` due to default provided from security perspective https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L2) 
- Closes #2132

## Breaking Changes
- No; variable types are same as they are today, its just the default value has changed to not be `null` and the conditional logic updated to support

## How Has This Been Tested?


bryantbiggs:
guess the buttons on mobile don't do the same thing (or what I thought) 😅

antonbabenko (approved):
Looks logical to me. :)

antonbabenko:
This PR is included in version 18.26.2 :tada:

morganrowse:
This (`18.26.1` -> `18.26.2`) broke our Terraform. Definitely a breaking change.

PR #2140, opened 2022-07-05
fix: Prevent 'coalescelist' from failing

bmcustodio:
## Description

In some circumstances, `apply` may fail with the following error:

    Error: Error in function call

      on .terraform/modules/eks.main/main.tf line 21, in resource "aws_eks_cluster" "this":
      21:     subnet_ids              = coalescelist(var.control_plane_subnet_ids, var.subnet_ids)
        ├────────────────
        │ var.control_plane_subnet_ids is empty list of string
        │ var.subnet_ids is empty list of string

    Call to function "coalescelist" failed: no non-null arguments.

This commit adds a list with an empty string to the arguments of
`coalescelist` in order to prevent Terraform from failing.

## Motivation and Context

See above.

## Breaking Changes

N/A

## How Has This Been Tested?

- [x] I have tested using a real project that depends on this module

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -21,7 +21,7 @@ resource "aws_eks_cluster" "this" {
 
   vpc_config {
     security_group_ids      = compact(distinct(concat(var.cluster_additional_security_group_ids, [local.cluster_security_group_id])))
-    subnet_ids              = coalescelist(var.control_plane_subnet_ids, var.subnet_ids)
+    subnet_ids              = coalescelist(var.control_plane_subnet_ids, var.subnet_ids, [""])
     endpoint_private_access = var.cluster_endpoint_private_access
     endpoint_public_access  = var.cluster_endpoint_public_access
     public_access_cidrs     = var.cluster_endpoint_public_access_cidrs


bryantbiggs:
both `vpc_config` and `subnet_ids` are required https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#subnet_ids

so I don't see this as a valid change

bmcustodio:
> so I don't see this as a valid change

Can you comment on the error, then, and propose a different fix / workaround?

bryantbiggs:
you haven't provided a value for `subnet_ids` or `control_plane_subnet_ids` which is causing the error

bryantbiggs:
closing with the information provided above

bmcustodio:
> you haven't provided a value for `subnet_ids` or `control_plane_subnet_ids` which is causing the error

I did provide a value for `subnet_ids` and `control_plane_subnets_ids` — they come from a data source that filters for private subnets, and there are at least three private subnets (created by https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest).

> closing with the information provided above

Not really any helpful, and I believe there's still an issue lying around, but that's OK I guess — we'll find another way. Thanks for your time.

bryantbiggs:
> I did provide a value for subnet_ids and control_plane_subnets_ids — they come from a data source that filters for private subnets, and there are at least three private subnets (created by [registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest)).

Without seeing any configurations or reproductions, how else can I provide any help? It sounds like your data sources are not returning any subnets which leads to this error you are seeing - other users are not reporting this error and if there was truly an error, we would see a lot of issues due to how much this module is utilized

PR #2204, opened 2022-08-16
add extra cluster_enabled_log_types

FernandoMiguel:
## Description
add extra cluster_enabled_log_types

## Motivation and Context
sync up from blueprints
https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/b9d7acce/variables.tf#L177

## How Has This Been Tested?


bryantbiggs:
this is left up to users to opt in to additional log types if they chose, but I don't think we'll be adding all of them to be enabled by default

FernandoMiguel:
> this is left up to users to opt in to additional log types if they chose, but I don't think we'll be adding all of them to be enabled by default

interesting, given that they are all enabled by default in blueprints, and not here.
why would you say the difference in approach and recommended default?

bryantbiggs:
the defaults that have been set here are set from a security perspective, I do not know the origins of the defaults for EKS Blueprints nor why it defaulted to all logging types being enabled by default

PR #2174, opened 2022-07-22
fix: Pin TLS provider version to 3.x versions only

bryantbiggs:
## Description
- Pin TLS provider version to 3.x versions only
- Correct KMS key creation boolean logic

## Motivation and Context
- Ref https://github.com/hashicorp/terraform-provider-tls/issues/244
- Closes #2170 
- Closes #2171
- Closes #2172
- Closes #2173 

## Breaking Changes
- No

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #2211, opened 2022-08-25
feat: Allow TLS provider to use versions 3.0+ (i.e. - `>= 3.0`)

KarstenSiemer:
## Description
In #2174 the tls provider was pinned to `~> 3.0` because of https://github.com/hashicorp/terraform-provider-tls/issues/244,
of which the corresponding problem was fixed in 4.0.1 of the provider according to the maintainer, which I can confirm.

This PR will upgrade the tls provider to the version which includes the fix.

## Motivation and Context

This is required, since we should look forward to being as up to date as possible and to prohibit any provider errors, because
a lot of modules have already upgraded to 4x.

## Breaking Changes
None, that I can think of.

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ We are grateful to the community for contributing bugfixes and improvements! Ple
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.13.1 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.72 |
 | <a name="requirement_kubernetes"></a> [kubernetes](#requirement\_kubernetes) | >= 2.10 |
-| <a name="requirement_tls"></a> [tls](#requirement\_tls) | ~> 3.0 |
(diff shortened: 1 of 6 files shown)


bryantbiggs:
if we enforce 4.x as the minimum version, then that falls under a breaking change. but if we can change `~> 3.0` to `>= 4.0` then that we could support today without a breaking change

KarstenSiemer:
@bryantbiggs thanks for the suggestion. I have updated the commit

bryantbiggs:
apologies - I made a typo, see the updated comment above

KarstenSiemer:
@bryantbiggs no, worries. I was a little confused but decided to trust 😂 
Hope it is fine now

KarstenSiemer:
Do you think we can merge this soon and publish a new version?

bryantbiggs (approved):
Thank you!

antonbabenko:
This PR is included in version 18.29.0 :tada:

PR #2216, opened 2022-09-01
feat: Add ability to use IAM groups in the aws-auth configmap

floppyfish14:
## Description
Updated variables.tf and and main.tf to add the variable and use it.

## Motivation and Context
This change allows the use of IAM groups for authentication with eks cluster

## Breaking Changes
Unsure; I would test this if I knew how

## How Has This Been Tested?
Ran a `terraform init` and a `terraform validate` locally after editing.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -445,6 +445,7 @@ locals {
       ],
       var.aws_auth_roles
     ))
+    mapGroups   = yamlencode(var.aws_auth_groups)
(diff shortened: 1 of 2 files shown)


bryantbiggs:
This is not a valid attribute that is supported by the aws-auth configmap

floppyfish14:
Good to know. Should have looked that up first.

Issue #2217, opened 2022-09-05
module overrides load_balancers configuration for self managed node group ASGs


arpanadhikari:
## Description

Since the new module version v18.x doesn't support attaching classic load balancers to ASGs. We have to separately attach load balancers using the `aws_autoscaling_attachment` terraform resource.

However, for terraform-aws-eks managed ASGs, every time we run a `terraform apply`, the module removes this association leading to the load balancers being removed from ASGs, causing an outage.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note
This is relevant for self managed node group ASGs.

## Versions

- Module version [Required]:

- Terraform version:
<1.1.3
- Provider version(s):
v18.18.x

## Reproduction Code [Required]
Create a new alb_attachment.tf
```hcl
resource "aws_autoscaling_attachment" "test" {
  count                                  = length(module.eks.self_managed_node_groups_autoscaling_group_names)
  autoscaling_group_name                 = module.eks.self_managed_node_groups_autoscaling_group_names[count.index]
  elb                                    = "elb-arn"
}
```

- Add the above  configuration to your eks module usage.
- Run terraform apply.
- Run terraform plan.
- Notice the module try to remove the previously created ELB->ASG association.
- 

Using Workspaces

Cache cleared.

See steps above for reproducing this issue.

## Expected behavior

Module should ignore changes to ELB associations with ASG since it's no longer managing this part of the configuration.

## Actual behavior

Module overrides externally added config for ELB-> ASG association.

arpanadhikari:
Created a PR to resolve this issue.
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2218

bryantbiggs:
I will post the question I left on the PR here as well - I'm curious, do you not utilize the AWS load balancer controller for ingress?

arpanadhikari:
> I will post the question I left on the PR here as well - I'm curious, do you not utilize the AWS load balancer controller for ingress?

We do. But some of the clusters do not have load balancer controllers so we have to attach long-lived load balancers deployed separately using terraform.

bryantbiggs:
see comment here https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2218#issuecomment-1238650527

PR #2218, opened 2022-09-05
fix: Ignore load_balancers, since module doesn't manage it

arpanadhikari:
## Description
See #2217

Since the new module version v18.x doesn't support attaching classic load balancers to ASGs. We have to separately attach load balancers using the aws_autoscaling_attachment terraform resource.

However, for terraform-aws-eks managed ASGs, every time we run a terraform apply, the module removes this association leading to the load balancers being removed from ASGs, causing an outage.

## Motivation and Context
See #2217

## How Has This Been Tested?
Change can be tested by using the following resource in modules configured with self-managed-eks-node-groups:

```hcl
resource "aws_autoscaling_attachment" "test" {
  count                                  = length(module.eks.self_managed_node_groups_autoscaling_group_names)
  autoscaling_group_name                 = module.eks.self_managed_node_groups_autoscaling_group_names[count.index]
  elb                                    = "elb-arn"
}
```

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -417,7 +417,8 @@ resource "aws_autoscaling_group" "this" {
   lifecycle {
     create_before_destroy = true
     ignore_changes = [
-      desired_capacity
+      desired_capacity,
+      load_balancers
     ]
   }
 }


bryantbiggs:
I'm curious, do you not utilize the AWS load balancer controller for ingress?

bryantbiggs:
we really strive to avoid these ignored changes unless there is a valid reason. This is the first time I am seeing it come up on this repo in the past 1-1.5 years so I'm inclined to say its a niche case and will be erring on the side of caution at this time by not adding it here

arpanadhikari:
> we really strive to avoid these ignored changes unless there is a valid reason. This is the first time I am seeing it come up on this repo in the past 1-1.5 years so I'm inclined to say its a niche case and will be erring on the side of caution at this time by not adding it here

Fair argument. Would you be open to making it optional? By adding a parameter to the module?

bryantbiggs:
Unfortunately, Terraform doesn't support variables within lifecycle blocks

PR #2220, opened 2022-09-07
chore: Correct example ASG tag(s) to match `cluster-autoscaler` requirements

noony:
## Description
Example for asg tagging to scale to 0 for managed node group doesn't work

## Motivation and Context
In the asg taint tags example the tags generated are like this : 
```
key : k8s.io/cluster-autoscaler/node-template/taint/nameOfTheTaint
value: true:NO_SCHEDULE
```

But [cluster autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup) requires another format : 
```
key : k8s.io/cluster-autoscaler/node-template/taint/nameOfTheTaint
value: true:NoSchedule
```

## Breaking Changes
No

## How Has This Been Tested?

### DIFF
--- a/examples/eks_managed_node_group/main.tf
+++ b/examples/eks_managed_node_group/main.tf
@@ -675,6 +675,14 @@ data "aws_ami" "eks_default_bottlerocket" {
 ################################################################################
 
 locals {
+
+  # We need to lookup K8s taint effect from the AWS API value
+  taint_effects = {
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
thanks - technically this is outside the scope of the module and was added by another user, but its valid to update to at least show something correct

PR #2229, opened 2022-09-12
feat: Allow changing instance name tag key

### LINKED ISSUE #2188: Create EC2 instances without "Name" tags
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅. It's about functionality of _this_ module (terraform-aws-eks).

Yes. I'm trying to convince EKS module to not create ASG/Launch Template configuration with Name tags propagated to instances. 

A boolean variable `propagate_name_tags_to_instances` (or equivalent), defaulting to `true` that would allow to disable this functionality.

## Additional context

My main issue is to get rid of additional `name` tags in Datadog metrics that emit their own (metric-specific) `name` tags (or just any metric). We use EC2 tags to Datadog tags propagation and we like it, but "name" is such a generic name, it doesn't mean anything (unless taken in context of a specific metric). Duplicated tags prevent us from creating Events-based monitoring. And the Name tags are useful only for browsing through AWS Console (UI) which is not often for people using Terraform (us). And nodes in question are not named in a unique way – they're named after their parent ASG. We can live without these tags easily if it fixes any of our issues.

I've considered several alternatives.

### Fix Datadog agent

The obvious solution is to make Datadog agents ignore that one EC2 instance tag, but there is no such option. (There is such option for GCE, though). I'm considering reporting this to Datadog as well.

### Override tag with tag values "null"

There is an option to remove Launch Template tags. Due to `merge` used in `eks` code, we could simply:
```
  launch_template_tags = {
    Name  = null
  }
```

And then Terraform removes null tags from Launch Templates.

But it can't be done for ASG `propagate_at_launch` tags, so instances are still Name-tagged.

### Override with `autoscaling_group_tags`

`autoscaling_group_tags` node group property allows to set up tags not propagated to instances. There is no merge, each tag is a `tag { ... }` schema. It looks like it Terraform doesn't sort these things internally or apply it in order – the results vary between different autoscaling groups. For some autoscaling groups the propagated tag wins, for others – the not propagated one. (It doesn't change for a single ASG when applied multiple times though).

moleskin-smile:
## Description

Allow specifying alternative key for the instance name tag ("Name" by default).

## Motivation and Context

Allow overriding the only hardcoded tag that couldn't be changed otherwise. The use case is described in #2188 in detail.

Fixes #2188.

## Breaking Changes

No breaking changes. Everything works the same if variable is not specified.

## How Has This Been Tested?


moleskin-smile:
One problem we could cause with this is this Terraform issue: https://github.com/hashicorp/terraform/issues/21566. Do we support Terraform versions released more than 2 years ago here? If so, I should probably change the syntax to interpolation `(var.instance_name_tag_key)` => `"${var.instance_name_tag_key}"`.

bryantbiggs:
hi @moleskin-smile - thanks for the PR but we will not be making this change at this time. It is very common to set the `Name` tag and this is not something that has been raised as a concern elsewhere so we will be keeping things as is. Thank you!

PR #2241, opened 2022-09-26
fix: add more social issues as to allow for fair representation

tibers:
## Description
We have a problem with code-as-activism that it frequently doesn't fairly represent minority and underprivileged populations. By focusing exclusively on one facet of the overwhelming struggle against the collective oppression of humanity, we risk losing perspectives which will ultimately help us all.

## Motivation and Context
Code-as-activism can be so much more than one single issue. We should strive to enumerate as many social issues as we can to help bring their attention to AWS customers.

## Breaking Changes
The only thing it breaks is code for fascists.

## How Has This Been Tested?


bryantbiggs:
thank you for the PR, but at this time we won't be making any further changes related to cluster conditional creation

PR #2245, opened 2022-09-27
fix: Properly doc and exemplify max-pods settings

bryanasdev000:
## Description

In the latest AWS AMI versions we need to make sure to overwrite kubelet invocation as it does not use its config file for checking max pods and instead rely on cli args to set the desired number of max pods.

This is related to https://github.com/awslabs/amazon-eks-ami/issues/844 and proper credits should be given to https://github.com/awslabs/amazon-eks-ami/issues/844#issuecomment-1048592041 (@lobeck).

Without this line the default number of max pods its still used from https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt or properly configured if using AWS CNI Prefix Delegation (https://github.com/awslabs/amazon-eks-ami/compare/v20220824...v20220914#diff-39737df68b6d07b99c2e837c32b4a0cceaf68bacfe968606694b9d6242ca91aeR10).

I am using this in production with clusters ranging from 1.21 to 1.23 without issues even on new AMI optimized releases with Cilium.

## Motivation and Context

To bypass AWS EKS config and set desired number of pods, assuming a custom CNI scenario.

## Breaking Changes

N/A

## How Has This Been Tested?

### DIFF
--- a/docs/user_data.md
+++ b/docs/user_data.md
@@ -64,6 +64,8 @@ Since the EKS Managed Node Group service provides the necessary bootstrap user d
   EOF
   # Source extra environment variables in bootstrap script
   sed -i '/^set -o errexit/a\\nsource /etc/profile.d/bootstrap.sh' /etc/eks/bootstrap.sh
+  # In case of the desired change is set via CLI args (like max-pods)
+  sed -i 's/KUBELET_EXTRA_ARGS=$2/KUBELET_EXTRA_ARGS="$2 $KUBELET_EXTRA_ARGS"/' /etc/eks/bootstrap.sh
   EOT
 ```
 
--- a/examples/eks_managed_node_group/main.tf
+++ b/examples/eks_managed_node_group/main.tf
@@ -228,6 +228,8 @@ module "eks" {
       EOF
       # Source extra environment variables in bootstrap script
       sed -i '/^set -o errexit/a\\nsource /etc/profile.d/bootstrap.sh' /etc/eks/bootstrap.sh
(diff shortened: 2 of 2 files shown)


bryantbiggs:
thank you for the PR but I am lost as to what we are trying to do here? we do provide some examples of user data settings for things like containerd, but we are not looking to be the source of truth on configuring the kubelet

bryanasdev000:
> thank you for the PR but I am lost as to what we are trying to do here? we do provide some examples of user data settings for things like containerd, but we are not looking to be the source of truth on configuring the kubelet

Indeed, but for now this module is the sanest way to configure max pods on EKS, which is extremely useful in a context we use another CNI like Cilium, Calico and others at AWS EKS.

In the documentation and examples a reference for setting max pods can be observed [like this](https://github.com/terraform-aws-modules/terraform-aws-eks/search?q=max+pods), so the idea here is to provide a working example on how to really use max pods and a "tip" that others options may need to override de eks boostrap script to properly handle it (and as I have done this in a hurry, I think that it deserves a better note).

Also, searching for it now, I think that this PR need to add this line to a few other files, so I am moving this to draft.

I know that this fall under a non-supported behavior as is a "problem" with AWS EKS, but I think that can be beneficial for the current user base for this module as using another CNI is not that uncommon.

bryanasdev000:
Not stale, will push soon.

bryantbiggs:
closing for now due to lack of updates for some time. we can revisit in the future if this is warranted

Issue #2246, opened 2022-09-28
Custom AMI: Pre-bootstrap user_data is inserted into launch templates of node groups when enable_bootstrap_user_data is disabled


bogdando:
## Description

There is also a misconfiguration in ``templates/linux_user_data.tpl`` which always inserts ``pre_bootstrap_user_data``, even if ``enable_bootstrap_user_data: false``. The latter setting should control insertion of ``pre_bootstrap_user_data`` just like it already does for ``post_bootstrap_user_data``.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: v18.29.1

- Terraform version:

Terraform v1.2.5
on linux_amd64

- Provider version(s):

+ provider registry.terraform.io/gavinbunney/kubectl v1.14.0
+ provider registry.terraform.io/hashicorp/aws v4.32.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.12.1
## Reproduction Code [Required]

```
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 4.32"
    }
  }
  required_version = ">= 0.13.1"
}

provider "aws" {
  region  = "eu-west-1"
}

data "aws_caller_identity" "current" {}

locals {
  name            = "test-1"
  cluster_version = "1.23"
  region          = "eu-west-1"
  azs             = ["eu-west-1c"]
}

module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "3.12.0"

  name = local.name
  cidr = "10.20.0.0/16"

  azs             = local.azs
  private_subnets = ["10.20.36.0/22"]

  enable_nat_gateway     = false
}

resource "tls_private_key" "this" {
  algorithm = "RSA"
}

resource "aws_key_pair" "this" {
  key_name_prefix = local.name
  public_key      = tls_private_key.this.public_key_openssh
}

module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "18.29.1"

  cluster_name                    = local.name
  cluster_version                 = local.cluster_version
[...]

Issue #2147, opened 2022-07-06
Custom AMI: node groups fail updating user_data because it is created not MIME formatted


bogdando:
## Description

tl;dr Amazon EC2 user data in launch templates that are used with EKS managed node groups must be in the MIME multi-part archive format. But setting the ``enable_bootstrap_user_data: true`` flag always creates it unformatted.

There is also a misconfiguration in ``templates/linux_user_data.tpl`` which always inserts ``pre_bootstrap_user_data``, even if ``enable_bootstrap_user_data: false``. The latter setting should control insertion of ``pre_bootstrap_user_data`` just like it already does for ``post_bootstrap_user_data``.

Now the longer version.
When deploying EKS node groups from launch templates with custom AMI types (platform "linux") and either EKS unmanaged, or with cloud-init user data provided, that bootstrap data will be pushed to AWS APIs as is, not MIME formatted. That is because the MIME-wrapping module [cloudinit_config" "linux_eks_managed_node_group"](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/_user_data/main.tf#L66) will NOT be called in such scenarios.

Flags ``is_eks_managed_node_group: false`` and/or ``enable_bootstrap_user_data : true`` disable the module triggering condition (is expected for the former perhaps, but not for the latter!) ``var.create && var.platform == "linux" && var.is_eks_managed_node_group && !var.enable_bootstrap_user_data && var.pre_bootstrap_user_data != "" && var.user_data_template_path == ""``.

That is a problem, since AWS API accepts that not-MIME formatted user data on create request, and fails later, when updating that node group's launch template (Ec2LaunchTemplateInvalidConfiguration: User data was not in the MIME multipart format).

Related https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1875
Related https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2059
Related https://github.com/hashicorp/terraform-provider-aws/issues/15007

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: v18.23.0

- Terraform version:

```
[...]

Issue #2059, opened 2022-05-07
eks_managed_node_groups: bootstrap_extra_args: "User data was not in the MIME multipart format"


qingvincentyin:
## Description

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

See:

- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1875

I don't think the underlying problem was resolved and might have gotten worse in `terraform-aws-modules` `v18.20.5` (or probably more likely because of `hashicorp/aws` provider `4.x.x`). I can reproduce the error in a very simple 1-step procedure below.

## Versions

Module version:

- https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v18.20.5

```console
$ tf version
Terraform v1.1.1
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v4.13.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.11.0
+ provider registry.terraform.io/hashicorp/tls v3.3.0


$ tf providers

Providers required by configuration:
.
├── provider[registry.terraform.io/hashicorp/aws] ~> 4.11
└── module.eks
    ├── provider[registry.terraform.io/hashicorp/tls] >= 3.0.0
    ├── provider[registry.terraform.io/hashicorp/kubernetes] >= 2.10.0
    ├── provider[registry.terraform.io/hashicorp/aws] >= 3.72.0
    ├── module.eks_managed_node_group
    │   ├── provider[registry.terraform.io/hashicorp/aws] >= 3.72.0
    │   └── module.user_data
    │       └── provider[registry.terraform.io/hashicorp/cloudinit] >= 2.0.0
    ├── module.fargate_profile
    │   └── provider[registry.terraform.io/hashicorp/aws] >= 3.72.0
    └── module.self_managed_node_group
        ├── provider[registry.terraform.io/hashicorp/aws] >= 3.72.0
        └── module.user_data
            └── provider[registry.terraform.io/hashicorp/cloudinit] >= 2.0.0
```

## Reproduction Code

Here's the complete minimal code to reproduce the problem.

```hcl
module "eks" {
  source          = "terraform-aws-modules/eks/aws"
  version         = "18.20.5"
  cluster_name    = "my-cluster1"
  cluster_version = "1.21"
  vpc_id          = "vpc-8333e1fb"
[...]

Issue #1875, opened 2022-02-14
Deploying managed node group with default user_data template fails - Ec2LaunchTemplateInvalidConfiguration: User data was not in the MIME multipart format.


idan-gur:
## Description

While trying to deploy a completely [private EKS cluster](https://docs.aws.amazon.com/eks/latest/userguide/private-clusters.html),
when adding the `enable_bootstrap_user_data = true` option to add the necessary flags to the bootstrap.sh script the apply fails.

## Versions

 - Terraform: 1.1.5

 - Provider(s): AWS 3.71.0

 - Module: 18.x

## Reproduction

- Create a managed node group EKS (platform linux) 
- Add the `enable_bootstrap_user_data = true` flag
- Run plan

### Code Snippet to Reproduce
```
eks_managed_node_groups = {
	  main_node_group = {
		    name = "***"
		    
		    enable_monitoring      = false
		    create_launch_template = true
		    ebs_optimized          = true
		    desired_capacity       = 1
		    max_capacity           = 1
		    min_capacity           = 1
		    
		    vpc_security_group_ids  = []
		    
		    disk_size       = 100
		    
		    instance_types = ["r5.xlarge"]
		    ami_type       = "AL2_x86_64" # "BOTTLEROCKET_ARM_64" "AL2_ARM_64"
		    public_ip      = false
		    
		    # Bootstrap data
		    enable_bootstrap_user_data = true
	}
}
```

## Expected behavior
Apply ran successfully

## Actual behavior
```
╷
│ Error: error waiting for EKS Node Group (***) to create: unexpected state 'CREATE_FAILED', wanted target 'ACTIVE'. last error: 1 error occurred:
│ 	* : Ec2LaunchTemplateInvalidConfiguration: User data was not in the MIME multipart format.
│ 
│ 
│ 
│   with module.eks.module.eks.module.eks_managed_node_group["***"].aws_eks_node_group.this[0],
│   on .terraform/modules/eks.eks/modules/eks-managed-node-group/main.tf line 269, in resource "aws_eks_node_group" "this":
│  269: resource "aws_eks_node_group" "this" {
│ 
╵
```

## Additional context
I think this is due to the _user_data module not using cloudinit_config when the `enable_bootstrap_user_data` is set.

bryantbiggs:
the MIME multipart is only required when using EKS managed node groups and NOT supplying a custom AMI (i.e. - no value provided for `ami_id`)

[...]

PR #2247, opened 2022-09-28
fix: Use MIME format for user data

bogdando:
## Description

When ``enable_bootstrap_user_data`` is set, Amazon EC2 user data in launch templates that are used with managed node groups must be in the MIME multi-part archive format. This is because user data is merged with Amazon EKS user data required for nodes to join the cluster.

When deploying EKS-managed node groups, make user data always MIME formatted, whenever a user supplied template, or the module supplied one is in use. Properly render the whole template contents instead of ``pre_bootstrap_user_data`` only.

Also, fix ``pre_/post_ bootstrap_user_data`` injection into the module provided template to be properly controled by the
``enable_bootstrap_user_data`` flag.

## Motivation and Context

Launch templates for EKS-managed (but not self-managed) node groups and custom user data will become MIME formatted.

Resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2147
Resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2246
Related https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1875
Related https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2059
Related https://github.com/hashicorp/terraform-provider-aws/issues/15007

## How Has This Been Tested?

### DIFF
--- a/examples/user_data/main.tf
+++ b/examples/user_data/main.tf
@@ -26,6 +26,9 @@ module "eks_mng_linux_additional" {
 module "eks_mng_linux_custom_ami" {
   source = "../../modules/_user_data"
 
+  # pre_/post_ bootstrap_user_data will be defined in MIME format
+  platform = "linux"
+
   cluster_name              = local.name
   cluster_endpoint          = local.cluster_endpoint
   cluster_auth_base64       = local.cluster_auth_base64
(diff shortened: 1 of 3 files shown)


bogdando:
@bryantbiggs PTAL. Now it has two distinct issues linked, each has a working reproducer in its description

bryantbiggs:
see https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2246#issuecomment-1260782090

bogdando:
There is two issues what this change resolves, both look valid as there are reproducers for each of it.
Ignoring the problem and rejecting a fix is not the best approach for such a great and widely used community module.

PR #2261, opened 2022-10-10
Added support for network_card_index

m4ce:
## Description
network_card_index not supported in eks submodule for managed/self-managed node groups

## Motivation and Context
the network card index is set to 0 on all nics in the launch template

## Breaking Changes
None that I know of.


antonbabenko:
This issue has been resolved in version 19.0.0 :tada:

PR #2264, opened 2022-10-11
fix: Invalid for_each argument with var.create_cluster_primary_security_group_tags = false

dex4er:
I've got today:

```
╷          
│ 
│   on .terraform/modules/app_cluster.eks/main.tf line 70, in resource "aws_ec2_tag" "cluster_primary_security_group":
│   70:   for_each = { for k, v in merge(var.tags, var.cluster_tags, data.aws_default_tags.current.tags) :
│   71:     k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags
│   72:   }
│     ├────────────────
│     │ data.aws_default_tags.current.tags is a map of string, known only after apply
│     │ local.create is true
│     │ var.cluster_tags is empty map of string
│     │ var.create_cluster_primary_security_group_tags is false
│     │ var.tags is map of string with 4 elements
│ 
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot
│ determine the full set of keys that will identify the instances of this resource.
│ 
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place
│ apply-time results only in the map values.
│ 
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends on, and
│ then apply a second time to fully converge.
╵
```

but I have `var.create_cluster_primary_security_group_tags = false`

It worked with this patch.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -68,9 +68,9 @@ resource "aws_ec2_tag" "cluster_primary_security_group" {
   # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2006
   # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2008
   # `aws_default_tags` is merged in to "dedupe" tags and stabilize tag updates
-  for_each = { for k, v in merge(var.tags, var.cluster_tags, data.aws_default_tags.current.tags) :
-    k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags
-  }
+  for_each = var.create_cluster_primary_security_group_tags ? { for k, v in merge(var.tags, var.cluster_tags, data.aws_default_tags.current.tags) :
+    k => v if local.create && k != "Name"
+  } : {}
 
   resource_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
   key         = each.key


bryantbiggs:
do you have a reproduction configuration that demonstrates your error message?

dex4er:
@bryantbiggs Unfortunately I don't have it: I have very complex EKS setup which was stopped in the middle of destroying then after I rerun `terraform apply` I've got this error.

Maybe this is just a quirk in Terraform. Maybe it always break if `data` is deleted then it can't calculate correctly `for_each` loop again. I would avoid to use any `data` in any `for_each` and in this case simple ternary operator helped.

dex4er:
@bryantbiggs 
Now I'm convinced https://github.com/terraform-aws-modules/terraform-aws-eks/commit/a74e98017b5dc7ed396cf26bfaf98ff7951c9e2e broke it. Currently I've got the same `Error: Invalid for_each argument` on another project after irrelevant change.

bryantbiggs:
unfortunately, without a reproduction I'm not really able to provide any guidance/support

dex4er:
I'm really disappointed that you don't want to accept trivial patch without long tests that will take a lot of time and $$ on AWS, when the error message is very clear:

```
The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will identify the instances of this resource.
```

It is because in `for_each` there is data source used and it is just a bad practice and it is documented: https://www.terraform.io/language/meta-arguments/for_each#limitations-on-values-used-in-for_each

> The keys of the map (or all the values in the case of a set of strings) must be known values, or you will get an error message that for_each has dependencies that cannot be determined before apply, and a -target may be needed.

bryantbiggs:
@jkotiuk change this

```hcl
  tags = merge(
    data.aws_default_tags.current.tags,
    {
      Terraform = "true"
    }
  )
```

To

```hcl
  tags = {
      Terraform = "true"
    }
```

jkotiuk:
That didn't help, I've tried that already

bryantbiggs:
and what does your provider block look like with the default tags - the reproduction provided does not show any errors

bryantbiggs:
FYI - we are removing all usage of `aws_default_tags` in v19. It simply does not work trying to bake it in and instead this needs to be fixed at the AWS provider level https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2250/commits/ed0c336d1c52c757800c96bb9a88e2fe5f358558

antonbabenko:
This issue has been resolved in version 19.0.0 :tada:

dex4er:
> This issue has been resolved in version 19.0.0 🎉

I want to thank you for almost (cluster_id -> cluster_name) smooth migration from 18.x to 19.x. Perfect job! Using this module is a pleasure.

bryantbiggs:
That is very kind, thank you

PR #2026, opened 2022-04-20
fix: Correct DNS suffix for OIDC provider

xchan:
## Description

This is a supplement fix for
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1904

## Motivation and Context

This MR is a combined fix for succefully launching Karpenter 0.8.2 on AWS EKS on China Region (cn-northwest-1)

## Breaking Changes

N/A

## How Has This Been Tested?
- [*] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [ ] I have executed `pre-commit run -a` on my pull request

AWS EKS version 1.22 + Karpenter 0.8.2 launched succefully with this fix with node scaling test done.

Issue #1904, opened 2022-02-28
incorrect eks service principal for AWS China


yuan-fei:
## Description
When creating EKS cluster IAM role in AWS China account, AWS complains the service principal is incorrect.
```
Error: failed creating IAM Role (wood-cluster-20220228072223989200000001): MalformedPolicyDocument: Invalid principal in policy: "SERVICE":"eks.amazonaws.com.cn"
│       status code: 400, request id: 4f866481-fd28-4baf-b433-447c82ca4e09
│ 
│   with module.eks.aws_iam_role.this[0],
│   on ../../main.tf line 191, in resource "aws_iam_role" "this":
│  191: resource "aws_iam_role" "this" {
│
```
The EKS service principal should be `eks.amazonaws.com` for AWS China (see the [AWS China EKS doc](https://docs.amazonaws.cn/en_us/eks/latest/userguide/service_IAM_role.html)) as well

EKS service principal is specified in the [TF code here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L186) 

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

 - Terraform: Terraform v1.0.11 on darwin_amd64

 - Provider(s):

```
provider registry.terraform.io/hashicorp/aws v4.2.0
provider registry.terraform.io/hashicorp/cloudinit v2.2.0
provider registry.terraform.io/hashicorp/tls v3.1.0
```
 - Module: [terraform-aws-eks](https://github.com/terraform-aws-modules/terraform-aws-eks)

## Reproduction

1.  create test code as bellow
2. `terraform apply`
### Code Snippet to Reproduce

```
provider "aws" {
  region                  = var.region
  shared_credentials_file = var.shared_credentials_file
}

locals {
  cluster_name                          = "wood"
  cluster_version                       = var.k8s_version
  cluster_additional_security_group_ids = var.cluster_additional_security_group_ids
  tags = {
    Example    = local.cluster_name
    GithubRepo = "terraform-aws-eks"
    GithubOrg  = "terraform-aws-modules"
  }

  vpc = {
    vpc_id          = var.vpc_id
[...]

PR #2271, opened 2022-10-12
fix: Wrong STS audience in OIDC provider in China partition

pkolakow:
## Description
Change STS Audience service endpoint to point to the correct partition.

## Motivation and Context
When deploying EKS cluster in china partition you have to specify the following variable because the EKS have `amazonaws.com` not `amazonaws.cn`  dns_suffix:
```
cluster_iam_role_dns_suffix = "amazonaws.com"
```
Unfortunatelly, this variable is also used to determine STS Audience in OIDC provider when IRSA is enabled.
The correct STS endpoint in China partion is `sts.amazonaws.com.cn` not `sts.amazonaws.com` and should not depend on `cluster_iam_role_dns_suffix` variable.

## Breaking Changes
The OIDC provider will be forcfully replaced but because of the wrong STS audience it wasn't usable anyway. 

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -211,7 +211,7 @@ data "tls_certificate" "this" {
 resource "aws_iam_openid_connect_provider" "oidc_provider" {
   count = local.create && var.enable_irsa ? 1 : 0
 
-  client_id_list  = distinct(compact(concat(["sts.${local.dns_suffix}"], var.openid_connect_audiences)))
+  client_id_list  = distinct(compact(concat(["sts.${data.aws_partition.current.dns_suffix}"], var.openid_connect_audiences)))
   thumbprint_list = concat([data.tls_certificate.this[0].certificates[0].sha1_fingerprint], var.custom_oidc_thumbprints)
   url             = aws_eks_cluster.this[0].identity[0].oidc[0].issuer
 


bryantbiggs:
I believe what we have is correct - you can see the history here:
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1904
- https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2026

pkolakow:
@bryantbiggs I disagree. IMO #2026 introduced a bug. I've got multiple environments in China partition and the correct way is to have the following trust relationship for the cluster IAM role:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "EKSClusterAssumeRole",
            "Effect": "Allow",
            "Principal": {
                "Service": "eks.amazonaws.com"
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```
and `sts.amazonaws.com.cn` in the audience in the OIDC provider.

They cannot have the same DNS suffix.

Thus, it is also required to this annotation to ServiceAccount:
```
eks.amazonaws.com/audience: "sts.amazonaws.com.cn"
```

Even if it is possible to use the `sts.amazonaws.com` audience and specify the corresponding annotation to ServiceAcount, the change introduced in #2026 is backward incompatible and will result in all Pods not being to authenticate to AWS for the time ServiceAccount's token is refreshed (also there would be a need to change all ServiceAcount manifests on the cluster).

Maybe it would be reasonable to introduce a separate variable for the default STS audience?

PR #2272, opened 2022-10-12
feat: Support default_tags in tag_specifications

ivankatliarchuk:
## Description
Default tags not applied to resources like `volume,instance and network-interface` for self-managed node groups in `aws_launch_template`

As it looks, there is currently no plans to propagate default tags to `tag_specifications` in   `aws` [provider](https://github.com/hashicorp/terraform-provider-aws/issues/19387)

## Motivation and Context

[it fixes similar closed issue](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1916)

## Breaking changes

No

## How Has This Been Tested?

- [x] tested in one of the custome cluster as well

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -43,6 +43,8 @@ module "user_data" {
 ################################################################################
 
 locals {
+  tags = var.use_default_tags ? merge(data.aws_default_tags.current.tags, var.tags) : var.tags
+
   launch_template_name_int = coalesce(var.launch_template_name, "${var.name}-node-group")
 
   security_group_ids = compact(concat([try(aws_security_group.this[0].id, ""), var.cluster_primary_security_group_id], var.vpc_security_group_ids))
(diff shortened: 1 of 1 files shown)


ivankatliarchuk:
[closed issue](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1968) aws provider does not set default tags for launch template for specific resources. I know it was closed, but maintaining a fork for this module is not great idea. Would like to know how can I prove that its required to be set explicitly?

ivankatliarchuk:
Related explanation in `aws` [provider](https://github.com/hashicorp/terraform-provider-aws/issues/19387)

ivankatliarchuk:
Source [code](https://github.com/hashicorp/terraform-provider-aws/blob/632cff7679cb6e1e14076b9aac3e68b73f584b70/internal/service/ec2/ec2_launch_template.go#L957). There is no tag propagation [here](https://github.com/hashicorp/terraform-provider-aws/blob/632cff7679cb6e1e14076b9aac3e68b73f584b70/internal/service/ec2/ec2_launch_template.go#L3102). It looks like a bug, to me, would be nice to support this functionality on module level while there is still inconsistent in default tags setup

bryantbiggs:
I don't think we will be entertaining any further changes for default tags due to the number of issues https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2264

default tags should be just that - default tags. If you have to start adding support for default tags into your resources and modules than its no longer default tags, its just tags. these issues are better addressed at the provider level since that is where the implementation is controlled https://github.com/hashicorp/terraform-provider-aws/issues?q=is%3Aissue+is%3Aopen+default+tags+sort%3Areactions-%2B1-desc

PR #2274, opened 2022-10-13
fix: Disable creation of cluster security group rules that map to node security group when `create_node_security_group` = `false`

### LINKED ISSUE #2258: Solutions for "expect exactly one securityGroup tagged with kubernetes.io/cluster/<NAME>" don't work as expected
## Description

[The recent addition in the FAQ](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md#i-received-an-error-expect-exactly-one-securitygroup-tagged-with-kubernetesioclustername-) about the `expect exactly one securityGroup tagged with kubernetes.io/cluster/<NAME>` is very much welcomed but it seems some of the solutions don't work properly.

Solution 2 & 3 don't work because `aws-load-balancer-controller` seems to not care about the actual value of the tag.
Whether it is set to empty string or something else, `aws-load-balancer-controller` still complains with `expect exactly one securityGroup tagged with kubernetes.io/cluster/...` 
Also Solution 3 might not work in the long run as EKS might put back the tag upon platform update.

Solution 1 doesn't work out of the box as creating a new cluster with the given setup results in the following errors:
```
Error: One of ['cidr_blocks', 'ipv6_cidr_blocks', 'self', 'source_security_group_id', 'prefix_list_ids'] must be set to create an AWS Security Group Rule
```

This is caused by [cluster_security_group_rules](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L131) referencing the `node_security_group` being always added to the cluster security group despite  `node_security_group_id` being set to null.

This can be solved by either providing our own `node_security_group_id` or disabling the creation or  the cluster_security_group, (`create_cluster_security_group = false`) but it causes the replacement of the EKS cluster for an existing one (and a priori, a custom cluster security group should be provided in that case)

The solution 4. works fine.

(It's in fact linked to https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1986 which is already closed)

## Expected behavior

The documentation should probably be updated to remove 2 & 3 solutions.

For solution 1. it is unclear to me if the module should be fixed to support having no node_security_group set, or if the doc should be updated to indicate that a custom node security group should be created.

bryantbiggs:
## Description

- Disable creation of cluster security group rules that map to node security group when `create_node_security_group` = `false`
- Update Fargate example to run only Fargate and show disabling of both cluster and node security groups

## Motivation and Context

- Resolves #2258 

## Breaking Changes

- No

## How Has This Been Tested?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -26,7 +26,7 @@ By default, EKS creates a cluster primary security group that is created outside
   attach_cluster_primary_security_group = true # default is false
 
   node_security_group_tags = {
-    "kubernetes.io/cluster/<CLUSTER_NAME>" = "" # or any other value other than "owned"
+    "kubernetes.io/cluster/<CLUSTER_NAME>" = null # or any other value other than "owned"
   }
 ```
 
@@ -36,7 +36,7 @@ By default, EKS creates a cluster primary security group that is created outside
   attach_cluster_primary_security_group = true # default is false
 
   cluster_tags = {
-    "kubernetes.io/cluster/<CLUSTER_NAME>" = "" # or any other value other than "owned"
+    "kubernetes.io/cluster/<CLUSTER_NAME>" = null # or any other value other than "owned"
   }
 ```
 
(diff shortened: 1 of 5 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 18.30.2 :tada:

PR #2300, opened 2022-11-15
feat: Add custom options for cluster_autoscaler node group (ASG)

felipeng:
## Description
With this change is possible to set custom options for the cluster autoscaler for particular AutoScalingGroup

## Motivation and Context
Currently is not possible to set custom tags on the AutoScalingGroup and he have a need to add, for example:
`"k8s.io/cluster-autoscaler/node-template/autoscaling-options/scaledownunneededtime" "20m0s"`

With this change is possible to:
```
eks_managed_node_groups = [
  my-ng = {
        name = "my-node-group-with-custom-autoscaler-option"
        cluster_autoscaler = {
          scaledownunneededtime = "20m0s"
        }
  }
]
```

And the workaround:
```
locals {
  cluster_autoscaler_options_tags = merge([
    for name, group in module.eks.eks_managed_node_groups : {
      for option_name, option_value in coalesce(group.node_group_cluster_autoscaler, {}) : "${name}|autoscaling-options|${option_name}" => {
        autoscaling_group = group.node_group_autoscaling_group_names[0],
        key               = "k8s.io/cluster-autoscaler/node-template/autoscaling-options/${option_name}",
        value             = option_value
      }
    }
  ]...)

  cluster_autoscaler_asg_tags =  local.cluster_autoscaler_options_tags
}
```

## Breaking Changes
No

## How Has This Been Tested?
I've tested with the example and provided and it is also working on our environment.


bryantbiggs:
this does not appear to have any affect on any resources - it looks like just a pass through from variable to output. therefore, we won't be accepting this change into the project. thank you though!

felipeng:
You are right, it is just to pass through from variable to output and it is required to be able to set the ASG tags via EKS module. Otherwise, how do you recommend to set the `"k8s.io/cluster-autoscaler/node-template/autoscaling-options/scaledownunneededtime" "20m0s"` to ASG tags?

bryantbiggs:
for an EKS managed node group, you can use the `aws_autoscaling_group_tag` resource

PR #2314, opened 2022-11-30
fix: Unable to change desired_size on managed node groups

Maddog2050:
## Description

Remove ignoring changes to `scaling_config` `desired_size` for managed node groups.

## Motivation and Context

On managed nodes you are unable to change the `desired_size` of the node group as the value is ignored. To change this you have to resort to changing it the CLI or Console.

This causes an issue if for example you have a `desired_size` of 3 and you wish to increase the `min_size` above 3.

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -339,9 +339,6 @@ resource "aws_eks_node_group" "this" {
 
   lifecycle {
     create_before_destroy = true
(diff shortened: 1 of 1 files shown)


bryantbiggs:
thank you but this is by design. clusters scaling should be handled by something like Karpenter or cluster-autoscaler which is why its ignored here by design

PR #2319, opened 2022-12-06
fix: Call to lookup() closed too early, breaks sg rule creation in cluster sg if custom source sg is defined.

AdamStaudt:
## Description

fixed typo in nested lookup() calls in `aws_security_group_rule.cluster` resource

## Motivation and Context

If a security group ID is supplied in the `source_security_group_id` parameter to `cluster_security_group_additional_rules`, the ternary operator will break, as the nested lookup() function will return a string rather than a bool.

## Breaking Changes

No breaking changes.

## How Has This Been Tested?

Change was tested on production code within our organisation, which creates a SG rule with an external SG as its source. The same input worked under v18 and broke under v19.

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -84,6 +84,15 @@ module "eks" {
       type                       = "ingress"
       source_node_security_group = true
     }
+    # Test: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2319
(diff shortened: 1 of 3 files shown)


bryantbiggs (approved):
thank you - looks like there was a bit more to iron out and I added some "test cases" to ensure the functionality works as intended

PR #2320, opened 2022-12-06
fix: skip public_access_cidrs if var.cluster_endpoint_public_access = false

smelchior:
## Description
Do not set public_access_cidr when public access is disabled.

## Motivation and Context

When upgrading from v18 -> v19 with an existing private-only cluster we were shown the following plan:
```
  # module.eks.aws_eks_cluster.this[0] will be updated in-place
  ~ resource "aws_eks_cluster" "this" {
        id                        = "xx"
        name                      = "xx"
        tags                      = {}
        # (11 unchanged attributes hidden)

      ~ vpc_config {
          ~ public_access_cidrs       = [
              - "0.0.0.0/0",
            ]
            # (6 unchanged attributes hidden)
        }

        # (2 unchanged blocks hidden)
    }
```
When applying we got:
```
│ Error: updating EKS Cluster (xx) VPC config: InvalidParameterException: Cluster is already at the desired configuration with endpointPrivateAccess: true , endpointPublicAccess: false, and Public Endpoint Restrictions: [0.0.0.0/0]
```

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -27,7 +27,7 @@ resource "aws_eks_cluster" "this" {
     subnet_ids              = coalescelist(var.control_plane_subnet_ids, var.subnet_ids)
     endpoint_private_access = var.cluster_endpoint_private_access
     endpoint_public_access  = var.cluster_endpoint_public_access
-    public_access_cidrs     = var.cluster_endpoint_public_access ? var.cluster_endpoint_public_access_cidrs : []
+    public_access_cidrs     = var.cluster_endpoint_public_access ? var.cluster_endpoint_public_access_cidrs : null
   }
 
   dynamic "kubernetes_network_config" {


yafanasiev:
Just hit this as well, can verify that setting `public_access_cidrs` to `null` resolves the issue. Thank you @smelchior!

bryantbiggs:
interesting - it looks like our previous default value is causing the issue here. the value is optional and does not have a default https://github.com/hashicorp/terraform-provider-aws/blob/a421ee175fdae8d1d2d5ef71b86d46809b4780fb/internal/service/eks/cluster.go#L257 but I'm guessing that because in prior versions the default value was set to `["0.0.0.0/0"]`, that we can't unset that value

main.tf
```diff
@@ -27,7 +27,7 @@ resource "aws_eks_cluster" "this" {
     subnet_ids              = coalescelist(var.control_plane_subnet_ids, var.subnet_ids)
     endpoint_private_access = var.cluster_endpoint_private_access
     endpoint_public_access  = var.cluster_endpoint_public_access
-    public_access_cidrs     = var.cluster_endpoint_public_access ? var.cluster_endpoint_public_access_cidrs : []
+    public_access_cidrs     = var.cluster_endpoint_public_access_cidrs
```
bryantbiggs, line 30: it looks like our default value of `["0.0.0.0/0"]` is the correct default and we can't set to an empty array so reverting to the previous setup

Reference: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-cluster-resourcesvpcconfig.html#cfn-eks-cluster-resourcesvpcconfig-publicaccesscidrs

bryantbiggs (approved):

antonbabenko:
This PR is included in version 19.0.2 :tada:

PR #2322, opened 2022-12-07
fix: Invalid value for "replace" parameter: argument must not be null.

andpozo:
## Description

replace function doesn't accept null as replacement value
## Motivation and Context

cause an issue when you try to use `module.eks.oidc_provider` it always be null

```shell
$ tf console
Acquiring state lock. This may take a few moments...
> replace("https://test.com.ar", "https://", null)
│
│ Error: Invalid function argument
│
│   on <console-input> line 1:
│   (source code not available)
│
│ Invalid value for "replace" parameter: argument must not be null.
╵

# wrapped inside `try` function it always be null
> try(replace("https://test.com.ar", "https://", null), null)
null
> try(replace("https://test.com.ar", "https://", ""), null)
"test.com.ar"
```

## How Has This Been Tested?

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -105,7 +105,7 @@ output "node_security_group_id" {
 
 output "oidc_provider" {
   description = "The OpenID Connect identity provider (issuer URL without leading `https://`)"
-  value       = try(replace(aws_eks_cluster.this[0].identity[0].oidc[0].issuer, "https://", null), null)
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
thank you for catching this - looks like this change was made inadvertently https://github.com/terraform-aws-modules/terraform-aws-eks/blob/efbe9526322cef7443734c6c8b1e2c166f1cd428/outputs.tf#L108

antonbabenko:
This PR is included in version 19.0.3 :tada:

PR #2323, opened 2022-12-07
fix: Allow all traffic on node to node

rickhlx:
## Description
Allow all on node to node traffic

## Motivation and Context
We noticed random timeout issues when accessing Services after removing this rule when upgrading to `19.0.0` version of this module. When the allow all traffic rule was added back, our timeout issues stopped.

The [Amazon EKS security group considerations](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) suggests the use of all traffic allowed from self on nodes.

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -108,10 +108,10 @@ locals {
 
   node_secuirty_group_recommended_rules = { for k, v in {
     ingress_nodes_ephemeral = {
-      description = "Node to node ingress on ephemeral ports"
-      protocol    = "tcp"
-      from_port   = 1025
-      to_port     = 65535
+      description = "Node to node ingress on all traffic"
+      protocol    = "-1"
+      from_port   = 0
+      to_port     = 0
       type        = "ingress"
       self        = true
     }


bryantbiggs:
this rule wasn't on the prior v18 version - users are free to add this rule into their security group though if they wish

bryantbiggs:
that does not quite jive with my understanding. the webhook for karpenter is coming from the control plane to the pod is [8443](https://github.com/aws/karpenter/blob/b6f87def5bb772362d7cec9fab7975b572d9a402/charts/karpenter/README.md?plain=1) which is one of the reasons why its added in as a "recommended" rule https://github.com/terraform-aws-modules/terraform-aws-eks/blob/7124d7603a81ad07bbcc7a932ba0f41989365471/node_groups.tf#L118

for other services - absolutely, they can use other ports that are not within the ephemeral ports range and that is where users are encouraged to open up the ports that their services are communicating on.

rickhlx:
The linked `8443` port is the container port for Karpenter, however the service port for Karpenter is set to [443](https://github.com/aws/karpenter/blob/b6f87def5bb772362d7cec9fab7975b572d9a402/charts/karpenter/templates/service.yaml#L20). 

https://karpenter.sh/v0.19.3/troubleshooting/#failed-calling-webhook-defaultingwebhookprovisionerskarpentersh

bryantbiggs:
hmm, maybe there is a larger issue here https://github.com/aws/karpenter/issues/2983

bryantbiggs:
have you seen this guidance https://github.com/aws/karpenter/issues/2876#issuecomment-1333526664

rickhlx:
Interesting, I'll take a look because we did recently update Karpenter. 

However, our issues related to the ports were not restricted to Karpenter, opening all node-to-node traffic did resolve those timeout issues. 

Just wanted to leave it recorded in case someone runs into these issues when using default settings.

bryantbiggs:
> opening all node-to-node traffic did resolve those timeout issues.

Yes, opening all access on security features does tend to resolve issues 😅 but I would just caution that this is not the approach this module takes nor promotes. If there is a valid case presented, we can evaluate it (like we did for adding 8443 and 9443 which is what a lot of common addons utilize on EKS for htttps traffic) but I don't necessarily agree with the statement that opening all ports/protocols is the answer here. Users are strongly encouraged to understand what access their services need and make the appropriate changes to benefit from a stronger security posture, but ultimately the choice is theres

rickhlx:
> Yes, opening all access on security features does tend to resolve issues 😅

LOL, very true! However, the config I was coming from was taken from the examples in this very repo, which is opening all internode traffic.

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/complete/main.tf#L99

I do agree we should know what port/protocol is causing our issues though, so if we can determine that, we can come back and make a suggestion.

Thanks for the discussion!

bryantbiggs:
thank you!

Issue #2327, opened 2022-12-07
KMS Policy error in 19.0 when using iam role for aws provider


pdeva:
## Description

Trying to create a new eks cluster with module version 19.0 runs into a `MalformedPolicyDocumentException` error while creating the kms key internally.

- [ x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 19.0.3

- Terraform version: 1.3.6

- Provider version(s): aws provider: 4.45.0

## Reproduction Code [Required]

This simple code is enough to reproduce the error:

```hcl
module "eks_cluster" {
  source          = "terraform-aws-modules/eks/aws"
  version         = "~> 19.0"
  vpc_id          = module.vpc.vpc_id
  cluster_name    = "k1"
  cluster_version = "1.24"
  subnet_ids      = module.vpc.private_subnet_ids

  cluster_endpoint_private_access = true
  cluster_endpoint_public_access  = false

}
```

AWS provider configuration:

```hcl
provider "aws" {
  region              = "us-east-1"
  allowed_account_ids = ["172512501118"]
  assume_role {
    role_arn = "arn:aws:iam::172512501118:role/terraform"
  }
}
```

the iam `role` above has `AdministrativeAccess` policy attached.

## Expected behavior

Cluster should be created without errors.

## Actual behavior

Getting this error:

```
Error: error creating KMS Key: MalformedPolicyDocumentException: The new key policy will not allow you to update the key policy in the future.
with module.us_east_1.module.eks_cluster.module.kms.aws_kms_key.this[0]
on .terraform/modules/us_east_1.eks_cluster.kms/main.tf line 8, in resource "aws_kms_key" "this":
resource "aws_kms_key" "this" {
```

### Terminal Output Screenshot(s)
<img width="1143" alt="Screenshot 2022-12-07 at 6 57 48 AM" src="https://user-images.githubusercontent.com/3190176/206212853-8b88384f-9b5c-43bd-b709-c8ed8555daa4.png">

bryantbiggs:
please check open issues first per the template: duplicate of #2321 and #2325

pdeva:
hi @bryantbiggs. the reason to create this issue was that those 2 issues seem to relate to upgrading from older version of module to newer version. And there the OP wants to turn off encryption completely.  

I think the right solution would have the cluster be created with encryption and without any errors, which is what I am trying to do here.

bryantbiggs:
[...]

PR #2347, opened 2022-12-16
fix: Use IAM session context data source to resolve the identities role when using `assumed_role`

bryantbiggs:
## Description
- Use IAM session context data source to resolve the identities role when using `assumed_role`

## Motivation and Context
- Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2327#issuecomment-1355581682
- Ref: https://github.com/hashicorp/terraform-provider-aws/issues/28381

## Breaking Changes
- No, this is backwards compatible

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,14 @@
 data "aws_partition" "current" {}
 data "aws_caller_identity" "current" {}
 
+data "aws_iam_session_context" "current" {
+  # This data source provides information on the IAM source role of an STS assumed role
+  # For non-role ARNs, this data source simply passes the ARN through issuer ARN
+  # Ref https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2327#issuecomment-1355581682
+  # Ref https://github.com/hashicorp/terraform-provider-aws/issues/28381
+  arn = data.aws_caller_identity.current.arn
+}
+
 locals {
   create = var.create && var.putin_khuylo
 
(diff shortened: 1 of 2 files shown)


main.tf
```diff
@@ -1,6 +1,14 @@
 data "aws_partition" "current" {}
 data "aws_caller_identity" "current" {}
 
+data "aws_iam_session_context" "current" {
```
antonbabenko, line 4: Should we consider using this resource instead of an existing one in other modules also? WDYT?

antonbabenko (approved):

main.tf
bryantbiggs: yes, I think anywhere that we are using `data.aws_caller_identity.current.arn` to get the current identity's ARN, we should use this session context data source to ensure we are getting the true ARN and not the ARN of the STS session

antonbabenko:
This PR is included in version 19.1.1 :tada:

main.tf
```diff
@@ -122,7 +130,7 @@ module "kms" {
   # Policy
   enable_default_policy     = var.kms_key_enable_default_policy
   key_owners                = var.kms_key_owners
-  key_administrators        = coalescelist(var.kms_key_administrators, [data.aws_caller_identity.current.arn])
+  key_administrators        = coalescelist(var.kms_key_administrators, [data.aws_iam_session_context.current.issuer_arn])
```
FernandoMiguel, line 133: great... this should fix the constant drift we are seeing for a ephemeral session

PR #2360, opened 2022-12-19
feat: Allow configuring which tags are passed on launch template tag specifications

bryantbiggs:
## Description
- Allow configuring which tags are passed on launch template tag specifications

## Motivation and Context
- It appears that not all regions support the same level of tagging
- Resolves #2355
- Resolves #2356
- Resolves #2361

## Breaking Changes
- No

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #2363, opened 2022-12-20
fix: Drop spot-instances-request from tag_specifications

pavlospt:
## Description
`spot-instances-request` is not supported in some regions, so we are dropping that from default `tag_specifications`.

## Motivation and Context
* https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2360#issuecomment-1358168905
* https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2360#issuecomment-1358488428


bryantbiggs:
so far I have only heard one user report one region (Ireland) - so I am less inclined to drop this one since its not as problematic. do you know what other regions do not support this?

lordspacehog:
It also causes issues if you try to launch node groups without using spot instances at all. much like the Elastic GPU instances it doesn't allow you to specify the tag if you're not launching spot instances.

bryantbiggs:
yes I understand the behavior - that was a bit surprising when it was discovered with the tagging request for GPUs. However, this still appears to be isolated to one region so far (that I can tell - docs are not producing any additional results so far)

pavlospt:
@bryantbiggs even if it is incosistency with just one region, why should it not be fixed? I mean AWS should be supposedly consistent, but since they are not IMO it feels wrong to enforce such a change and possibly be breaking even just this one region!

bryantbiggs (approved):

bryantbiggs:
`Failed: You cannot specify tags for Spot instances requests if there are no Spot instances requests being created by the request. Launching EC2 instance failed.` - well this is some somewhat surprising behavior so yes, reverting this out and making it an opt in

antonbabenko:
This PR is included in version 19.4.2 :tada:

PR #2364, opened 2022-12-20
Fix: self_managed_node_group_defaults tag_specification

lordspacehog:
## Description
Fix reference for self_manage_node_groups to pull tag_specification from the correct defaults

## Motivation and Context
Module was not allowing self managed node group defines to respect tag_specification from self managed node group defaults definition


bryantbiggs (approved):
thank you!

antonbabenko:
This PR is included in version 19.4.1 :tada:

Issue #2379, opened 2023-01-05
make manage_aws_auth_configmap=false also ignore metadata


mattparkes:
## Is your request related to a problem? Please describe.
We have a slightly unique situation in that we provision some of our cluster with this module, but then manage the contents of the `aws-auth` ConfigMap with GitOps and ArgoCD. 

We've achieved 99% of what we want with `create_aws_auth_configmap=true` (so we can bootstrap our cluster with a minimum viable auth config) and `manage_aws_auth_configmap=false` (so that ArgoCD handles the contents of the repo from then on).

The _only_ issue we have is that ArgoCD automatically adds a label to every resource it updates and because lifecycle_ignore is only set on `data`, we always have this change pending:

```
Terraform will perform the following actions:

  # module.eks-cluster.kubernetes_config_map.aws_auth[0] will be updated in-place
  ~ resource "kubernetes_config_map" "aws_auth" {
        id          = "kube-system/aws-auth"
        # (3 unchanged attributes hidden)

      ~ metadata {
          ~ labels           = {
              - "argocd.argoproj.io/instance" = "foo-bar-aws-auth" -> null
            }
            name             = "aws-auth"
            # (5 unchanged attributes hidden)
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

## Describe the solution you'd like.
I'd like to change the `ignore_changes` on `kubernetes_config_map.aws_auth` from `[data]` to `[data, metadata.labels, metadata.annotations]` _(actually needs to be `[data, metadata[0].labels, metadata[0].annotations]` for ...reasons)_.

This would allow people to add  whatever metadata or annotations they want to this resource.

## Describe alternatives you've considered.
We could make another input called something like `aws_auth_labels=[]`, but this seems less desirable than my proposed solution and clutters the inputs with stuff that will likely never get used, as you'd also probably have to do `aws_auth_annotations=[]` whilst you're there to cover all bases.

antonbabenko:
This issue has been resolved in version 19.5.0 :tada:

PR #2380, opened 2023-01-05
feat: Ignore changes to labels and annotations on on `aws-auth` ConfigMap

mattparkes:
## Description
As per #2379, we manage the contents of the `aws-auth` ConfigMap with GitOps and ArgoCD.

The only issue we have is that ArgoCD automatically adds a label to every resource it updates and because `lifecycle_ignore` is only set on `data` we always have `terraform` wanting to remove the `metadata.label` added by ArgoCD (even though we have `manage_aws_auth_configmap=false` set, which semantically implies that this won't happen).

## Motivation and Context
Fixes: #2379

## Breaking Changes
This causes has no breaking changes.

## How Has This Been Tested?
- [x] I have tested and validated these changes using our testing environment.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -520,7 +520,7 @@ resource "kubernetes_config_map" "aws_auth" {
   lifecycle {
     # We are ignoring the data here since we will manage it with the resource below
     # This is only intended to be used in scenarios where the configmap does not exist
-    ignore_changes = [data]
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
makes sense, thank you!

antonbabenko:
This PR is included in version 19.5.0 :tada:

PR #2382, opened 2023-01-05
Change the way creating well known policy attachment

maordavidov:
## Description
Change the way creating well known policy attachment. Instead of creating them with for_each which cannot
be executed before some resources are created, we will use explicit policy creation. It fix the issue when running terraform plan before EKS cluster is created.

## Motivation and Context
When running terraform plan, before the EKS cluster is created there is an error that the values are determined in runtime, so the for_each fail the terraform plan. 

## Breaking Changes
It will replace the policy attachments, but shouldn't break anything

## How Has This Been Tested?

### DIFF
--- a/modules/fargate-profile/main.tf
+++ b/modules/fargate-profile/main.tf
@@ -40,13 +40,15 @@ resource "aws_iam_role" "this" {
   tags = merge(var.tags, var.iam_role_tags)
 }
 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
Please start with an issue that states the issue, any output, and a reproduction to the issue. However, this change won't be made and is working as intended

Issue #2385, opened 2023-01-05
AMI lookup in self-managed-node-group module should only be required if creating a launch template


demolitionmode:
## Is your request related to a problem? Please describe.

I'm using an instance of the `self-managed-node-group` module to create **only** an IAM role (as a workaround for https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1753), yet I am required to pass a cluster version for the purpose of looking up an AMI for which I will not use.

I am using the following to call the module

```terraform
module "self_managed_node_group_iam_role" {
  source  = "terraform-aws-modules/eks/aws//modules/self-managed-node-group"
  version = "~> 19.0"

  for_each = local.self_managed_node_groups

  create_autoscaling_group = false
  create_launch_template   = false
  create_schedule          = false

  cluster_version = each.value.version
}
```

I want to be able to omit `cluster_version` from the above config, as it is unrelated to creating the IAM Role.

## Describe the solution you'd like.

Given that the `data.aws_ami.eks_default` is only used in the `aws_launch_template.this` resource, which has a conditional `var.create_launch_template`, the data block should be updated to have the same conditional.

PR #2386, opened 2023-01-05
Make AMI lookup conditional on creation of launch template

demolitionmode:
## Description

Given that `data.aws_ami.eks_default` is only used in the `aws_launch_template.this` resource, which has a conditional `var.create_launch_template`, the data block has been updated to have the same conditional. This means that `var.cluster_version` is no longer needed to be passed when `var.create_launch_template` is set to false (ie. when the available EKS AMIs make no difference to the resources actually being created).

## Motivation and Context

Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2385

## Breaking Changes

No breaking changes

## How Has This Been Tested?

Tested with change made against latest on master branch, and the variables shown in related issue.

Prior to change, omission of `cluster_version` results in:

```bash
│ Error: Invalid template interpolation value
│
│   on .terraform/modules/self_managed_node_group_iam_role/modules/self-managed-node-group/main.tf line 9, in data "aws_ami" "eks_default":
│    9:     values = ["amazon-eks-node-${var.cluster_version}-v*"]
│     ├────────────────
│     │ var.cluster_version is null
│
│ The expression result is null. Cannot include a null value in a string
│ template.
```

After change, omission of `cluster_version` is valid

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -2,7 +2,7 @@ data "aws_partition" "current" {}
 data "aws_caller_identity" "current" {}
 
 data "aws_ami" "eks_default" {
-  count = var.create ? 1 : 0
+  count = var.create && var.create_launch_template ? 1 : 0
 
   filter {
     name   = "name"


bryantbiggs (approved):
makes sense since the launch template is the only consumer of the resource - thank you!

antonbabenko:
This PR is included in version 19.5.1 :tada:

demolitionmode:
Thank you for your quick review! 🙇

Issue #2337, opened 2022-12-09
Error: Invalid for_each argument


AndreiBanaruTakeda:
## Description

Please provide a clear and concise description of the issue you are encountering, and a reproduction of your configuration (see the `examples/*` directory for references that you can copy+paste and tailor to match your configs if you are unable to copy your exact configuration). The reproduction MUST be executable by running `terraform init && terraform apply` without any further changes.

If your request is for a new feature, please use the `Feature request` template.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: 19.0.4

- Terraform version: 1.2.7

- Provider version(s): 

## Reproduction Code [Required]

```hcl
module "eks_managed_node_groups" {
  source  = "terraform-aws-modules/eks/aws//modules/eks-managed-node-group"
  version = "19.0.4"

  for_each        = { for index, name in var.node_group_config : index => name }
  cluster_name    = var.cluster_name
  name            = join("", [var.cluster_name, each.value.capacity_type == "ON_DEMAND" ? "-D-NG-" : "-S-NG-", each.key < 9 ? "00" : "0", each.key + 1])
  use_name_prefix = false

  vpc_security_group_ids = var.vpc_security_group_ids

  create_iam_role            = false
  iam_role_arn               = var.iam_role_arn
  iam_role_attach_cni_policy = false

  subnet_ids = var.subnet_ids

  min_size     = each.value.minimum_size
  max_size     = each.value.maximum_size
  desired_size = each.value.desired_size

  labels = each.value.labels

  create_launch_template          = true
  launch_template_name            = join("", [var.cluster_name, each.value.capacity_type == "ON_DEMAND" ? "-D-NG-" : "-S-NG-", each.key < 9 ? "00" : "0", each.key + 1])
  launch_template_use_name_prefix = false

  block_device_mappings = {
    sda1 = {
      device_name = "/dev/xvda"
      ebs = {
        volume_type = "gp3"
        volume_size = each.value.disk_size
      }
    }
  }

  ami_id                     = var.ami_id
  ami_type                   = "CUSTOM"
[...]

PR #2388, opened 2023-01-08
Moving the conditional out of the for_each

wpbeckwith:
## Description
Currently by having the conditional check in the for_each terraform will fail to plan with the dreaded "Error: Invalid for_each argument" error.  This happens even if you are not trying to use the module provided role.  Moving the conditional check out allows terraform to plan successfully.

## Motivation and Context

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2337

## Breaking Changes

No
## How Has This Been Tested?

I have added the following to the bottom of the examples/eks_managed_node_group/main.tf
```

module "foo_mng" {
  source  = "../../modules/eks-managed-node-group"

  name            = "foo"
  cluster_name    = "foo"
  cluster_version = "1.24"
  create_iam_role = false
  iam_role_arn    = "arn:aws:iam::123456789012:role/foo-eks-node-group"

 cluster_primary_security_group_id = module.eks.node_security_group_id

  subnet_ids             = module.vpc.private_subnets
  min_size               = 1
  desired_size           = 1
  max_size               = 1
  instance_types         = ["t3.large"]
}

```
With the above, the example continues to plan successfully.  In my env, the PR change allows me to successfully plan and create a cluster while the previous code caused a for_each plan failure.

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -431,11 +431,11 @@ resource "aws_iam_role" "this" {
 
 # Policies attached ref https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group
 resource "aws_iam_role_policy_attachment" "this" {
-  for_each = { for k, v in toset(compact([
+  for_each = var.create && var.create_iam_role ? { for k, v in toset(compact([
     "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy",
     "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly",
     var.iam_role_attach_cni_policy ? local.cni_policy : "",
-  ])) : k => v if var.create && var.create_iam_role }
+  ])) : k => v } : {}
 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
This is not the correct way

wpbeckwith:
What is not the correct way? I forked the repo, made a change and then opened a PR to send the change back.  What's incorrect?

bryantbiggs:
The current implementation is the correct manner. This form raises issues with inconsistent types

wpbeckwith:
Not trying to be extra dense, but I'm not following.  Can you clarify a bit more what you mean by the "This form raises issues with inconsistent types"?  The PR does not change any types, so that's what is confusing.

wpbeckwith:
Regardless, I see reading back on the PR while the code may be correct I have since created a better example of the issue in repo https://github.com/wpbeckwith/terraform-aws-eks-2337.  So a more accurate PR should be created.

sanarena:
This patch worked for me, I will be using this patch until official AWS module fix this issue. @bryantbiggs

PR #2307, opened 2022-11-22
fix: Include all certificate fingerprints in the OIDC provider thumbprint list

ThetaSinner:
## Description
Include all certificate fingerprints in the thumbprint list, rather than just the first one

## Motivation and Context
I'm following the Terraform documentation for enabling IRSA https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#enabling-iam-roles-for-service-accounts. This suggests that all the certificates thumbprints should be listed. This module currently only selects the first one, which should be the highest root certificate in the chain pulled back by the `tls_certificate` data lookup. I can't see anything to suggest that this is intentional.

## Breaking Changes
No breaking changes

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects
- [x] I have executed `pre-commit run -a` on my pull request

I only have access to my work environment. I have tested this change against our development and staging environments without issues.

PR #2393, opened 2023-01-12
fix: Revert PR 2307 - Get only the root CA fingerprint

alfredo-gil:
## Description
With this [PR 2307](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2307) the module is getting all certificate fingerprints in the OIDC provider thumbprint list, but as it's recommended by AWS it's recommended only to get the Root CA fingerprint. The documentation can be found [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc_verify-thumbprint.html)

## Motivation and Context

With this change, we are trying to follow the AWS recommendations for getting only the fingerprint of the `top intermediate CA in the certificate authority chain` as it's mentioned in the point 5 of the [official documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc_verify-thumbprint.html)
```
5. In your command window, scroll up until you see a certificate similar to the following example. If you see more than one certificate, find the last certificate displayed (at the end of the command output). This contains the certificate of the top intermediate CA in the certificate authority chain.
```

It reverts this PR https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2307
## Breaking Changes
No breaking changes

## How Has This Been Tested?

I only have access to my work environment. I have tested this change against our different environments without issues

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -228,7 +228,7 @@ resource "aws_iam_openid_connect_provider" "oidc_provider" {
   count = local.create && var.enable_irsa && !local.create_outposts_local_cluster ? 1 : 0
 
   client_id_list  = distinct(compact(concat(["sts.${local.dns_suffix}"], var.openid_connect_audiences)))
-  thumbprint_list = concat(data.tls_certificate.this[0].certificates[*].sha1_fingerprint, var.custom_oidc_thumbprints)
+  thumbprint_list = concat(data.tls_certificate.this[0].certificates[0].sha1_fingerprint, var.custom_oidc_thumbprints)
   url             = aws_eks_cluster.this[0].identity[0].oidc[0].issuer
 
   tags = merge(


bryantbiggs:
are you seeing issues with having all of the available certificates?

alfredo-gil:
Currently, I don't see a specific issue, but if we need to add more fingerprints to the thumbprint list we are reaching the limit that is 5, and if the AWS official recommendation is to use only the root CA I think is better to keep it like this from a security perspective

bryantbiggs:
without a valid reason, I don't see any justification to revert this. If the AWS API returns multiple thumbprints, and the OIDC provider accepts multiple thumbprints, I think it makes sense to pass all of those provided by the AWS API

alfredo-gil:
The reason is to configure the thumbprint following the official AWS documentation, from my point of view it makes sense to follow the AWS documentation.

bryantbiggs:
thank you but we will not be making this change at this time

PR #2419, opened 2023-01-24
chore: Add more granular types to aws_auth variables

dev-slatto:
## Description

To make variables regarding the auth config map a bit more type safe, I've added the object structure that's expected from Kubernetes in the ConfigMap. 

## Motivation and Context

To avoid sending non-compliant objects into the aws-auth ConfigMap.

## Breaking Changes

Should not be any breaking changes here. If any users are experiencing issues there is a high probability that their ConfigMap isn't working anyways. 

## How Has This Been Tested?


bryantbiggs:
what would be the benefit here, what does "type safe" offer in this context? Not all of these fields are required, but this change would make all fields required

dev-slatto:
Hey @bryantbiggs! 👋🏼 
First of all you're able to get auto completion on the module when utilising the variable, something that's super helpful in my opinion. Second, the module have already a default value set in this case. So there will not be any changes in the behaviour of the module if you decide to not use the variable when using this module downstream. There is really no use case I can find where you don't want to use all the fields in these objects as the config wouldn't work without all of them?

bryantbiggs:
this change alone https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2419/files#diff-05b5a57c136b6ff596500bcbfdcff145ef6cddea2a0e86d184d9daa9a65a288eR581 is a breaking change because if users have not set a value, it now breaks their use until they provide a value. And, a value is not required for this attribute

I'm not sure I see the benefit of this addition at this time. I know we would like to get to a point where the inputs have a better "known" shape, but that situation only recently improved starting in Terraform 1.3 and I don't think we need to force users to adopt the latest version of Terraform at this time just to have better support for Terraform variable type definitions with optional parameters

dev-slatto:
Could you please share what version of Terraform you're experiencing a breaking change on? Since you have set a default value on your variable, users that have not set a value will not experience anything breaking. I've personally implemented that type change in several production and dev clusters today without finding any issues or changes for that sake when not setting a value. Note that the default value was set before this PR so that's not something new and should therefore not create any major changes.

dev-slatto:
@bryantbiggs any thoughts or info about my comment above so I can debug some more? 😄

bryantbiggs:
Thank you but we won't be making this change at this time

PR #2437, opened 2023-01-31
feat: Allow to pass prefix for rule names

omerlh:
## Description
Added variable to control the prefix used for event bridge rules.

## Motivation and Context

We are trying to use ABAC to control permissions of our CI/CD, using `ResourceTag` IAM condition wherever possible. It seems that EventBridge IAM permissions does not allow that so I would like to limit permissions by resource ARN, and for that I would like to control the prefix used for all rules - so I could give our CI/CD tool permissions to control only rules matching this prefix.

## How Has This Been Tested?
(Still testing, opened for feedback because it's my first PR)

### DIFF
--- a/modules/karpenter/main.tf
+++ b/modules/karpenter/main.tf
@@ -261,7 +261,7 @@ locals {
 resource "aws_cloudwatch_event_rule" "this" {
   for_each = { for k, v in local.events : k => v if local.enable_spot_termination }
 
-  name_prefix   = "Karpenter${each.value.name}-"
+  name_prefix   = "${var.rule_name_prefix}${each.value.name}-"
   description   = each.value.description
   event_pattern = jsonencode(each.value.event_pattern)
 
--- a/modules/karpenter/variables.tf
+++ b/modules/karpenter/variables.tf
@@ -224,3 +224,13 @@ variable "create_instance_profile" {
   type        = bool
   default     = true
 }
+
+################################################################################
+# Event Bridge Rules
+################################################################################
+
+variable "rule_name_prefix" {
+  description = "Prefix used for all event bridge rules"
+  type        = string
+  default     = "Karpenter"
+}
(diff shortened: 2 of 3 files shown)


modules/karpenter/main.tf
```diff
@@ -256,12 +256,14 @@ locals {
       }
     }
   }
+
+  rule_name_prefix = coalesce(var.rule_name_prefix, "Karpenter")
```
bryantbiggs, line 260: Remove this

modules/karpenter/main.tf
```diff
@@ -256,12 +256,14 @@ locals {
...
 }
 
 resource "aws_cloudwatch_event_rule" "this" {
   for_each = { for k, v in local.events : k => v if local.enable_spot_termination }
 
-  name_prefix   = "Karpenter${each.value.name}-"
+  name_prefix   = "${local.rule_name_prefix}${each.value.name}-"
```
bryantbiggs, line 266:
```suggestion
  name_prefix   = "${var.rule_name_prefix}${each.value.name}-"
```

modules/karpenter/variables.tf
```diff
@@ -224,3 +224,13 @@ variable "create_instance_profile" {
...
 }
+
+################################################################################
+# Event Bridge Rules
+################################################################################
+
+variable "rule_name_prefix" {
+  description = "Prefix used for all event bridge rules"
+  type        = string
+  default     = null
```
bryantbiggs, line 235:
```suggestion
  default     = "Karpenter"
```

modules/karpenter/main.tf
omerlh: fixed

modules/karpenter/main.tf
omerlh: fixed

modules/karpenter/variables.tf
omerlh: fixed

bryantbiggs (approved):

antonbabenko:
This PR is included in version 19.7.0 :tada:

Issue #2439, opened 2023-02-01
Move to IAM Policy vs Inline Policy for logs:CreateLogGroup


zestrells:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.
Running AWS Config, a finding that comes up is with the dynamic inline policy `logs:CreateLogGroup`. Ideally, we would move towards a IAM policy rather than inline policy [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L298) so we can remediate this AWS Config finding.

## Describe the solution you'd like.
Ideally, the solution would be to turn the inline policy into a dynamic IAM policy.

## Describe alternatives you've considered.
Alternatives include ignoring this finding and writing it as an exception.

## Additional context
This is related to turning on AWS Config for Gov Cloud Fedramp Compliancy.

PR #2440, opened 2023-02-01
feat: Move to IAM Policy vs Inline Policy for logs:CreateLogGroup

zestrells:
## Description
Move to IAM Policy vs Inline Policy for logs:CreateLogGroup
Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2439

## Motivation and Context
Running AWS Config, a finding that comes up is with the dynamic inline policy logs:CreateLogGroup. Ideally, we would move towards a IAM policy rather than inline policy [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L298) so we can remediate this AWS Config finding.

## Breaking Changes
No Breaking Changes

## How Has This Been Tested?


bryantbiggs:
Thank you but we are not planning to make this change at this time

PR #2470, opened 2023-02-13
Ease kubernetes provider configuration

TheSAS:
## Description
Improve the way to pass token to the kubernetes provider.

## Motivation and Context
It solves the issue of the initial `aws` configuration that is called externally. In most cases the AWS provider configuration used in terraform would  differ from the config of `aws` command.

## Breaking Changes
No.

## How Has This Been Tested?

P.S. It might need more testing, as i was just started and torn down the EKS cluster.


bryantbiggs:
thank you but the way it is defined today is the correct method and the method that is recommended by the providers

PR #2478, opened 2023-02-15
feat: Add support for enabling addons before data plane compute is created

### LINKED ISSUE #2467: Support for "early" cluster_addons (vpc-cni with ENABLE_PREFIX_DELEGATION)

Yes

As described on [EKS Cluster w/ Prefix Delegation](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/ipv4-prefix-delegation) (https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2155), the `vpc-cni` is a special cluster addon that needs to be configured **before** any EC2 instances (node groups) are created. If you naively apply it [like this but without `vpc_cni_irsa`](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/eks_managed_node_group/main.tf#L67-L72), initial nodes have a different configuration than later ones.

The [recommended workaround](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/examples/ipv4-prefix-delegation/main.tf#L82) is to configure the `aws_eks_addon` `vpc-cni` separately without the [the `depends_on` directive](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L397-L401).

To generalize it seems there are "early" and "normal" `cluster_addons`, some need to be applied before node groups are created other ones after. It would therefore be a trivial change to improve the Terraform module UX by changing the following `depends_on` directive:

- https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L397-L401

By introducing `local.early_cluster_addons = ["vpc-cni"]` and using:
```
  depends_on = contains(local.early_cluster_addons, each.key) ? [] : [
    module.fargate_profile,
    module.eks_managed_node_group,
    module.self_managed_node_group,
  ]
```

Or by introducing an `apply_early` option for `cluster_addons` and using:
```
  depends_on = each.value.apply_early ? [] : [
    module.fargate_profile,
    module.eks_managed_node_group,
    module.self_managed_node_group,
  ]
```

bryantbiggs:
## Description
- Add support for creating addons before data plane compute is created
- Add dependency for IPv6 policy to ensure policy cleanup/deletion happens without error

## Motivation and Context
- Currently, the addon implementation has a hard dependency on the compute constructs (nodegroups/Fargate). This was added previously due to the fact that the EKS API would attempt to create the addon before there was compute which resulted in a timeout failure. By making the addons depend on the compute, this ensured there were instances available to deploy the addons to. However, this is not valid for ALL addons; namely daemonsets need to be configured and deployed BEFORE compute. With this second addon resource that is not dependent on the compute, coupled with the sleep timer, this gives the addon enough time to provision before compute is created
- Under the current example and module structure, if you attempt to create the IPv6 CNI IAM policy with the module, and then destroy the cluster, it will error with a dependency violation claiming that the policy cannot be deleted when its used by other resources. By adding the dependency for the policy, this has resolved the dependency conflict and allows the cluster and policies to be destroyed without error
- Resolves #2467 

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -19,6 +19,20 @@ locals {
   }
 }
 
+resource "time_sleep" "this" {
+  count = var.create ? 1 : 0
+
+  create_duration = var.dataplane_wait_duration
+
+  triggers = {
+    cluster_name     = aws_eks_cluster.this[0].name
+    cluster_endpoint = aws_eks_cluster.this[0].endpoint
+    cluster_version  = aws_eks_cluster.this[0].version
+
+    cluster_certificate_authority_data = aws_eks_cluster.this[0].certificate_authority[0].data
+  }
+}
+
 ################################################################################
 # EKS IPV6 CNI Policy
 # TODO - hopefully AWS releases a managed policy which can replace this
@@ -220,7 +234,7 @@ module "fargate_profile" {
   create = try(each.value.create, true)
 
   # Fargate Profile
-  cluster_name      = aws_eks_cluster.this[0].name
+  cluster_name      = time_sleep.this[0].triggers["cluster_name"]
   cluster_ip_family = var.cluster_ip_family
   name              = try(each.value.name, each.key)
   subnet_ids        = try(each.value.subnet_ids, var.fargate_profile_defaults.subnet_ids, var.subnet_ids)
@@ -255,8 +269,8 @@ module "eks_managed_node_group" {
 
   create = try(each.value.create, true)
 
-  cluster_name      = aws_eks_cluster.this[0].name
-  cluster_version   = try(each.value.cluster_version, var.eks_managed_node_group_defaults.cluster_version, aws_eks_cluster.this[0].version)
+  cluster_name      = time_sleep.this[0].triggers["cluster_name"]
+  cluster_version   = try(each.value.cluster_version, var.eks_managed_node_group_defaults.cluster_version, time_sleep.this[0].triggers["cluster_version"])
   cluster_ip_family = var.cluster_ip_family
 
   # EKS Managed Node Group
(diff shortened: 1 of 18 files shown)


node_groups.tf
```diff
@@ -19,6 +19,20 @@ locals {
   }
 }
 
+resource "time_sleep" "this" {
+  count = var.create ? 1 : 0
+
+  create_duration = var.dataplane_wait_duration
+
+  triggers = {
+    cluster_name     = aws_eks_cluster.this[0].name
```
bryantbiggs, line 28: because all of these would trigger a downstream update, they make valid triggers here to ensure the nodegroups/Fargate profiles wait for this duration before proceeding

antonbabenko, line 28: I understand this now, but maybe add a comment somewhere near or inside of this resource for the next review when we will forget the whole story.

The reasoning and the solution are neat but not very obvious. :)

antonbabenko (commented):
LGTM. I have just a minor suggestion to improve the understandability of it.

node_groups.tf
bryantbiggs: updated in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2478/commits/99a837c324bfedab8f686ea3cb94410ae2c5d550

antonbabenko (approved):

antonbabenko:
This PR is included in version 19.9.0 :tada:

node_groups.tf
```diff
@@ -19,6 +19,25 @@ locals {
   }
 }
 
+# This sleep resource is used to provide a timed gap between the cluster creation and the downstream dependencies
```
code-eg, line 22: Is there any documented reason we cannot use a `depends_on` object on all compute modules referencing `aws_eks_addon.before_compute`? Is this purely to avoid module dependencies?

If there was an actual dependency in place, there could be additional benefits (e.g. creating the VPC CNI with IRSA and eliminating the need for removing any instance roles after cluster creation). Anecdotally `cluster_addon` creation is incredibly variable (I've seen it finish in seconds, and sometime in minutes).

PR #2489, opened 2023-02-22
docs: Add info about new variable to incompatible changes

buettner123:
## Description
The new variable `node_security_group_enable_recommended_rules` is set to true by default and might cause incompatibilities with previous, additional rules.

## Motivation and Context
While upgrading from version 18 to 19 I experienced issues with some custom `node_security_group_additional_rules` because I missed to remove all duplicates that are now part of `node_security_group_enable_recommended_rules`.

I think this should be listed in the incompatible changes, as it breaks the apply run if you miss to remove the duplicate rules.
The error you might receive will look like the following:
```
Error: [WARN] A duplicate Security Group rule was found on (sg-xxx). This may be a side effect of a now-fixed Terraform issue causing two security groups with identical attributes but different source_security_group_ids to overwrite each other in the state. See https://github.com/hashicorp/terraform/pull/2376 for more information and instructions for recovery. Error: InvalidPermission.Duplicate: the specified rule "peer: sg-xxx, TCP, from port: 8443, to port: 8443, ALLOW" already exists status code: 400, request id: xxx
with module.eks.aws_security_group_rule.node["ingress_cluster_8443_webhook"]
on eks/node_groups.tf line 207, in resource "aws_security_group_rule" "node":
```

## Breaking Changes
no breaking change

## How Has This Been Tested?
only changes the docs


bryantbiggs (approved):

antonbabenko:
This PR is included in version 19.10.1 :tada:

PR #2502, opened 2023-03-03
fix: Fix for_each limitation on the iam role resources at node groups instances

nathaclmpaulino:
## Description
This PR changes the object that interacts with the `for_each` statement to avoid the limitation on it, described on the `Limitations on values used in for_each` section at for_each [docs](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each#limitations-on-values-used-in-for_each) on Terraform.

For that, instead of using a previous `set` type, a `map` type was used with hardcoded values for the keys on that map, avoiding this limitation and allowing one single `terraform apply` to perform all changes on the resources.

## Motivation and Context

This PR is related to issue #2458. 

Currently, we have a limitation in the `for_each` statement on Terraform, which is the keys of the map (or all sets of strings) must be known values, or you will get an error message that `for_each` has dependencies that cannot be determined before apply, and a `-target` may be needed.

On this specific issue, this behavior is only noticed using Self-Managed Node Groups, but that can be noticed using EKS Managed Node Groups because of the way the IAM roles are managed in them, so this PR solves both situations.

## Breaking Changes

This will trigger a recreation of the IAM roles attached to the EKS Managed Node Groups and Self-Managed Node Groups because this PR changes the way these resources are named. To avoid that we can use the `terraform state mv` command to rename the resources and fix the problem.

## How Has This Been Tested?

I have tested using a use case that I discussed in issue #2458. This is a use case in the company that I work for. I can't put the entire code here (NDA purposes), however, I'll be glad to tweak some of the current examples and post them as a new test case. What do you think?

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -398,6 +398,13 @@ locals {
   iam_role_name          = coalesce(var.iam_role_name, "${var.name}-eks-node-group")
   iam_role_policy_prefix = "arn:${data.aws_partition.current.partition}:iam::aws:policy"
   cni_policy             = var.cluster_ip_family == "ipv6" ? "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:policy/AmazonEKS_CNI_IPv6_Policy" : "${local.iam_role_policy_prefix}/AmazonEKS_CNI_Policy"
+
+  iam_role_policy_map = merge({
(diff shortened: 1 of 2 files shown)


bryantbiggs:
lets stick to the issue first - these changes are not warranted nor do they have the intended influence as you describe

ryanpeach:
Why would moving them to the locals help?

ryanpeach:
For the record using this guys fork resolved the issue for me.

ryanpeach:
Sorry, I was wrong, this does not resolve the issue.

Issue #2411, opened 2023-01-19
Error: Invalid for_each argument for  eks_managed_node_group (create_iam_role = false)


VivienMoretto87:
## Description

When I try to create a new node_group eks with eks_managed_node_group (outside of the eks module ), when I apply the code : Terraform send me the following error : 

╷
│ Error: Invalid for_each argument
│ 
│   on .terraform/modules/airflow_k8s.eks_managed_node_group/modules/eks-managed-node-group/main.tf line 434, in resource "aws_iam_role_policy_attachment" "this":
│  434:   for_each = { for k, v in toset(compact([
│  435:     "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy",
│  436:     "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly",
│  437:     var.iam_role_attach_cni_policy ? local.cni_policy : "",
│  438:   ])) : k => v if var.create && var.create_iam_role }
│     ├────────────────
│     │ local.cni_policy is a string, known only after apply
│     │ local.iam_role_policy_prefix is a string, known only after apply
│     │ var.create is true
│     │ var.create_iam_role is false
│     │ var.iam_role_attach_cni_policy is true
│ 
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform
│ cannot determine the full set of keys that will identify the instances of this resource.
│ 
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place
│ apply-time results only in the map values.
│ 
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends
│ on, and then apply a second time to fully converge.
  

## Versions

- Module version : 19.5.1

- Terraform version: v1.3.7

- Provider version(s):
 registry.terraform.io/hashicorp/aws  : 4.50.0
## Reproduction Code [Required]

module "eks_managed_node_group" {
  source = "terraform-aws-modules/eks/aws//modules/eks-managed-node-group"
  version = "19.5.1"

  name            = "eks-ng_example"
  cluster_name    = "cluster_name"
  cluster_version = "1.24"

  subnet_ids = data.terraform_remote_state.network.outputs.private_subnets_ids
  
  cluster_primary_security_group_id = data.terraform_remote_state.k8s.outputs.cluster_primary_security_group_id
  vpc_security_group_ids = [data.terraform_remote_state.k8s.outputs.vpc_security_group_ids]
[...]

PR #2509, opened 2023-03-08
var.create && var.create_iam_role evaluated before the creation of the map not after

ryanpeach:
Description
This PR fixes the issue with the for_each expression in the code, where var.create_iam_role was unable to overwrite a broken for_each expression. It changes the code to evaluate user variables first, and then if they are true, evaluates the list comprehension.

Motivation and Context
The issue with the for_each expression in the code has been reported in multiple issues (#2337, #1753, and #2411), and this PR aims to fix it for users creating their own iam role. This change evaluates user variables first, and then if they are true, evaluates the list comprehension, which solves an issue if some locals are only known after apply. See full comment https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2458#issuecomment-1459407299

Breaking Changes
There are no breaking changes with this PR.

How Has This Been Tested?
- I tested on a local terraform deployment
- I ran pre-commit as requested

I apologize as I have no reproducible minimum for you to evaluate on, hopefully someone else can provide it.

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -431,11 +431,11 @@ resource "aws_iam_role" "this" {
 
 # Policies attached ref https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group
 resource "aws_iam_role_policy_attachment" "this" {
-  for_each = { for k, v in toset(compact([
+  for_each = var.create && var.create_iam_role ? { for k, v in toset(compact([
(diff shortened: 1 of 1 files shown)


bryantbiggs:
> I apologize as I have no reproducible minimum for you to evaluate on, hopefully someone else can provide it.

We will not accept changes unless someone can provide a reproduction first that demonstrates the issue

ryanpeach:
I understand the desire to have a reproduction, but I, unfortunately, have too much on my plate.

You are seeing a lot of issues on this one line and my change disables that line in a way the user can then fix it themselves.

PR #2522, opened 2023-03-16
feat: Trigger specific event for karpenter 

ivan-amity:
## Description

Added specific EC2 events from AWS health and EC2 State Changes 

## Motivation and Context

This is already used in AWS EKS Blueprints Karpenter AddOns. These changes will trigger only specific health events for EC2 and its state.

## Breaking Changes

No breaking changes

## How Has This Been Tested?

### DIFF
--- a/modules/karpenter/main.tf
+++ b/modules/karpenter/main.tf
@@ -235,6 +235,9 @@ locals {
       event_pattern = {
         source      = ["aws.health"]
         detail-type = ["AWS Health Event"]
+        detail = {
+          service = ["EC2"] # Specific Health Event EC2 Service
+        }
       }
     }
     spot_interupt = {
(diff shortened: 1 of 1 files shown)


modules/karpenter/main.tf
```diff
@@ -259,6 +262,9 @@ locals {
       event_pattern = {
         source      = ["aws.ec2"]
         detail-type = ["EC2 Instance State-change Notification"]
+        detail = {
```
bryantbiggs, line 265: This is not what Karpenter has and we follow their guidance https://github.com/aws/karpenter/blob/7bd5857e4df31855b975b3c6383b142d0cd29321/website/content/en/v0.27.0/getting-started/getting-started-with-eksctl/cloudformation.yaml#L133-L143

bryantbiggs (changes requested):
Thank you but I don't believe these changes are correct - you can refer to the CloudFormation template provided by the Karpenter project; that is what we follow here

PR #2530, opened 2023-03-21
feat: Define instance types on launch template to avoid node group creation and deletion when changing instance_types value

noony:
## Description

Replace aws_eks_node_group.instance_types by aws_launch_template.instance_requirements.allowed_instance_types.

If possible i'll need your help to test it.

## Motivation and Context

This will replace a node group creation, then old nodegroup deletion by a simple update of the launch template on the nodegroup

This is very useful because of this [issue](https://github.com/aws/containers-roadmap/issues/1636) and will prevent downtime for instance_types change use case.

## Breaking Changes

No breaking change it allow the same as before, but it will also allow specifying more than before like : 
```
instance_types = ["c5*", "c6*"]
```

aws [documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceRequirements.html) see AllowedInstanceTypes part

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/README.md
+++ b/modules/eks-managed-node-group/README.md
@@ -64,13 +64,13 @@ module "eks_managed_node_group" {
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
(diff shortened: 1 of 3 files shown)


bryantbiggs:
1. Why would this solve the linked issue?
2. What do you use for autoscaling the nodes in your data plane?

PR #2531, opened 2023-03-21
fix: Support path in roles for aws_auth

oreonl:
## Description
This change strips the path from any rolearns for workers/fargate profiles when adding them to the aws_auth configmap. 

## Motivation and Context
This enables nodes to authenticate to the cluster when an iam_role_path is used since the authenticator [doesn't support paths in the role](https://github.com/aws/containers-roadmap/issues/573).

## Breaking Changes
No breaking changes

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -494,7 +494,7 @@ locals {
   aws_auth_configmap_data = {
     mapRoles = yamlencode(concat(
       [for role_arn in local.node_iam_role_arns_non_windows : {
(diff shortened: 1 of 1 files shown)


bryantbiggs:
thank you for the PR! However, we are not going to be accepting this change at this time. It is a well known requirement of the `aws-iam-authenticator` and we don't want to take on string manipulation and any unintended side effects here

PR #2566, opened 2023-04-13
added prefix_separator variable to the self-managed-node-group module

pettinz:
## Description
This change adds the variable `prefix_separator` to the self-managed-node-group module. This variable set the separator to use between the prefix and the generated timestamp for resource names. It defaults to the previously hardcoded `-`.
The `prefix_separator` variable is used in the main module but it is missing in the internal modules. The usage is the same.

## Motivation and Context
The prefix separator for resources in the self-managed-node-group module is hardcoded and set to `-`. Upgrading from v17, this leads to force replacements for the following resources, appending a `-` to the name prefix:
- aws_launch_template
- aws_iam_instance_profile
- aws_autoscaling_group

This change allows to keep the naming conventions for existing resources without a create replacement and then destroy.

## Breaking Changes
No breaking changes found during testing. If not specified, it does not involve any change to the previous configuration.

## How Has This Been Tested?
It is an optional variable to add to the self-managed-node-group module, or to the variables `self_managed_node_group_defaults` or `self_managed_node_groups` items of the main module.


bryantbiggs:
thank you but we are not looking to add this into the module

Issue #2275, opened 2022-10-18
Provider default_tags are missing for `eks_managed_node_group/aws_launch_template`


nantiferov:
## Description

Please provide a clear and concise description of the issue you are encountering, and a reproduction of your configuration (see the `examples/*` directory for references that you can copy+paste and tailor to match your configs if you are unable to copy your exact configuration). The reproduction MUST be executable by running `terraform init && terraform apply` without any further changes.

If your request is for a new feature, please use the `Feature request` template.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 18.30.0

- Terraform version: v1.2.9
- Provider version(s): v4.32.0

## Reproduction Code [Required]

```hcl
provider "aws" {
  default_tags { tags = { foo="bar" } }
}

module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "18.30.0"
....
  eks_managed_node_groups = {
  ...
  }
}

```

Try to use default_tags with aws_eks_node_groups like in example and worker nodes will miss default_tags from provider.

## Expected behavior

Tags from provider default_tags will be assigned to all created resources.

## Actual behavior

`eks_managed_node_group/aws_launch_template` is missing them and currently we have to hot fix it by defining the same tag with

```hcl
  eks_managed_node_group_defaults = {
   launch_template_tags = { foo="bar" }
}
```

### Terminal Output Screenshot(s)

None

## Additional context

Similar to https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2214

bryantbiggs:
we will not be providing any additional support for workarounds regarding provider tags in this module. Provider tags are meant to work at the provider level and the module should not be concerned with their usage. We offer a variable that is similar in nature, `var.tags`, but we are not able to support ensuring that the provider tags functionality works as it was intended. Please file an issue with the upstream provider (or better yet, raise your concern on one of the many open issues regarding the provider tags https://github.com/hashicorp/terraform-provider-aws/issues?q=is%3Aissue+is%3Aopen+default+tags+sort%3Areactions-%2B1-desc).

Attempting to provide minimal support for workarounds has already proven problematic https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2264 and will be reverted in v19

nantiferov:
It is fixed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2276 and I don't see any issues with this.

[...]

Issue #2214, opened 2022-08-31
Idempotency issue when using Provider tags


gjkamstra:
## Description

I am using provider tags as described in http://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-configuration-block. These tags don't get applied to the security groups, since they get created by EKS and not Terraform. Therefore I replicate these tags on the cluster_tags variable of the module. I do that by retrieving them from the provider settings using the aws_default_tags datasource. 
However when i do that, `terraform apply` keeps telling me that it needs to update the tags. This is due to the `cluster_tags` getting merged with `tags` and then set on the `eks` resource. 

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]:
18.28.0
- Terraform version:
Terraform v1.2.7
on darwin_amd64

- Provider version(s):
+ provider registry.terraform.io/hashicorp/aws v4.27.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.12.0
+ provider registry.terraform.io/hashicorp/null v3.1.1
+ provider registry.terraform.io/hashicorp/tls v3.4.0

## Reproduction Code [Required]

Example: https://github.com/terraform-aws-modules/terraform-aws-eks/compare/master...gjkamstra:terraform-aws-eks:provider-tag-example 

Execute `terraform apply` and say yes.
Execute it again, Terraform will say that there are still updates required to the tags. It will keep doing that

## Expected behavior

No changes required, after we just did `terraform apply`

## Actual behavior

Tag changes need to be applied

### Terminal Output Screenshot(s)
````
plan
module.eks.aws_eks_cluster.this[0] will be updated in-place
  ~ resource "aws_eks_cluster" "this" {
        id                        = "gjktest-cluster"
        name                      = "gjktest-cluster"
      ~ tags                      = {
          + "Environment"                            = "MyFancyEnvironment"
            # (1 unchanged element hidden)
        }
        # (11 unchanged attributes hidden)

        # (4 unchanged blocks hidden)
    }
````
Expanded plan:
````
{
    "change": {
        "actions": [
            "update"
        ],
        "before": {
            .......
            "tags": {
                "karpenter.sh/discovery/gjktest-cluster": "gjktest-cluster"
            },
            "tags_all": {
                "Environment": "MyFancyEnvironment",
[...]

Issue #2079, opened 2022-05-18
default_tags bug Error: Provider produced inconsistent final plan


barantomasz83:
## Description
Hi,
I am migrating my stack from version 17 to 18. Currently I am using 'default_tags' for provider to set tags for each resource, like below, and it works fine:
provider "aws" {
  alias  = "ireland"
  region = "eu-west-1"
  default_tags {
    tags = merge(local.AWS_DEFAULT_RESOURCES_TAGS, var.AWS_RESOURCES_TAGS)
  }
}

In version 18 I get: "│ Error: Provider produced inconsistent final plan" for each defined tag

When I tried to set the same tags directly in eks module and overwrite default one I got  :Error: "tags" are identical to those in the "default_tags" configuration block of the provider"

I found solution by overwriting default tags and adding one extra directly in eks module but I think it should be repaired in future versions.

![image](https://user-images.githubusercontent.com/67588485/169007384-03415697-95d7-4ad1-84dc-0c08f61ab415.png)

## Versions
- Module version [Required]: "18.21.0"

- Terraform version: v1.1.9

- Provider version(s): 4.14.0

## Reproduction Code [Required]
provider "aws" {
  alias  = "ireland"
  region = "eu-west-1"
  default_tags {
    tags = merge(local.AWS_DEFAULT_RESOURCES_TAGS, var.AWS_RESOURCES_TAGS)
  }
}
locals {
  AWS_DEFAULT_RESOURCES_TAGS = {
    TERRAFORM                     = "true"
    EKS_CLUSTER_NAME              = module.eks_cluster_first.eks_cluster_name
  }
}

1. set default_tags for aws provider
2. create eks cluster with self managed NG

## Actual behavior

│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for module.eks_cluster_first.module.eks.module.self_managed_node_group["mixed"].aws_security_group.this[0] to include new values learned so far during apply, provider                                                                                                                                                                               provider
│ "registry.terraform.io/hashicorp/aws" produced an invalid new value for .tags_all: new element "TERRAFORM" has appeared.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

### Terminal Output Screenshot(s)

![image](https://user-images.githubusercontent.com/67588485/169001941-1ea3bde1-c1c8-41e0-8b3c-fd04f3676781.png)

bryantbiggs:
[...]

Issue #1651, opened 2021-10-19
Cannot connect to cluster after adding "default_tags" to aws provider


gligaTrickest:
## Description

Terraform resources were already existing so this was an update, the only thing that I changed was adding `default_tags` to the `aws` provider like so:
~~~diff
 provider "aws" {
   region = var.region
+  default_tags {
+    tags = {
+      Env = "dev"
+    }
+  }
 }
~~~
What happened after this is bunch of cluster connection issues like:
~~~
│ Error: Get "http://localhost/api/v1/namespaces/kube-system/configmaps/aws-auth": dial tcp 127.0.0.1:80: connect: connection refused
│ 
│   with module.cluster.module.eks.kubernetes_config_map.aws_auth[0],
│   on .terraform/modules/cluster.eks/aws_auth.tf line 63, in resource "kubernetes_config_map" "aws_auth":
│   63: resource "kubernetes_config_map" "aws_auth" {
~~~

**When I remove `default_tags` everything works.**

## Versions
Terraform v1.0.7
on darwin_arm64
+ provider registry.terraform.io/gavinbunney/kubectl v1.13.0
+ provider registry.terraform.io/hashicorp/aws v3.63.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/helm v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.6.0
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
+ provider registry.terraform.io/kbst/kustomization v0.5.0
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1

`terraform-aws-modules/eks/aws` : `17.22.0`

## Reproduction

Add `default_tags` to the `aws` provider.

### Code Snippet to Reproduce
~~~diff
 provider "aws" {
   region = var.region
+  default_tags {
+    tags = {
+      Env = "dev"
+    }
+  }
 }
~~~

## Expected behavior
Plan to pass.

## Actual behavior
Bunch of failed cluster failed connection errors.

PR #2577, opened 2023-04-17
feat: Add variable to set custom tags on cluster primary SG

jnonino:
## Description

The purpose of this change is to add the possibility to set additional tags to the cluster primary security group.

## Motivation and Context

With version 19, `aws_default_tags` were removed because it may lead to update conflicts and should be managed by the user in the provider configuration. I agree with this and in the projects where I work, we are setting all the default tags in the provider. But when upgrading from version 18 to version 19 this cause the deletion of several tags from the cluster primary security group that are desired to have.

Exploring the code, I found this:
```
resource "aws_ec2_tag" "cluster_primary_security_group" {
  # This should not affect the name of the cluster primary security group
  # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2006
  # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2008
  for_each = { for k, v in merge(var.tags, var.cluster_tags) :
    k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags && v != null
  }

  resource_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
  key         = each.key
  value       = each. Value
}
```

Considering we are setting default tags in the provider, with the previous configuration, the cluster primary security group was getting the default tags. Now, it is only receiving tags from `var.tags` and `var.cluster_tags` so if I want the cluster primary security group to have the default tags, I only can do it using those variables because this SG is not created by the module it does not get the tags from the provider. The problem of using `var.tags` and `var.cluster_tags` is that those variables are also applied to the other resources in the module, causing terraform to throw this error.

```
Error: "tags" are identical to those in the "default_tags" configuration block of the provider: please de-duplicate and try again
```

With this new variable it would be possible to assign tags to the cluster primary security group only, without impacting other resources.

## Breaking Changes

None known.

## How Has This Been Tested?

- [x] I have tested and validated these changes using one my own cluster


ivankatliarchuk:
Question to maintainers. Any reason why default tags `data.aws_default_tags.current.tags` no longer supported for primary secruity group? 

Screenshot 2023-04-17 at 18 26 32

ivankatliarchuk:
@bryantbiggs I got your point. But this security group by EKS, how you expect `terraform-aws-provider` to propagate provider tags? There is no way to propagate down default tags for this particular resoucrce.

bryantbiggs:
I don't know what you mean by `For exact reason there is a line`

But you can add any tags you would like via the `cluster_tags` variable

ivankatliarchuk:
I mean, that the resource is created outside the `terraform-aws-provider`, event given the fact that Terraform requested EKS creation. For example resource https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster send an API to request a cluster creation only. EKS in the background creates this primary SG.

ivankatliarchuk:
I might not explaining correctly, but the reason why the resource exists
```
resource "aws_ec2_tag" "cluster_primary_security_group" { 
  for_each = { for k, v in merge(var.tags, var.cluster_tags, data.aws_default_tags.current.tags) :
    k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags
  }

  resource_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id
  key         = each.key
  value       = each.value
}
```

Is because
- SG is implicilty created outside of Terraform, and the only way to attach any tags to it is to reference a resource and attach all the tags one buy one.
- The resource `aws_ec2_tag` manages individual resource tags, and is imposible to combine `default_tags` and `aws_ec2_tag`

bryantbiggs:
correct - I added this resource so that tags are propagated to the security group that the EKS service created since Terraform did NOT create it. if you wish to add any additional tags to this security group you can provide them via `cluster_tags` (or more generically, `tags`)

demolitionmode:
A rough example of using the `cluster_tags` variable would be:

```terraform
provider "aws" {
  default_tags {
    tags = {
      team = "foobar"
    }
  }
}

module "eks" {
  ...
  cluster_tags = {
    team = "foobar" # This causes a conflict with default_tags
  }
}
```

The default_tags configuration in the provider is used to tag all resources with `team: "foobar"`, which also applies to the `aws_eks_cluster.this` resource; it does not apply to the primary security group however as that is not provisioned via the aws provider. Due to the conflicting tags, they cannot be used in conjunction to apply the same tags.

I believe this is where the below error arises

> Error: "tags" are identical to those in the "default_tags" configuration block of the provider: please de-duplicate and try again

bryantbiggs:
again - we are not taking on any additional work to mitigate the current deficiencies of the provider default tags and instead you should defer to the upcoming changes https://github.com/hashicorp/terraform-provider-aws/issues/29747

demolitionmode:
Specifically the `Have to ability to define identical tags that are overwritten on the resource level` item, from your perspective is that the expected change that would/should solve this issue?

ivankatliarchuk:
I'll leave it here if someone is going to have a similar issue.

The [case](https://github.com/hashicorp/terraform-provider-aws/issues/29747) is only partially related. How you even expect a resource `aws_ec2_tag` to support provider level `default_tags` for resources created outside of a terraform?

Example tagging, in yaml for simplicity

```
default_tags:
  owner: paas
  managed_by: terraform

tags:
  service: kubernetes

cluster_tags:
  name: cluster-01
```

With the input above, SG will only get following tags set

```
sg_tags:
  service: kubernetes
  name: cluster-01
```

If we decide to merge default and cluster tags, below input is not going to work at least in meantime. There is a plan to support duplicates, but why would we even duplicate tags.

```
default_tags:
  owner: paas
  managed_by: terraform

tags:
  service: kubernetes

cluster_tags:
  owner: paas
  managed_by: terraform
  name: cluster-01
```

So the options left
- Attach provider `default_tags`to SG tags outside of a module
- Do not use `default_tags`
- Wait of a `terraform-aws-provider` issue resolution and duplicate TAG(s)

bryantbiggs:
> Specifically the Have to ability to define identical tags that are overwritten on the resource level item, from your perspective is that the expected change that would/should solve this issue?

Correct

> How you expect a resource aws_ec2_tag to support provider level default_tags for resources created outside of a terraform?

I don't expect anything - I am simply working with what exists. The EKS service creates this security group and its outside of Terraform's control - which is why the tag resource was added. If you wish to add tags to this security group, you can use `tags` or `cluster_tags`. I am NOT trying to make provider tags feel transparent to all the resources that are not under Terraform's control

jnonino:
Regarding the last question you answered, currently there is no feature in AWS provider to overwrite tags and avoid error because of duplication.
```
Error: "tags" are identical to those in the "default_tags" configuration block of the provider: please de-duplicate and try again
```

You removed all references of `aws_default_tags` to avoid update conflicts; this is the responsibility of the provider and should be handled at the provider level and we agree with that, that's why we use the provider to set default tags. But given the cluster primary security group is not created by Terraform in your module, it does not get the default tags. Your suggestion to use `tags` or `cluster_tags` will add the tags for the cluster primary SG but it will also add the tags to all the other resources, therefore is not suitable when you you have default tags coming from the provider because it lead to the error described above.

So you say you are simply working with what exists, but you are actually working based on a provider feature that may be released sometime in the future and we are not sure it will actually solve the problem. And even when that solution solved the duplication error it will be weird to have default tags in the provider and at the same time set the variable `tags` or `cluster_tags` with the same default tags coming from the provider just for the module to assign the tags to the cluster primary SG.

bryantbiggs:
And yet you say this proposed change in your PR will work ...

- https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/UPGRADE-19.0.md#removed
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1455
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2275
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1651
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1578
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2079
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1916
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2214
- https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2264

jnonino:
I actually tested it, it works for the intended purpose, you can set whatever tags you want to the cluster primary group. If you check my changes, I'm not doing anything with the provider tags or anything like that. I just added a variable so you can add tags to the cluster primary SG not using `tags` or `cluster_tags` because those are applied to all the other resources.

Does this change needs to be there forever? No, once the provider supports to overwrite tags instead of throwing a duplication error this variable can be removed or refactored or have a discussion again. But now, the situation is, I have a terraform project using your module but also creating other resources and I'd like to have every resource tagged with some default tags, how do we do that? setting default tags in the provider. But resources not created by terraform, like the cluster primary SG are not picking up those tags for obvious reasons. And the solutions you are giving me to add the default tags to those resources too are:
1. Use the variables `tags` or `cluster_tags`. But if I used those, I have a conflict with the default tags coming from the provider because of duplication.
2. Wait for a provider feature that would allow overriding tags, but it is not released yet.

In conclusion, your module is not providing any way to apply the default tags that are being applied to the rest of the resources to resources like the cluster primary SG. And I get it, that SG is created by EKS, not by terraform but it is indirectly created by your module because the SG gets created when you create an EKS cluster which is why we your module, to create EKS clusters.

PR #2249, opened 2022-09-29
feat: Add output for cluster TLS certificate SHA1 fingerprint and provider tags to cluster primary security group

bryantbiggs:
## Description
- Add output for cluster TLS certificate SHA1 fingerprint
- Add provider default tags to EKS created security group for tag stability
- Update user data variable description to clarify its usage with `ami_id`

## Motivation and Context
- Resolves #2227 
- Resolves #2214

## Breaking Changes
- No

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #2586, opened 2023-04-24
feat: Add variable cluster_primary_security_group_tags for cluster primary security group tags

balonik:
## Description
Add new input variable `cluster_primary_security_group_tags` to specifically set cluster primary security group tags.

## Motivation and Context
Alternative to https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2249 which was later reverted in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2250.

I am trying to apply AWS provider default tags on Cluster primary security group, I have used `cluster_tags` with `data.aws_default_tags` data source. But this is causing tags changes to be planned/applied every run, because of https://github.com/hashicorp/terraform-provider-aws/issues/18311 and fact that `cluster_tags` is also used in `aws_eks_cluster` resource. Using dedicated variable I can do something like:

`cluster_primary_security_group_tags = data.aws_default_tags.current.tags` instead of `cluster_tags = data.aws_default_tags.current.tags`.

Having a dedicated variable also helps in cases where tags on EKS cluster and SG must differ.

Next step would be to remove `var.cluster_tags` from the SG. 

## Breaking Changes
N/A

## How Has This Been Tested?


bryantbiggs:
as stated in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2577, this is not a change we are going to support at this time

PR #2603, opened 2023-05-09
feat: Root user master access

rblaine95:
## Description

Always give the `root` user of the current account `system:master` access.

## Motivation and Context

We had an issue where we were using AWS SSO (IAM Identity Center) assumed roles to access our EKS cluster.
We inadvertently caused the Assumable Roles to be destroyed and recreated which caused their names to be updated (`AWSReservedSSO_XXXX_11111` -> `AWSReservedSSO_XXXX_22222`) which caused us to lose all auth with the EKS cluster and get locked out.

We tried recovering access by using the Root user of the account, but that didn't work because the Root user didn't have permissions according to the `aws-auth` configmap.

## Breaking Changes

None

## How Has This Been Tested?

- [X] I have run this in my own environment on our EKS clusters and validated that it works as expected. 
I set the following config using `19.13.1` of this Terraform module:
```hcl
locals {
  eks_auth_users = [{
    userarn  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:user/robbie.blaine"
    username = "robbie.blaine"
    groups   = ["system:masters"]
  }]
}

module "eks" {
  version = "19.13.1"

  aws_auth_users = concat([{ # Always give `root` user master access
      userarn  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"
      username = "root"
      groups   = ["system:masters"]
    }], local.eks_auth_users)
}
```
And then validated the contents of the `aws-auth` configmap.

It may be better to use [`setunion()`](https://developer.hashicorp.com/terraform/language/functions/setunion) instead of `concat()`, but let me know what you think.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -528,7 +528,13 @@ locals {
       ],
       var.aws_auth_roles
     ))
-    mapUsers    = yamlencode(var.aws_auth_users)
+    mapUsers = yamlencode(concat([
+      { # Always give `root` user of current account `system:master` access
+        userarn  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"
+        username = "root"
+        groups   = ["system:masters"]
+      }
+    ], var.aws_auth_users))
     mapAccounts = yamlencode(var.aws_auth_accounts)
   }
 }


bryantbiggs:
thank you for the PR! However, this is a change best left to users rather than the module forcing it onto users. 

Also, I would watch https://github.com/aws/containers-roadmap/issues/185 as the recommended solution for this

PR #2615, opened 2023-05-14
docs: Update karpenter example, add SSM IAM policy for Karpenter startup debugging

bdols:
## Description
I've updated the karpenter version in the example, and the terrform-aws-modules/vpc/aws version.
Also, the karpenter helm install needed to have tolerations for the fargate compute-type in order for it to start successfully with the taints on the fargate instances.
And, repeated `terraform apply` were upgrading the helm chart with a new token from the ECR service, which wasn't necessary and slowed state applications, I found setting the registry credentials in the helm provider worked around this.
Finally, the Karpenter project recommends using SSM session manager to debug node startup (instead of setting up ssh key pairs for access) and the arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore policy is attached to Karpenter's IAM role in their setup docs for migrating from Cluster Autoscaler. I added that policy to iam_role_additional_policies so that Karpenter nodes would have SSM session manager access available.

## Motivation and Context
I found that the provided karpenter example didn't start without some changes.
I don't see any open issues along the lines of what I experienced.

## Breaking Changes
N/A

## How Has This Been Tested?

I verified that nodes started up by Karpenter have SSM management enabled with `aws ssm describe-instance-information --output text` and `aws ssm start-session --target i-[instance-id]`

- [ X ] I have tested and validated these changes using one or more of the provided `examples/*` projects

### DIFF
--- a/examples/karpenter/main.tf
+++ b/examples/karpenter/main.tf
@@ -31,6 +31,11 @@ provider "helm" {
       args = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
     }
   }
+  registry {
+    url      = "oci://public.ecr.aws/"
+    username = data.aws_ecrpublic_authorization_token.token.user_name
+    password = data.aws_ecrpublic_authorization_token.token.password
+  }
 }
 
 provider "kubectl" {
@@ -160,6 +165,9 @@ module "karpenter" {
   cluster_name           = module.eks.cluster_name
   irsa_oidc_provider_arn = module.eks.oidc_provider_arn
 
+  # add SSM access for debugging
+  iam_role_additional_policies = [ "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore" ]
+
   policies = {
     AmazonSSMManagedInstanceCore = "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
   }
@@ -173,10 +181,8 @@ resource "helm_release" "karpenter" {
 
   name                = "karpenter"
   repository          = "oci://public.ecr.aws/karpenter"
-  repository_username = data.aws_ecrpublic_authorization_token.token.user_name
-  repository_password = data.aws_ecrpublic_authorization_token.token.password
   chart               = "karpenter"
-  version             = "v0.21.1"
+  version             = "v0.27.3"
 
   set {
     name  = "settings.aws.clusterName"
@@ -201,6 +207,15 @@ resource "helm_release" "karpenter" {
   set {
     name  = "settings.aws.interruptionQueueName"
     value = module.karpenter.queue_name
+  }
+
+  set {
+    name  = "tolerations[0].key"
+    value = "eks.amazonaws.com/compute-type"
+  }
+  set {
+    name  = "tolerations[0].value"
+    value = "fargate"
   }
 }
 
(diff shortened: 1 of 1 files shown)


examples/karpenter/main.tf
```diff
@@ -31,6 +31,11 @@ provider "helm" {
       args = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
     }
   }
+  registry {
+    url      = "oci://public.ecr.aws/"
+    username = data.aws_ecrpublic_authorization_token.token.user_name
```
bryantbiggs, line 36: whether users want to manage the secrets via the provider or the resource is left up to them. we have shown in the resource because more than likely they are deploying other resources which will now fail using these ECR credentials

examples/karpenter/main.tf
```diff
@@ -160,6 +165,9 @@ module "karpenter" {
   cluster_name           = module.eks.cluster_name
   irsa_oidc_provider_arn = module.eks.oidc_provider_arn
 
+  # add SSM access for debugging
+  iam_role_additional_policies = [ "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore" ]
```
bryantbiggs, line 169: This is already attached below, so why is this necessary here?

bdols: yes this confused me too as I was reading the karpenter docs and wondered why I wasn't able to connect with the session manager.
`policies` in the karpenter module attaches, in the example, the SSM role to the service account, so, the karpenter deployment needs this to create nodes. I tried removing the policies setting and saw that karpenter was unable to provision new nodes until I added it back. In the [migration document](https://karpenter.sh/v0.27.3/getting-started/migrating-from-cas/), the helm template/install step annotates the serviceaccount via a helm value.  I don't know why the SSM role is attached in the example when, in the module, the irsa aws_iam_policy_document seems to be trying to mirror the KarpenterControllerRole seen in the migration document. 

For  the iam_role_additional_policies, I suppose it could be considered optional - it's not necessary for karpenter to run in my estimation, but it's the recommended way to [troubleshoot karpenter](https://karpenter.sh/v0.27.3/troubleshooting/#node-notready), with `aws ssm start-session`. In the same [migration document](https://karpenter.sh/v0.27.3/getting-started/migrating-from-cas/#create-iam-roles), the SSM policy is attached again to the NodeRole policy and the NodeRole policy is attached to a NodeInstanceProfile and the NodeInstanceProfile is passed in as another value to the karpenter helm chart.

I can make it optional/comment it out if you'd like, I found the example provided confusing when I wasn't able to connect with SSM while the policy was provided already in one place in the example.

but now I'm wondering why arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore isn't established in the module instead of being needed to the users of the module, if it is indeed necessary, and whether the module aws_iam_policy_document in the module has diverged from karpenter upstream.

examples/karpenter/main.tf
bdols: OK, if you want to proceed with this PR, I can remove this. In my case, I manage my helm installations with other tooling and found that the extra 5-10 second wait to upgrade the helm chart for an updated token was needless while I iterated on karpenter options. I really haven't used terraform long enough to know what the ramifications are of using a provider vs resource. Thank you for the feedback.

bryantbiggs:
closing for out for now

PR #2631, opened 2023-05-25
fix: Ensure `isra_tag_values` can be tried before defaulting to `cluster_name` on Karpenter module 

sbhrule15:
## Description
"Coalescelist" uses the first non-empty list. Since var.cluster_name defaults to an empty string, which creates a single-element array with an empty string as the only element (as well as it is almost always specified by the user for other resources such as the queue name and other cluster-named resources in this module) the custom tag values never take affect.

https://developer.hashicorp.com/terraform/language/functions/coalescelist

## Motivation and Context
Without this change, passing in the cluster name and specifying custom tag values is not possible. this change is necessary to be able to share resources between clusters. For example, we use the same set of subnets for multiple development clusters with karpenter.sh/discovery/{cluster_name} = "shared" on the subnets for discovery.

## Breaking Changes
This does not introduce a breaking change.

## How Has This Been Tested?
- I have manually installed this updated module into our playground account with success.


bryantbiggs (approved):
good catch, thank you!

antonbabenko:
This PR is included in version 19.15.2 :tada:

Issue #2635, opened 2023-05-30
Deprecated warning with new hashicorp/aws v5.0.1 provider


khosro-copper:
With latest hashicorp/aws v5.0.1 provider as well as latest terraform-aws-eks module, when I run terraform plan, I see this deprecated warning for aws addons:

   ```
Warning: Argument is deprecated

   with aws_eks_addon.this["coredns"],
   on main.tf line 388, in resource "aws_eks_addon" "this":
  388:   resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")

 The "resolve_conflicts" attribute can't be set to "PRESERVE" on initial
 resource creation. Use "resolve_conflicts_on_create" and/or
 "resolve_conflicts_on_update" instead

 (and one more similar warning elsewhere)
```
If update main.tf with recommended suggestion, it should work as I test it locally.
Thanks!

**Comment by vstthomas** [comment]
Sorry, I'm still seeing this post-v5+.

```hcl
Warning: Argument is deprecated

  with module.eks.aws_eks_addon.this["kube-proxy"],
  on .terraform/modules/eks/main.tf line 399, in resource "aws_eks_addon" "this":
 399:   resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")

The "resolve_conflicts" attribute can't be set to "PRESERVE" on initial resource creation. Use "resolve_conflicts_on_create" and/or "resolve_conflicts_on_update" instead

(and 2 more similar warnings elsewhere)
```

Applies to [L399](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/ec454c51cb0026254b68cafe7aa9d05f873af990/main.tf#L399) and [L427](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/ec454c51cb0026254b68cafe7aa9d05f873af990/main.tf#L427).

```sh
% tf version
Terraform v1.6.3
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v5.26.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.2
+ provider registry.terraform.io/hashicorp/kubernetes v2.23.0
+ provider registry.terraform.io/hashicorp/time v0.9.1
+ provider registry.terraform.io/hashicorp/tls v4.0.4
```

bryantbiggs:
thats because it hasn't been resolved in this module yet - it will be resolved once v20.x is released

[...]

PR #2637, opened 2023-06-01
fix: Deprecated argument resolve_conflicts

basigabri:
## Description
Bug-fix: WARNING The resolve_conflicts attribute can't be set to PRESERVE on initial resource creation.
Use resolve_conflicts_on_create and/or resolve_conflicts_on_update instead

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -386,11 +386,11 @@ resource "aws_eks_addon" "this" {
   cluster_name = aws_eks_cluster.this[0].name
   addon_name   = try(each.value.name, each.key)
 
-  addon_version            = try(each.value.addon_version, data.aws_eks_addon_version.this[each.key].version)
-  configuration_values     = try(each.value.configuration_values, null)
-  preserve                 = try(each.value.preserve, null)
-  resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")
-  service_account_role_arn = try(each.value.service_account_role_arn, null)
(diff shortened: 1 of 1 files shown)


MarcHenriot:
Thanks for the pr !

I would recommend adding both fields `resolve_conflicts_on_create` and `resolve_conflicts_on_update` in the `aws_eks_addon` resources 

Here is an example of add-on block :

```hcl
cluster_addons = {
    coredns = {
      resolve_conflicts_on_create = "OVERWRITE"
      resolve_conflicts_on_update = "OVERWRITE"
    }
}
```

in the `main.tf` :

```hcl
resource "aws_eks_addon" "this" {
...
  preserve                    = try(each.value.preserve, null)
  resolve_conflicts_on_create = try(each.value.resolve_conflicts_on_create , "OVERWRITE")
  resolve_conflicts_on_update = try(each.value.resolve_conflicts_on_update , "OVERWRITE")
  service_account_role_arn    = try(each.value.service_account_role_arn, null)
...
}
```

PS : I'm not sure, but maybe the `aws_eks_addon` resource `before_compute` has to be updated too.

bryantbiggs:
we will track this in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2635 - this is a breaking change so its not something taken lightly. its unfortunate that a warning has to be shown, but today the provider still supports both routes for setting these values

PR #2658, opened 2023-06-21
feat: Add additional log-group tags

WJay-tec:
## Description
1) added additional variable for cloudwatch_log_groups

## Motivation and Context
We have a use case where custom tags is needed to be created ONLY for our cloudwatch_log_groups, it is a tag that will be read by our application which runs periodically to determine whether the loggroup should be archived or not.

## Breaking Changes
No breaking changes is expected

- I have executed `pre-commit run -a` on my pull request


WJay-tec:
Hi @bryantbiggs can i get a review for this?

bryantbiggs:
thank you but we will not be adding additional variables for this functionality

PR #2670, opened 2023-06-28
Fix Permissions issue for spot instances

harsha-s:
Pods are not getting scheduled due to permissions issues  below

controller.provisioner launching machine, creating cloud provider instance, with fleet error(s), AuthFailure.ServiceLinkedRoleCreationNotPermitted: The provided credentials do not have permission to create the service-linked role for EC2 Spot Instances.

## Motivation and Context

Pods not getting scheduled due to permissions issue

## How Has This Been Tested?
- [ X] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)

### DIFF
--- a/examples/karpenter/main.tf
+++ b/examples/karpenter/main.tf
@@ -284,6 +284,10 @@ resource "kubectl_manifest" "karpenter_example_deployment" {
 # Supporting Resources
 ################################################################################
 
+resource "aws_iam_service_linked_role" "spot" {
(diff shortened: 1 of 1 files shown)


bryantbiggs:
yes, this can be rather frustrating. however, we can't add this here because there are a number of users which will have this role already in their account and that would cause a conflict. Whenever a service linked role is involved, its best to refer to the documentation for that one time creation event

PR #2674, opened 2023-07-02
Update documentation about enabling spot instances

harsha-s:
## How Has This Been Tested?


bryantbiggs:
thank you but I think instead we will rely on the Karpenter documentation since this is not specific to the module

PR #2680, opened 2023-07-11
fix: Split deprecated resolve_conflicts to resolve_conflicts_on_create/update

johnbr:
## Description
Deprecation warnings around resolve_conflicts, change to resolve_conflicts_on_create and resolve_conflicts_on_update

## Motivation and Context
â Warning: Argument is deprecated
â 
â   with module.eks_cluster.module.eks.aws_eks_addon.this["kube-proxy"],
â   on .terraform/modules/eks_cluster.eks/main.tf line 392, in resource "aws_eks_addon" "this":
â  392:   resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")
â 
â The "resolve_conflicts" attribute can't be set to "PRESERVE" on initial resource creation. Use "resolve_conflicts_on_create" and/or "resolve_conflicts_on_update" instead

## Breaking Changes
Unknown!

## How Has This Been Tested?
I have tested these changes in my current environment and deprecation warnings no longer present.

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 4.47"
+      version = ">= 5.0.0"
     }
     tls = {
       source  = "hashicorp/tls"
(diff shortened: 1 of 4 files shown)


versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 4.47"
+      version = ">= 5.0.0"
```
bryantbiggs, line 7: this is a breaking change

bryantbiggs:
this will be resolved at the next breaking change

rafilkmp3:
So @bryantbiggs  in the next update of the module, the deprecation notice will be fixed? or do I need to change anything on our side ?

bryantbiggs:
you can pin the version of the AWS provider you use to < 5.0 if you are not using any features in 5.0. Unfortunately, Terraform does not provide a way to silence warnings at this time. Both methods are still perfectly valid, the warning is just stating that `resolve_conflicts` will be removed in v6.0 of the AWS provider which would be roughly a year from now

PR #2697, opened 2023-07-21
fix(versions): Avoid deprecated resolve_conflicts warning by restricting AWS provider version to <5.0

TPXP:
## Description
Since we don't want to bump the AWS provider version we're using, let's add a constraint the other way to avoid the warning.

ref: #2680

## Motivation and Context
The latest version of this module does not set an upper bound to the AWS provider version it accepts, which means v5 of the provider can be used, causing the following warning to show up when using EKS addons:

```
╷
│ Warning: Argument is deprecated
│
│   with aws_eks_addon.this["vpc-cni"],
│   on main.tf line 392, in resource "aws_eks_addon" "this":
│  392:   resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")
│
│ The "resolve_conflicts" attribute can't be set to "PRESERVE" on initial
│ resource creation. Use "resolve_conflicts_on_create" and/or
│ "resolve_conflicts_on_update" instead
│
│ (and 2 more similar warnings elsewhere)
╵
```

## Breaking Changes
None, this is a bugfix

## How Has This Been Tested?

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 4.47"
+      version = ">= 4.47, < 5.0"
     }
     tls = {
       source  = "hashicorp/tls"


liviudm:
Why not update the module to use `resolve_conflicts_on_create` and `resolve_conflicts_on_create` instead of restricting the provider version?

TPXP:
@liviudm This has been suggested here, and rejected because it would be a breaking change https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2680

bryantbiggs:
Please see https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2680#issuecomment-1634633459

We will resolve this warning at the next breaking change

PR #2698, opened 2023-07-21
fix: Updated the examples based on change-doc for v19

shashanktyagi3192:
## Description

As per the Updating to v19.x document [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/UPGRADE-19.0.md), the `block_device_mappings` object has been changed from map of map to array of maps but the examples were not updated to reflect the change.
>block_device_mappings previously required a map of maps but has since changed to an array of maps. Users can remove the outer key for each block device mapping and replace the outermost map {} with an array []. There are no state changes required for this change.
## Motivation and Context

This change helps correct the example implementation for Update version 19.x. Currently, the example `block_device_mapping` is misleading and needs to be fixed.

## Breaking Changes

This is not a breaking change

## How Has This Been Tested?

### DIFF
--- a/examples/eks_managed_node_group/main.tf
+++ b/examples/eks_managed_node_group/main.tf
@@ -230,8 +230,8 @@ module "eks" {
       disable_api_termination = false
       enable_monitoring       = true
 
-      block_device_mappings = {
-        xvda = {
+      block_device_mappings = [
+        {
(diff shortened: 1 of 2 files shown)


bryantbiggs:
either form is acceptable, so this is still valid

PR #2716, opened 2023-08-09
Oscarmendoza123 patch 1

oscarmendoza123:
## Description
Adding the same option as IAM roles to optionally use a name prefix. The will default to the same way the module is being used today but will give end users the option to not use the prefix.

## Motivation and Context
` Member must have length less than or equal to 64` error when trying to use this currently

## Breaking Changes
NA

## How Has This Been Tested?
- [x ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x ] I have tested and validated these changes using one or more of the provided `examples/*` projects
- [x ] I have executed `pre-commit run -a` on my pull request

### DIFF
--- a/modules/karpenter/main.tf
+++ b/modules/karpenter/main.tf
@@ -279,7 +279,9 @@ locals {
 resource "aws_cloudwatch_event_rule" "this" {
   for_each = { for k, v in local.events : k => v if local.enable_spot_termination }
 
-  name_prefix   = "${var.rule_name_prefix}${each.value.name}-"
+  name        = var.rule_use_name_prefix ? null : "${var.rule_name_prefix}${each.value.name}"
+  name_prefix = var.rule_use_name_prefix ? "${var.rule_name_prefix}${each.value.name}-" : null
+
   description   = each.value.description
   event_pattern = jsonencode(each.value.event_pattern)
 
--- a/modules/karpenter/variables.tf
+++ b/modules/karpenter/variables.tf
@@ -246,6 +246,11 @@ variable "create_instance_profile" {
 ################################################################################
 # Event Bridge Rules
 ################################################################################
+variable "rule_use_name_prefix" {
+  description = "Determines whether the Cloud Watch Event name (`rule_name_prefix`) is used as a prefix"
+  type        = bool
+  default     = true
+}
 
 variable "rule_name_prefix" {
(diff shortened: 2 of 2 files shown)


bryantbiggs:
we do no want to remove the prefix option because it is susceptible to naming collisions. You should be able to manage this with `rule_name_prefix` instead

oscarmendoza123:
I am not removing the prefix option? I am keeping it as the default actually. And I think it should be up to the users if they want to use it or not in the same way IAM role and instance profile use it in this same exact module...

oscarmendoza123:
Also, rude you closed the PR without having a conversation first. I am using `rule_name_prefix` but it only works if you prefix is under a small amount of characters. For small teams using AWS Organizations who have one cluster per account, collations are not an issue but naming standards are.

bryantbiggs:
Happy to have a conversation through an issue first with outlines the issue, why its relevant, etc. Theres also git history which has a plethora of information, including why some changes exist (i.e. - `rule_name_prefix`)

oscarmendoza123:
I'm not arguing the need for `rule_name_prefix` at all... I'm leaving that as the default so users will see no change. I'm adding functionality that already exists for other resources in this same module. Why not just remove the option for `name` everywhere and only use `name_prefix`?... becuase you want to give users the flexibility which is what I am adding.

FernandoMiguel:
> Why not just remove the option for `name` everywhere and only use `name_prefix`?
some folks seem to hate "random names"... i dont get it either

bryantbiggs:
> I'm not arguing the need for rule_name_prefix at all... I'm leaving that as the default so users will see no change. I'm adding functionality that already exists for other resources in this same module. Why not just remove the option for name everywhere and only use name_prefix?... becuase you want to give users the flexibility which is what I am adding.

there are a lot of factors that go into decisions and unfortunately its hard to balance between taking the time to completely explain the history, rationale, etc., when weighed against statements of "I'm giving users flexibility". Therefore, I will always approach PRs that lack context, motivation, error messages with the stance of - what problem are we trying to solve, why is this a problem worth solving, etc. 

another area that we get beat up on modules is their complexity - people view the number of variables and instantly perceive this as overly complex and "scary" even though only a handful of variables are required or will ever be used by users. in short - we need a strong motivational factor to consider a change

PR #2727, opened 2023-08-23
Fixing errors while creating new EKS cluster Karpenter / Fargate profiles

maordavidov:
## Description
Create separate role policy attachments instead of using for_each in Karpenter and Fargate.

## Motivation and Context
While creating a new EKS cluster and using the modules of Karpenter / Fargate separately from the EKS cluster module, there is an error that the variable `local.iam_role_policy_prefix` is unknown and you should apply some resources first using `-target` to resolve the unknown dependency.

The error is because the for_each phase happens during plan and not during apply, while the resource will be resolved during apply.... therefore the removal of for_each for those 3 resources (which is also simplifying the code in my perspective) resolve the issue.

Screenshot 2023-08-23 at 13 23 31

## Breaking Changes
No breaking change

## How Has This Been Tested?

### DIFF
--- a/modules/fargate-profile/main.tf
+++ b/modules/fargate-profile/main.tf
@@ -40,13 +40,15 @@ resource "aws_iam_role" "this" {
   tags = merge(var.tags, var.iam_role_tags)
 }
 
-resource "aws_iam_role_policy_attachment" "this" {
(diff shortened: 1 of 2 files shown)


bryantbiggs:
please start with an issue that provides the information requested in the issue template

PR #2729, opened 2023-08-28
fix: Clarify space for additional security sgs on nodes

joseph-netog:
## Description
making a map for managed nodes to add additional sgs is confusing

just want it to be a clear place where the additional sg ids go.

## Motivation and Context
Its confusing why a map is needed for managed node sg ids when the cluster has a top level variable for the same reason.

## Breaking Changes
The only place i found that uses this map field `var.eks_managed_node_group_defaults.vpc_security_group_ids`

is the concat operation to add the ids to the final security group list.

## How Has This Been Tested?


bryantbiggs:
this is not correct - the current implementation is the correct way for adding additional security groups

Issue #2751, opened 2023-09-19
Allow excluding the cluster security group from the list of additional security groups.


tnils:
## Is your request related to a problem? Please describe.
I am importing EKS clusters created with eksctl. In my existing clusters, the cluster security group is not on the list of additional security groups. I can't import my existing EKS cluster without it being forced to recreate because the module forces the cluster security group to be part of the additional security groups.

## Describe the solution you'd like.
Adding a variable like `cluster_additional_security_group_ids_include_cluster_sg_id` which defaults to `true` to keep the current behavior, but allowing it to be set to `false` so that the cluster security group id is not added to the list of additional security group ids.

## Describe alternatives you've considered.
The only alternatives are to not use the module, or recreate my cluster.

bryantbiggs:
Can you provide a reproduction? This doesn't quite make sense but seeing a repro will help clarify

b-tanner:
I just opened https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2752, maybe that will help it make more sense.

I'm using `terraform import` to import and existing EKS cluster. In my existing cluster, the Cluster Security Group is not in the list of Additional Security Groups:

![image](https://github.com/terraform-aws-modules/terraform-aws-eks/assets/141063987/6edf27fd-ffbd-4918-8427-468dfe77fe08)

As the module is currently, it adds the cluster security group to the list of additional security groups, but this change forces a replacement when running `terraform apply`. The submitted PR lets a variable `cluster_additional_security_group_ids_exclude_cluster_sg_id` be set to `true`, so the list of additional security groups will not have the cluster security group added.

bryantbiggs:
which cluster security group - there is the primary security group which the EKS service creates automatically, and this module creates on additional security group which allows users to control the security groups rules

If you do not want to have this 2nd cluster security group created, you can set `create_cluster_security_group = false`

b-tanner:
When my cluster was initially created with eksctl (`eksctl create cluster --name <name>`), it created two security groups. When viewing my cluster in the console, one of them is listed under Cluster Security Group, and one of them is under Additional Security Groups, as shown in the screenshot above.

I was hoping to be able to `terraform import` all my existing EKS resources using this module with module variables set such that when running `terraform apply` there would be few changes, and nothing replaced. 

[...]

PR #2752, opened 2023-09-20
feat: Add cluster_additional_security_group_ids_exclude_cluster_sg_id…

b-tanner:
… for excluding the cluster security group id from the list of additional security groups.

## Description
Adds a variable cluster_additional_security_group_ids_exclude_cluster_sg_id that defaults to false. When set to true, the list of additional security groups added to the eks cluster will not include the cluster security group. By default, the cluster security group is included.

## Motivation and Context
This is useful in the scenario where an existing cluster is being imported with `terraform import` and the cluster security group was not included in additional security groups. Without this option, the cluster has to be recreated because the module tries to change the list of additional security groups, which forces a replacement

Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2751

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -31,7 +31,7 @@ resource "aws_eks_cluster" "this" {
   enabled_cluster_log_types = var.cluster_enabled_log_types
 
   vpc_config {
-    security_group_ids      = compact(distinct(concat(var.cluster_additional_security_group_ids, [local.cluster_security_group_id])))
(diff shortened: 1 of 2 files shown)


bryantbiggs:
lets stick to the issue before we start opening PRs

PR #2759, opened 2023-09-29
feat: Support addon resolve_conflicts on_create/on_update

johanneswuerbach:
## Description

Support the new `resolve_conflicts_on_create` and `resolve_conflicts_on_update` addon properties in addition to keep `resolve_conflicts` support.

The implemented logic:
* If nothing is set, set `resolve_conflicts_on_create=OVERWRITE ` and `resolve_conflicts_on_update=OVERWRITE` (removes the deprecation warning).
* If `resolve_conflicts` is set, forward it to the resource and don't set the `OVERWRITE` defaults (to prevent the property conflict).
* If `resolve_conflicts_on_create` is set forward to the resource
* If `resolve_conflicts_on_update` is set forward to the resource
* If `resolve_conflicts` and `resolve_conflicts_on_*` is set, forward to the resource, but this would result in an error from the aws provider as setting those properties at the same time conflicts.

## Motivation and Context

Fixes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2635

## Breaking Changes

No, it shouldn't.

## How Has This Been Tested?

I've tested different property combinations in `examples/eks_managed_node_group` and check the resulting terraform plan.

### DIFF
--- a/docs/irsa_integration.md
+++ b/docs/irsa_integration.md
@@ -12,8 +12,9 @@ module "eks" {
 
   cluster_addons = {
     vpc-cni = {
-      resolve_conflicts        = "OVERWRITE"
-      service_account_role_arn = module.vpc_cni_irsa.iam_role_arn
+      resolve_conflicts_on_create = "OVERWRITE"
+      resolve_conflicts_on_update = "OVERWRITE"
+      service_account_role_arn    = module.vpc_cni_irsa.iam_role_arn
     }
   }
 
--- a/main.tf
+++ b/main.tf
@@ -386,11 +386,13 @@ resource "aws_eks_addon" "this" {
   cluster_name = aws_eks_cluster.this[0].name
   addon_name   = try(each.value.name, each.key)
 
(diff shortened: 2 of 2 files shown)


bryantbiggs:
please see https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2635#issuecomment-1572070213

johanneswuerbach:
Thanks for the review @bryantbiggs, but I'm not sure I understand.

This implementation shouldn't be breaking any configuration (everything works as before), but avoids the warning if nothing is configured and allows users to gradually adopt the new properties.

bryantbiggs:
`resolve_conflicts_on_create` and `resolve_conflicts_on_update` are only available starting on v5.0 of the AWS provider - which upgrading to require this minimum is a breaking change for users

bryantbiggs:
apologies - wrong button 😅

johanneswuerbach:
Sorry, I totally missed that this module still supports AWS v4 :-/

bryantbiggs:
we will have v20 coming end of next month, early Dec and this will be resolved then

Issue #2763, opened 2023-10-03
CloudWatch Logs Log Group already exists (event with inline_policy Deny)


cdenneen:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/c7565e265e23cbc622041fc153193f490cfe948f/main.tf#L294-L318

_Originally posted by @bryantbiggs in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2486#issuecomment-1441751885_

So even with the supplied Deny the cluster-recreation complains because the logGroup was still written to.

Part of me almost suggests removing the depends_on from eks cluster to the log group and have the log group depend on the cluster. This way when the cluster is actually deleted THEN the log group gets deleted. This will get any last minute streams that might come through and hopefully prevent recreation.

@haarchri has this deny still worked for you? I know it was part of your original merge but it seems it keeps happening to me. My alternative would be to avoid creating replacement clusters with ones I just destroyed but this still wouldn't cleanup rogue log groups that should have properly been cleaned up.

cdenneen:
So the issue based on CloudTrail is that it's not the role recreating it but the cluster itself:

```
{
    "eventVersion": "1.09",
    "userIdentity": {
        "type": "AssumedRole",
        "principalId": "REDACTED",
        "arn": "arn:aws:sts::REDACTED:assumed-role/AWSServiceRoleForAmazonEKS/REDACTED",
        "accountId": "REDACTED",
        "accessKeyId": "REDACTED",
        "sessionContext": {
            "sessionIssuer": {
                "type": "Role",
                "principalId": "REDACTED",
                "arn": "arn:aws:iam::REDACTED:role/aws-service-role/eks.amazonaws.com/AWSServiceRoleForAmazonEKS",
                "accountId": "REDACTED",
                "userName": "AWSServiceRoleForAmazonEKS"
            },
            "attributes": {
                "creationDate": "2023-10-03T18:54:33Z",
                "mfaAuthenticated": "false"
            }
        },
        "invokedBy": "eks.amazonaws.com"
    },
    "eventTime": "2023-10-03T19:08:17Z",
    "eventSource": "logs.amazonaws.com",
    "eventName": "CreateLogGroup",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "eks.amazonaws.com",
    "userAgent": "eks.amazonaws.com",
    "requestParameters": {
        "logGroupName": "/aws/eks/test-use1/cluster"
    },
[...]

PR #2764, opened 2023-10-03
CloudWatch LogGroup depends_on

cdenneen:
Due to the logGroup being recreated upon destroy when streams are still sending events the logGroup resource is left orphaned in AWS. 

If not recreating the cluster this just leaves orphaned resource. If being recreated the apply of the newly created cluster will fail due to the log group already existing.

This is to change the dependency order to only delete the cloudwatch log group after the cluster has been successfully destroyed. #2763

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -85,7 +85,6 @@ resource "aws_eks_cluster" "this" {
     aws_iam_role_policy_attachment.this,
     aws_security_group_rule.cluster,
     aws_security_group_rule.node,
-    aws_cloudwatch_log_group.this,
     aws_iam_policy.cni_ipv6_policy,
   ]
 }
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -114,6 +113,10 @@ resource "aws_cloudwatch_log_group" "this" {
     var.tags,
     { Name = "/aws/eks/${var.cluster_name}/cluster" }
   )
+
+  depends_on = [
```
bryantbiggs, line 117: this will fail everytime - the cluster will be created, and it will create the log group and Terraform will fail with a "resource with the same name already exists"

cdenneen: Good point... didn't think of that... there has to be a way to fix this issue some other way since the deny doesn't appear to be working as expected. I know it's implemented different than the original PR, curious as to if that's making a difference in some way.

PR #2765, opened 2023-10-04
fix: Create dependency in the nodegroups on the aws_cloudwatch_log_group

### LINKED ISSUE #2763: CloudWatch Logs Log Group already exists (event with inline_policy Deny)

cdenneen:
Create dependency in the nodegroups on the aws_cloudwatch_log_group.
Once the nodegroup is deleted then the logGroup can.

Fixes #2763

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -371,6 +371,10 @@ module "eks_managed_node_group" {
   cluster_primary_security_group_id = try(each.value.attach_cluster_primary_security_group, var.eks_managed_node_group_defaults.attach_cluster_primary_security_group, false) ? aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id : null
 
   tags = merge(var.tags, try(each.value.tags, var.eks_managed_node_group_defaults.tags, {}))
+
+  depends_on = [
+    aws_cloudwatch_log_group.this,
+  ]
 }
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


cdenneen:
@bryantbiggs Since the node_groups can't be created without the cluster then putting a dependency on the log_group within their resource should allow for the node_group to be deleted first before the log_group. This would allow the workloads causing the recreation of the log_group to be terminated prior to log_group being destroyed.

bryantbiggs:
unfortunately, this isn't a scenario we are going to be able to solve with various hacks. there are other various ways that users can accomplish this if they are facing issues - such as disabling logs before de-provisioning the cluster

cdenneen:
@bryantbiggs still not sure why this isn’t an acceptable solution. The node group depends on the cluster and the cluster depends on the log_group so it shouldn’t affect anything, in fact it should address this issue. 
in an automation scenario where you are using TF to destroy a stack it doesn’t disable logging (unless you want to make it do that). Also for example if you are destroying a cluster why must it first remove the OIDC if configured. This causes a cluster update which gets destroyed anyway. Unless the eks module will disable cluster logging as part of the destroy I see no way in CI scenarios like this where people should have to disable logging, manually delete log groups, etc. 
I see no harm in this dependency being added.

bryantbiggs:
> Terraform is what builds and controls the lifecycle of these resources

That is the crux of the problem - Terraform handled the resources it was responsible for, but EKS will re-create the log group if it needs to send logs. This is outside of Terraform's control and therefore outside of this module's control

You will either need to disable logging, or de-provision the resources running on the cluster that are creating log events

PR #2767, opened 2023-10-05
fix: Added conditonal for creation of  kubernetes.io/cluster/cluster_name: Owned tag on Nodegroup SG

garrett-thinkbigger:
## Description
- Added boolean conditonal for creation of  kubernetes.io/cluster/cluster_name: Owned tag on Nodegroup SG
- Added default value for non-disruptive change

## Motivation and Context
- The presence of this tag is causing an issue for target group discovery with the Cluster Autoscaler
- Removal of the tag resolves the issue
- Have created non-disruptive change to address issue 

## Breaking Changes
Non-disruptive change to address issue, given variable default value of true

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -194,6 +194,7 @@ resource "aws_security_group" "node" {
     var.tags,
     {
       "Name"                                      = local.node_sg_name
+      "kubernetes.io/cluster/${var.cluster_name}" = var.nodegroup_sg_kubernetes_owner_tag_enabled == true ? "owned" : null
     },
     var.node_security_group_tags
   )
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -194,7 +194,6 @@ resource "aws_security_group" "node" {
     var.tags,
     {
       "Name"                                      = local.node_sg_name
-      "kubernetes.io/cluster/${var.cluster_name}" = "owned"
     },
     var.node_security_group_tags
   )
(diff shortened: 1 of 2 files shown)


bryantbiggs:
thank you for the PR - however, you can change this value today using the `node_security_group_tags` variable so we won't be needing this change

garrett-thinkbigger:
> thank you for the PR - however, you can change this value today using the `node_security_group_tags` variable so we won't be needing this change

Maybe I'm misunderstanding but the variable you've suggested above doesn't resolve the issue. What I am trying to achieve here is that the troublesome tag doesn't get created at all, if the variable is true.

The variable you've suggested would appear to create additional tags. Have I misunderstood? 

`  tags = merge(
    var.tags,
    {
      "Name"                                      = local.node_sg_name
      "kubernetes.io/cluster/${var.cluster_name}" = var.nodegroup_sg_kubernetes_owner_tag_enabled == true ? "owned" : null
    },
    var.node_security_group_tags
  )
`

bryantbiggs:
this change proposed here doesn't conditionally remove the tag - it conditionally toggles the tag value but the tag key is still there

if you want to change a tag's value, you can apply the same tag key later in the merge to override it, which is what `var.node_security_group_tags` does

bryantbiggs:
also, please see here https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md#i-received-an-error-expect-exactly-one-securitygroup-tagged-with-kubernetesioclustername-

Issue #2768, opened 2023-10-06
OIDC provider thumbprints inconsistent and sometimes incorrect


danielfrankcom:
## Description

When I use `terraform plan` the `thumbprint_list` of the included `oidc_provider` changes regularly, even when run back-to-back with no code changes. Additionally, the correct root thumbprint is not always included.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.
- [x] ✋ I cleared the `.terraform` cache and am able to reproduce the issue with a fresh init/apply. 

## Versions

- Module version:
  - eks: 19.16.0
  - vpc: 5.0.0

- Terraform & provider version(s):
```
$ terraform providers -version
Terraform v1.6.0
on windows_amd64
+ provider registry.terraform.io/hashicorp/aws v5.19.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.2
+ provider registry.terraform.io/hashicorp/kubernetes v2.23.0
+ provider registry.terraform.io/hashicorp/time v0.9.1
+ provider registry.terraform.io/hashicorp/tls v4.0.4
```

## Reproduction Code

```terraform
provider "aws" {
  region  = "ca-central-1"
  profile = "sandbox"
}

data "aws_availability_zones" "available" {}

locals {
  vpc_cidr = "10.0.0.0/16"
  azs      = slice(data.aws_availability_zones.available.names, 0, 3)
}

module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "5.0.0"

  name = "bug-report"

  azs             = local.azs
  private_subnets = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 4, k)]
}

module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "19.16.0"

  cluster_name    = "bug-report"
  cluster_version = "1.27"

  vpc_id     = module.vpc.vpc_id
  subnet_ids = module.vpc.private_subnets
}

output "fingerprint" {
  value = module.eks.cluster_tls_certificate_sha1_fingerprint
}
```

1. `terraform init`
2. `terraform apply`
3. `terraform plan` (repeat until Terraform shows changes, usually happens within 2-3 attempts)

## Expected behavior

```
No changes. Your infrastructure matches the configuration.
```

I also expect the root certificate for my region (9e99a48a9960b14926bb7f3b02e22da2b0ab7280) to be in the thumbprint list for the provider.

## Actual behavior

[...]

Issue #2732, opened 2023-09-04
Thumbprint list should only include CA fingerprint


gabegorelick:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.

https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2307 updated the OIDC provider `thumbprint_list` to include all  certificate fingerprints. This was valid at the time, since that's what the Terraform docs recommended. However, the docs have since been updated to recommend against this (https://github.com/hashicorp/terraform-provider-aws/issues/32847). 

Here's what's in the [latest docs](https://registry.terraform.io/providers/hashicorp/aws/5.15.0/docs/resources/eks_cluster):

> thumbprint_list = [data.tls_certificate.example.certificates[0].sha1_fingerprint]

Additionally, including all fingerprints can cause issues since AWS only allows a maximum of 5 to be specified (https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2632). So if you want to provide extra fingerprints, there may not be room.

I also think that official AWS tooling (Console, eksctl) only populates the CA fingerprint. Here's eksctl, for example:
https://github.com/eksctl-io/eksctl/blob/fb50ae558a9f9c3259fc42bda7cc3957a1d824d5/pkg/iam/oidc/api.go#L113

## Describe the solution you'd like.

Go back to the behavior before #2307 and only include the CA fingerprint in `thumbprint_list`.

## Describe alternatives you've considered.

Do nothing and live with the fact that this module behaves differently from what AWS and Terraform recommend.

bryantbiggs:
if you'd like to open a PR, I'll merge it

PR #2769, opened 2023-10-06
fix: Only include CA thumbprint in OIDC provider list

danielfrankcom:
## Description

This change modifies the OIDC provider `thumbprint_list`, so that it only contains the CA fingerprint, rather than all fingerprints.

## Motivation and Context

Addresses #2732, by reverting the change from #2307.

According to the description of #2732, #2307 introduced the inclusion of all certificate fingerprints in the `thumbprint_list`. Based on a later documentation change in hashicorp/terraform-provider-aws#32847, only the CA fingerprint should be included in the `thumbprint_list`.

From the [latest docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#enabling-iam-roles-for-service-accounts), the following should be used:

> thumbprint_list = [data.tls_certificate.example.certificates[0].sha1_fingerprint]

## Breaking Changes

This shouldn't break existing functionality, since the root CA encompasses the authority provided by any other certificates that were previously included.

## How Has This Been Tested?

I tested this change by modifying the reproduction code provided in #2768 to reference the modified `eks` module.

Before the change, 4 fingerprints were included for the OIDC provider. After the change, only the single root CA fingerprint was included.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -232,7 +232,7 @@ resource "aws_iam_openid_connect_provider" "oidc_provider" {
   count = local.create && var.enable_irsa && !local.create_outposts_local_cluster ? 1 : 0
 
   client_id_list  = distinct(compact(concat(["sts.${local.dns_suffix}"], var.openid_connect_audiences)))
-  thumbprint_list = concat(data.tls_certificate.this[0].certificates[*].sha1_fingerprint, var.custom_oidc_thumbprints)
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
thank you!

antonbabenko:
This PR is included in version 19.17.1 :tada:

PR #2770, opened 2023-10-08
feat: Expect exactly one securityGroup tagged with kubernetes.io/cluster/eks-1 for ENI

taimax13:
## Description
"kubernetes.io/cluster/${var.cluster_name}" = "owned" - removing hardcoded  tag from node sg
## Motivation and Context
ENI can have only one sg with this tag
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1897

## Breaking Changes
Can be added into var.tags if needed on the node level

## How Has This Been Tested?
Cluster re-created, changes didn't affect anything before this tag was removed manually, but due to multiple terraform applied it recreates all the time casein manual follow-up

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -194,7 +194,6 @@ resource "aws_security_group" "node" {
     var.tags,
     {
       "Name"                                      = local.node_sg_name
-      "kubernetes.io/cluster/${var.cluster_name}" = "owned"
     },
     var.node_security_group_tags
   )


bryantbiggs:
FYI https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md#i-received-an-error-expect-exactly-one-securitygroup-tagged-with-kubernetesioclustername-

taimax13:
@bryantbiggs FYI the documentation explained that if you want to use it separately you'll need to create sg for node ```  // The following variables are necessary if you decide to use the module outside of the parent EKS module context.
  // Without it, the security groups of the nodes are empty and thus won't join the cluster.
  cluster_primary_security_group_id = module.eks.cluster_primary_security_group_id
  vpc_security_group_ids            = [module.eks.node_security_group_id]``` from here https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/eks-managed-node-group/README.md

But you guys do not care about this usage, right?

bryantbiggs:
> But you guys do not care about this usage, right?

I don't follow - we do support that usage pattern

taimax13:
ok, cool - we will stick to fork then :)

PR #2771, opened 2023-10-09
fix: Karpenter node IAM role policies variable should be a map of strings, not list

### LINKED ISSUE #2750: [karpenter] Set SSMSessionManager on node IAM role by default
Karpenter docs recommend setting four role for AWS EKS clusters on the node IAM role (https://karpenter.sh/docs/getting-started/migrating-from-cas/). This module sets the first three but not `AmazonSSMManagedInstanceCore`, which is required for creating SSM sessions on Karpenter made nodes. 

This could be implemented like the `iam_role_attach_cni_policy` variable to attach the SSM policy (https://github.com/terraform-aws-modules/terraform-aws-eks/blob/c013f7bfbc9060f8463987cd12dbca9adfd4049d/modules/karpenter/main.tf#L336C9-L336C35)

In addition, there was some confusion in the issue https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2540 which resulted in an incorrect example entry https://github.com/terraform-aws-modules/terraform-aws-eks/blob/918aa7cc40cbc072836410747834de64d84f514d/examples/karpenter/main.tf#L163 The following config

```
  policies = {
    AmazonSSMManagedInstanceCore = "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
  }
```
Adds the `AmazonSSMManagedInstanceCore` policy to the ISRA role, when it should actually be added to the node role, like so

```
iam_role_additional_policies = ["arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"]
```

bryantbiggs:
## Description
- Karpenter node IAM role policies variable should be a map of strings, not list
- Update Karpenter example to show attaching policies on the node IAM role which is more common that attaching additional  policies on the IRSA role

## Motivation and Context
- Resolves #2750

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
           - '--args=--only=terraform_standard_module_structure'
           - '--args=--only=terraform_workspace_remote'
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.5.0
     hooks:
       - id: check-merge-conflict
       - id: end-of-file-fixer
--- a/examples/karpenter/main.tf
+++ b/examples/karpenter/main.tf
@@ -160,7 +160,13 @@ module "karpenter" {
   cluster_name           = module.eks.cluster_name
   irsa_oidc_provider_arn = module.eks.oidc_provider_arn
 
-  policies = {
+  # Used to attach additional IAM policies to the Karpenter controller IRSA role
+  # policies = {
+  #   "xxx" = "yyy"
+  # }
+
+  # Used to attach additional IAM policies to the Karpenter node IAM role
+  iam_role_additional_policies = {
     AmazonSSMManagedInstanceCore = "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
   }
 
@@ -176,7 +182,7 @@ resource "helm_release" "karpenter" {
   repository_username = data.aws_ecrpublic_authorization_token.token.user_name
   repository_password = data.aws_ecrpublic_authorization_token.token.password
   chart               = "karpenter"
-  version             = "v0.21.1"
+  version             = "v0.29.0"
 
   set {
     name  = "settings.aws.clusterName"
--- a/modules/karpenter/README.md
+++ b/modules/karpenter/README.md
@@ -42,6 +42,11 @@ module "karpenter" {
   irsa_oidc_provider_arn          = module.eks.oidc_provider_arn
   irsa_namespace_service_accounts = ["karpenter:karpenter"]
 
+  # Attach additional IAM policies to the Karpenter node IAM role
+  iam_role_additional_policies = {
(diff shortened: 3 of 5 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 19.17.2 :tada:

kkopachev:
Was not that a breaking change?

kkopachev:
I had a code which was passing this old style as a list and it's breaking now.
```
Error: Invalid value for input variable

  on ../../modules/k8s-addons/karpenter/main.tf line 8, in module "karpenter":
   8:   iam_role_additional_policies = [
   9:     "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
  10:   ]

The given value is not suitable for
module.karpenter.module.karpenter.var.iam_role_additional_policies declared
at
.terraform/modules/karpenter.karpenter/modules/karpenter/variables.tf:224,1-40:
map of string required.
```

Module code was
```
module "karpenter" {
  source  = "terraform-aws-modules/eks/aws//modules/karpenter"
  version = "~>19.13"

  cluster_name           = var.cluster_name
  irsa_oidc_provider_arn = var.oidc_provider_arn

  iam_role_additional_policies = [
    "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
  ]
}
```

Not a big deal, but unexpected

kamirendawkins:
Just wanted to hop on this this, its not a huge deal but watching a number of workspaces suddenly have issue leads me to feel that this is not fix but a potentially breaking change for those that are automatically pulling in patch levels.

mkunkel:
We also found this to be a breaking change. We have two workspaces that suddenly began to fail with this change.

bryantbiggs:
my apologies for the disruption, that was not intended. I am somewhat shocked the prior variable type didn't raise issues before because the usage is written to accept a map, not a list https://github.com/terraform-aws-modules/terraform-aws-eks/blob/51cc6bec880ac8dc361b60a4b05d5f2bcd98eb6a/modules/karpenter/main.tf#L355

I guess HCL must silently convert the list index to the `key` when using `for_each` over a list 🤔

kkopachev:
If you had already old format variable and just changing it to be a map, beware. Unless you use `moved` block to map old to new, there is a chance that terraform will try to create same policy (and ignore duplicate error) and then delete old policy, meaning that you'll end up without a policy at all. 
Either apply twice, or use `moved` block

absolutemikex:
I won't upgrade yet - but what is the recommended change to handle this gracefully? As others have mentioned I have this in my TF.

```   
 iam_role_additional_policies = [
  "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
    ]
```
Was the intention just to have you swap it to `{}` ?

bryantbiggs:
You would change that to:

```hcl
iam_role_additional_policies = {
  AmazonSSMManagedInstanceCore = "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"
} 
```

Note: the key portion doesn't really matter so you can name the key anything you like, it just has to be unique within the map

PR #2776, opened 2023-10-12
Add root_block_device input for module self_managed_node_group

vleedev:
## Description

Replace the root volume by specifying `root_block_device` as [describe](https://registry.terraform.io/providers/hashicorp/aws/5.20.1/docs/resources/instance#ebs-ephemeral-and-root-block-devices) for module `self_managed_node_group`
## Motivation and Context

- Currently, we can achieve this purpose by declaring in the `block_device_mappings` but we must specify `device_name` correctly. `device_name`  depends on the `root_block_device` attribute of the `AMI` and it is `/dev/xvda` as usual.
- With this feature, we won't worry about filling exactly `device_name` if we want to replace the root block device.
- Replacing the `AMI` supports `ebs` device type via the `root_device_type` attribute.
## Breaking Changes

None
## How Has This Been Tested?
- [ ok] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ok] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [ ok] I have executed `pre-commit run -a` on my pull request

### DIFF
--- a/examples/self_managed_node_group/main.tf
+++ b/examples/self_managed_node_group/main.tf
@@ -69,6 +69,10 @@ module "eks" {
       "k8s.io/cluster-autoscaler/enabled" : true,
       "k8s.io/cluster-autoscaler/${local.name}" : "owned",
     }
+    root_block_device = {
+      volume_type = "standard"
+      volume_size = 15
+    }
   }
 
   self_managed_node_groups = {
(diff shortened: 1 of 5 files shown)


bryantbiggs:
thank you - however, this is not a change that we will be accepting at this time

PR #2806, opened 2023-11-07
feat: Add prometheus-service port 9090 to recommended sec groups

panuhorsmalahti:
## Description
Adding port 9090, typically used by Prometheus service in Kubernetes clusters to the default recommend security groups.

When Prometheus is deployed with e.g. Lens Metrics provider this port is used.

## Motivation and Context
This will make Prometheus metrics work by default for different providers for various applications that display Prometheus metrics.

## Breaking Changes
None

## How Has This Been Tested?
- [X] I tested this rule with `node_security_group_additional_rules`

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -152,6 +152,15 @@ locals {
       type                          = "ingress"
       source_cluster_security_group = true
     }
+    # prometheus-service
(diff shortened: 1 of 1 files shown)


bryantbiggs:
thank you, but unfortunately we will not be adding this at this time.

PR #2800, opened 2023-10-31
feat: Add Karpenter v1beta1 compatibility

mbarrien:
## Description
Add IAM support for Karpenter v1beta1 resources.

## Motivation and Context
The recent Karpenter v0.32 release added support for v1beta1 resources, as described in https://github.com/aws/karpenter/blob/main/designs/v1beta1-full-changelist.md. As part of this change, new permissions are needed such as the ability to create instance profiles.

The new controller IAM policy is described in https://github.com/aws/karpenter/blob/main/website/content/en/preview/upgrading/v1beta1-controller-policy.json

This PR adds a minimal set of changes for v0.32 compatibility, specifically around instance profiles. This gives the IRSA role the ability to create any instance profile in the account, without attempting to scope it down based on tags as https://github.com/aws/karpenter/pull/3948 would have done.

## Breaking Changes
No breaking changes, only adds permissions.

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #2812, opened 2023-11-09
fix: Remove additional conditional on Karpenter instance profile creation to support upgrading

bryantbiggs:
## Description
- Remove additional conditional on Karpenter instance profile creation to support upgrading

## Motivation and Context
- When upgrading to 0.32, Karpenter will now create the instance profile used by the nodes that it creates. The permissions to support this were added in #2800. However, removing the instance profile too soon can cause disruptions since existing nodes will still be using this profile until Karpenter replaces all of those existing nodes with new nodes that are using an instance profile that *IT has created. Since there is already a conditional on the profile creation (`create_instance_profile`), we are removing the additional conditional added in #2800 to allow users to opt into the new permissions required for 0.32, and then later opt out of the module created instance profile once they feel its safe to do so (i.e. - after all nodes have been replaced with nodes using the Karpenter created instance profile)

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/modules/karpenter/main.tf
+++ b/modules/karpenter/main.tf
@@ -386,7 +386,7 @@ locals {
 }
 
 resource "aws_iam_instance_profile" "this" {
-  count = var.create && var.create_instance_profile && !var.enable_karpenter_instance_profile_creation ? 1 : 0
+  count = var.create && var.create_instance_profile ? 1 : 0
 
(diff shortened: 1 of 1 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 19.19.1 :tada:

PR #2818, opened 2023-11-15
fix: Optionalize cluster endpoint and cert data in templates

### LINKED ISSUE #2787: invalid userdata due to optional attributes
## Description

TLDR: The variables `cluster_endpoint` and `cluster_auth_base64` are optional, this allows for a semi broken userdata.

We are configuring the EKS cluster and nodegroups separate, so we manually call eks-managed-node-group. 

We have now noticed, that we're not passing the endpoint and certificate values, as those are optional. However the `_user_data`
 / template don't cope with that properly.

This results in a userdata with empty variables:

    B64_CLUSTER_CA=
    API_SERVER_URL=
    /etc/eks/bootstrap.sh go-main-prod  --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $API_SERVER_URL
    echo "go get 'em little kubelet!"

During boot we thus get:

    cloud-init[3329]: + /etc/eks/bootstrap.sh go-main-prod --b64-cluster-ca --apiserver-endpoint
    cloud-init[3329]: 2023-10-18T14:39:58+0000 [eks-bootstrap] INFO: starting...
    cloud-init[3329]: 2023-10-18T14:39:58+0000 [eks-bootstrap] INFO: --b64-cluster-ca='--apiserver-endpoint'
    ...
    cloud-init[3329]: 2023-10-18T14:39:58+0000 [eks-bootstrap] INFO: --cluster-ca or --api-server-endpoint is not defined, describing cluster...

So only because of the order, the bootstrap is not failing completely.

## Versions

- Module version [Required]: v.19.17.2

- Terraform version: not relevant
- Provider version(s): not relevant

calling eks_worker_groups without `cluster_endpoint` and `cluster_auth_base64`

    module "eks_worker_groups" {
      source  = "terraform-aws-modules/eks/aws//modules/eks-managed-node-group"
      version = "v19.15.4"
      ....

not relevant

## Expected behavior

`cluster_endpoint` and `cluster_auth_base64` should be mandatory
OR
`_user_data` should filter

## Actual behavior

semi-broken userdata gets created

lobeck:
## Description
The cluster endpoint and certificate data should not be rendered in the userdata template if they are not passed to the module.

## Motivation and Context
This is the case, when the cluster is configured for calico or another CNI plugin. The values are optional in terraform and should only be rendered when passed, else it creates an invalid bash command with possible unintended side effects.

The bootstrap script still works without the values due to an internal fallback to a describe cluster call.

fixes #2787

## Breaking Changes
no

## How Has This Been Tested?

### DIFF
--- a/templates/linux_user_data.tpl
+++ b/templates/linux_user_data.tpl
@@ -7,8 +7,8 @@ ${pre_bootstrap_user_data ~}
 export SERVICE_IPV4_CIDR=${cluster_service_ipv4_cidr}
 %{ endif ~}
 %{ if enable_bootstrap_user_data ~}
-B64_CLUSTER_CA=${cluster_auth_base64}
-API_SERVER_URL=${cluster_endpoint}
-/etc/eks/bootstrap.sh ${cluster_name} ${bootstrap_extra_args} --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $API_SERVER_URL
+%{ if cluster_auth_base64 != "" }B64_CLUSTER_CA=${cluster_auth_base64}%{ endif }
+%{ if cluster_endpoint != "" }API_SERVER_URL=${cluster_endpoint}%{ endif }
+/etc/eks/bootstrap.sh ${cluster_name} ${bootstrap_extra_args} %{ if cluster_auth_base64 != "" }--b64-cluster-ca $B64_CLUSTER_CA%{ endif } %{ if cluster_endpoint != "" }--apiserver-endpoint $API_SERVER_URL%{ endif }
 ${post_bootstrap_user_data ~}
 %{ endif ~}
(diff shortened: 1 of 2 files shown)


bryantbiggs:
in what scenario would you want to avoid passing these values from Terraform when they are already known in Terraform?

lobeck:
There is no convincing reason not to set it. But there's a high chance, that it gets missed, as the parameters are optional.

One such case is, that a custom CNI plugin gets used where objects in the cluster need to be reconfigured before the nodes are started.

bryantbiggs:
I don't think this change makes sense in this project - it will also be quite disruptive to users since any changes to user data will show up in the next plan/apply for them and affect nearly all nodes created by this module

> One such case is, that a custom CNI plugin gets used where objects in the cluster need to be reconfigured before the nodes are started.

I don't see how this relates to these values in the user data

PR #2848, opened 2023-12-08
fix: Update description for karpenter "create" var

arunderwood:
## Description

Updating the documentation for the `create` variable to be more accurate.

## Motivation and Context

The current description string is confusing since the variable doesn't seem to control any managed node groups.  It's possible that this variable should simply be removed since it's use is unclear.

## Breaking Changes
This is a documentation only change

## How Has This Been Tested?
This is a documentation only change


antonbabenko:
This issue has been resolved in version 20.0.0 :tada:

PR #2856, opened 2023-12-15
fix: Do not fetch versions of the add-on if an explicit version is specified

### LINKED ISSUE #2855: The default version of the add-on is retrieved even if an explicit version is specified
## Description

The EKS module uses `aws_eks_addon_version` data source to retrieve default/latest version of all add-ons, even the ones where the versions are explicitly specified through the `addon_version` map entry.

This has become an issue with the EFS CSI add-on recently, because the default version for it does not seem to exist:

```
$ aws eks describe-addon-versions --kubernetes-version 1.27 --addon-name aws-efs-csi-driver --query addons[0].addonVersions[*].compatibilities[0].defaultVersion
[
    false,
    false,
    false,
    false,
    false
]
```

As a result, even if I pass a version to the module explicitly, I have an following error on `terraform plan`.

## Versions

- Module version `19.20.0`:

- Terraform version:
```
Terraform v1.5.7
on linux_amd64
```
- Provider version(s):
```
+ provider registry.terraform.io/hashicorp/aws v5.29.0
```

```hcl
module "eks" {
  version = "19.20.0"

  create = true

  cluster_name    = "test"
  cluster_version = "1.27"

  cluster_addons = {
    aws-efs-csi-driver = {
      addon_version            = "v1.7.1-eksbuild.1"
      most_recent              = false
    }
  }
}

```

```
terraform plan
```

## Expected behavior

Data source is not instantiated for any add-ons with an explicit version.
No error on the plan.

## Actual behavior

Error on the plan.

```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: reading EKS Add-On version info (aws-efs-csi-driver, 1.27): empty result
│ 
│   with module.eks.data.aws_eks_addon_version.this["aws-efs-csi-driver"],
│   on .terraform/modules/eks/main.tf line 439, in data "aws_eks_addon_version" "this":
│  439: data "aws_eks_addon_version" "this" {
│ 
╵
```

## Additional context

N/A

sergei-ivanov:
## Description

Fixes #2855

## Motivation and Context

Currently the usage of the module is broken because of the broken add-on metadata on the AWS side.
The fix allows circumventing the metadata issue by using explicit add-on version.

## Breaking Changes

No

## How Has This Been Tested?

The change has been tested extensively in our own environment.
All combinations of explicit, latest, default versions of the add-ons were tested.
I haven't updated the examples because they all rely on the latest versions of the add-ons. Having specific add-on versions in the code risks creating a maintenance burden.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -394,7 +394,7 @@ resource "aws_eks_addon" "this" {
   cluster_name = aws_eks_cluster.this[0].name
   addon_name   = try(each.value.name, each.key)
 
-  addon_version            = coalesce(try(each.value.addon_version, null), data.aws_eks_addon_version.this[each.key].version)
+  addon_version            = coalesce(try(each.value.addon_version, null), try(data.aws_eks_addon_version.this[each.key].version, null))
   configuration_values     = try(each.value.configuration_values, null)
   preserve                 = try(each.value.preserve, null)
   resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")
@@ -422,7 +422,7 @@ resource "aws_eks_addon" "before_compute" {
   cluster_name = aws_eks_cluster.this[0].name
   addon_name   = try(each.value.name, each.key)
 
-  addon_version            = coalesce(try(each.value.addon_version, null), data.aws_eks_addon_version.this[each.key].version)
+  addon_version            = coalesce(try(each.value.addon_version, null), try(data.aws_eks_addon_version.this[each.key].version, null))
   configuration_values     = try(each.value.configuration_values, null)
   preserve                 = try(each.value.preserve, null)
   resolve_conflicts        = try(each.value.resolve_conflicts, "OVERWRITE")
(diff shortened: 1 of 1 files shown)


bryantbiggs:
this was a hiccup in the AWS API, we  no longer need any changes here - but thank you!

sergei-ivanov:
@bryantbiggs would it still make sense to apply this change in order to: 1) reduce the number of data source refreshes if the versions are already given, and 2) guard against any similar future problems with AWS API?

bryantbiggs:
this is an extreme edge case, and folks should be using the data source and avoid static version to ensure addons are kept updated in lockstep with the cluster

sergei-ivanov:
We pin down the versions because we prefer upgrading the higher environments in a controlled way, especially production. Relying on the data source may produce spontaneous drifts, and nobody likes surprises during deployment to production environment.

PR #2858, opened 2023-12-19
feat: Replace resolve_conflicts with resolve_conflicts_on_create/…

### LINKED ISSUE #2525: error configmap  "aws-auth" does not exist and  creating IAM Policy ekstest-additional: ConcurrentModification: Found Tagris has different internal
## Description
se esta implementando un eks con nodos administrados pero s iempre falla
por favor revisar o dar una solucion 

adjunto codigo y el error que sale 

```
provider "kubernetes" {
  host                   = module.eks.cluster_endpoint
  cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
  token                  = data.aws_eks_cluster_auth.default.token

  exec {
    api_version = "client.authentication.k8s.io/v1beta1"
    command     = "aws"
    # This requires the awscli to be installed locally where Terraform is executed
    args = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
  }
}

data "aws_availability_zones" "available" {}

locals {
  name            = "ekstest"
  cluster_version = "1.25"
  region          = "eu-west-2"

  
  tags = {
    Cluster   = local.name
    Project = "culqi"
    PCI= "false"
  }
}

resource "aws_eks_addon" "coredns" {
  cluster_name      = module.eks.cluster_id
  addon_name        = "coredns"
  resolve_conflicts = "OVERWRITE"
  depends_on = [
    module.eks
  ]
}


# EKS Module

module "eks" {
  source = "../../module/eks"

  cluster_name                   = local.name
  cluster_version                = local.cluster_version
  cluster_enabled_log_types= [
    "api",
    "audit",
    "authenticator",
    "controllerManager",
    "scheduler"
  ]
  cluster_endpoint_public_access = true #double ccheck

  # IPV6
  cluster_ip_family = "ipv6"
  create_cni_ipv6_iam_policy = true


  vpc_id                   = data.aws_vpc.vpc_name-application.id
  subnet_ids               = ["${data.aws_subnet.subnets01.id}", "${data.aws_subnet.subnets02.id}", "${data.aws_subnet.subnets03.id}"]
  #control_plane_subnet_ids = module.vpc.intra_subnets

 


  eks_managed_node_group_defaults = {
    ami_type       = "AL2_x86_64"
    instance_types = ["m5.xlarge"]

    attach_cluster_primary_security_group = true
   
    iam_role_attach_cni_policy = true
  }
  cluster_addons = {
     
      kube-proxy = {
        most_recent = true
      }
      vpc-cni = {
        most_recent              = true
        before_compute           = true
        service_account_role_arn = module.vpc_cni_irsa.iam_role_arn
          env = {
          }
        })
      }
    }



  manage_aws_auth_configmap = true
  eks_managed_node_groups = {
 
    complete = {
      name            = "default_node"
      use_name_prefix = true
      

      subnet_ids = ["${data.aws_subnet.subnets01.id}", "${data.aws_subnet.subnets02.id}", "${data.aws_subnet.subnets03.id}"]
      use_custom_launch_template = false

      min_size     = 2
      max_size     = 10
      desired_size = 4

      capacity_type        = "SPOT"
      force_update_version = true
      instance_types       = ["m5.xlarge"]
      disk_size = 180
      labels = local.tags

      ami_id                     = data.aws_ami.eks_default.image_id
      enable_bootstrap_user_data = true

      pre_bootstrap_user_data = <<-EOT
        export FOO=bar
      EOT

      post_bootstrap_user_data = <<-EOT
        echo "you are free little kubelet!"
      EOT

      

      taints = [
        {
          key    = "dedicated"
          value  = "gpuGroup"
          effect = "NO_SCHEDULE"
        }
      ]

      update_config = {
        max_unavailable_percentage = 33 # or set `max_unavailable`
      }

      description = "EKS managed node group  launch template"

      ebs_optimized           = true
      disable_api_termination = false
      enable_monitoring       = true

      block_device_mappings = {
        xvda = {
          device_name = "/dev/xvda"
          ebs = {
            volume_size           = 180
            volume_type           = "gp3"
            iops                  = 3000
            throughput            = 150
            encrypted             = true
            kms_key_id            = module.ebs_kms_key.key_arn
            delete_on_termination = true
          }
        }
      }

      # Remote access cannot be specified with a launch template
      remote_access = {
        ec2_ssh_key               = module.key_pair.key_pair_name
        source_security_group_ids = [aws_security_group.remote_access.id]
      }

      metadata_options = {
        http_endpoint               = "enabled"
        http_tokens                 = "required"
        http_put_response_hop_limit = 2
        instance_metadata_tags      = "disabled"
      }

      create_iam_role          = true
      iam_role_name            = "eks-managed-node-group-complete-example"
      iam_role_use_name_prefix = false
      iam_role_description     = "EKS managed node group complete example role"
      iam_role_tags = {
        Purpose = "Protector of the kubelet"
      }
      iam_role_additional_policies = {
        AmazonEC2ContainerRegistryReadOnly = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
        additional                         = aws_iam_policy.node_additional.arn
      }

      tags = {
        ExtraTag = "EKS managed node group"
      }
    }
  }
  
# Extend cluster security group rules
  cluster_security_group_additional_rules = {
    ingress = {
      description                = "EKS Cluster allows 443 port to get API call"
      type                       = "ingress"
      from_port                  = 443
      to_port                    = 443
      protocol                   = "TCP"
      cidr_blocks                = ["0.0.0.0/0"]
      source_node_security_group = false
    }
  }
  tags = local.tags
}

# Supporting Resources

module "vpc_cni_irsa" {
  source  = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts-eks"
  version = "~> 5.0"

  role_name_prefix      = "VPC-CNI-IRSA"
  attach_vpc_cni_policy = true
  vpc_cni_enable_ipv6   = true

  oidc_providers = {
    main = {
      provider_arn               = module.eks.oidc_provider_arn
      namespace_service_accounts = ["kube-system:aws-node"]
    }
  }

  tags = local.tags
}

module "ebs_kms_key" {
  source  = "terraform-aws-modules/kms/aws"
  version = "~> 1.5"

  description = "Customer managed key to encrypt EKS managed node group volumes"

  # Policy
  key_administrators = [
    data.aws_caller_identity.current.arn
  ]

  key_service_roles_for_autoscaling = [
    # required for the ASG to manage encrypted volumes for nodes
    "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/aws-service-role/autoscaling.amazonaws.com/AWSServiceRoleForAutoScaling",
    # required for the cluster / persistentvolume-controller to create encrypted PVCs
    module.eks.cluster_iam_role_arn,
  ]

  # Aliases
  aliases = ["eks/${local.name}/ebs"]

  tags = local.tags
}

module "key_pair" {
  source  = "terraform-aws-modules/key-pair/aws"
  version = "~> 2.0"

  key_name_prefix    = local.name
  create_private_key = true

  tags = local.tags
}

resource "aws_security_group" "remote_access" {
  name_prefix = "${local.name}-remote-access"
  description = "Allow remote SSH access"
  vpc_id      = data.aws_vpc.vpc_name-application.id

  ingress {
    description = "SSH access"
    from_port   = 22
    to_port     = 22
    protocol    = "tcp"
    cidr_blocks = ["10.0.0.0/8"]
  }

   egress {
    from_port        = 0
    to_port          = 0
    protocol         = "-1"
    cidr_blocks      = ["0.0.0.0/0"]
    ipv6_cidr_blocks = ["::/0"]
  }

  tags = merge(local.tags, { Name = "${local.name}-remote" })
}



resource "aws_iam_policy" "node_additional" {
  name        = "${local.name}-additional"
  description = "node additional policy"

  policy = jsonencode({
    Version = "2012-10-17"
    Statement = [
      {
        Action = [
          "ec2:Describe*",
        ]
        Effect   = "Allow"
        Resource = "*"
      },
    ]
  })

  tags = local.tags
}

data "aws_ami" "eks_default" {
  most_recent = true
  owners      = ["amazon"]

  filter {
    name   = "name"
    values = ["amazon-eks-node-${local.cluster_version}-v*"]
  }
}



data "aws_vpc" "vpc_name-application" {
  filter {
    name   = "tag:Name"
    values = ["${var.vpc_name-application}"]
  }
}


data "aws_subnet" "subnets01" {
   filter {
     name   = "tag:Name"
     values = ["vpc-application-private-us-west-2b"]
   }
}

data "aws_subnet" "subnets02" {
   filter {
     name   = "tag:Name"
     values = ["vpc-application-private-us-west-2a"]
   }
}

data "aws_subnet" "subnets03" {
   filter {
     name   = "tag:Name"
     values = ["vpc-application-private-us-west-2c"]
   }
}

data "aws_eks_cluster_auth" "default" {
  name = local.name
}

resource "aws_iam_policy" "additional" {
  name        = "${local.name}-additional"
  description = "Example usage of node additional policy"

  policy = jsonencode({
    Version = "2012-10-17"
    Statement = [
      {
        Action = [
          "ec2:Describe*",
        ]
        Effect   = "Allow"
        Resource = "*"
      },
    ]
  })

  tags = local.tags
}
```

- [ ] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 
terraform-aws-eks 19.10.1

- Terraform version: 
Terraform v1.3.9

2023-03-17 21_40_25-main tf - eks - Visual Studio Code

### LINKED ISSUE #2541: I/O Timeout when trying to create the aws-auth configmap
## Description

In the latest versions of this module, the `aws-auth` ConfigMap is manipulated, but there is no mechanism that waits for the cluster to be ready to accept connections. This leads to I/O Timeout errors as seen in the screenshot below.

What's frustrating is, this doesn't happen all the time. It DOES happen frequently (but not always) in our use case of:

* Kubernetes public endpoint is disabled
* Self-managed nodegroups only

This is clear evidence that there is a race condition happening. Sometimes Terraform will try to create the aws-auth configmap and the EKS cluster will have had enough time to become ready, and the `terraform apply` succeeds. Other times Terraform will try to create the configmap and it will fail because the cluster isn't ready to accept connections yet.

In previous versions of this module we've seen:

* A `null_resource` with a `local_exec` that uses curl
* Use of the HTTP provider to wait for the cluster to be ready
* Management of the aws-auth configmap removed completely
* Management of the aws-auth configmap brought back, but without any "wait_for_cluster_ready" mechanism

## Versions

- Module version [Required]: v19.11.0

- Terraform version: v1.3.9
- Provider version(s):
```
+ provider registry.terraform.io/hashicorp/aws v4.59.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.18.1
+ provider registry.terraform.io/hashicorp/time v0.9.1
+ provider registry.terraform.io/hashicorp/tls v4.0.4
```

I haven't had time to create a minimal example. I can go do so if necessary. I'm hoping the issue is self-evident here as this is something that has been a common issue for as long as Terraform has been used to deploy EKS clusters.

Deploy this module with the public k8s endpoint disabled and with self-managed nodegroups only. Set both `create_aws_auth_configmap` and `manage_aws_auth_configmap` to true.

## Expected behavior

The aws-auth configmap is always created successfully

## Actual behavior

Sometimes a failure occurs due to the cluster not being ready to accept connections yet

image

## Additional context

We are previous users of https://github.com/aws-ia/terraform-aws-eks-blueprints but given the new [direction](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/docs/v5-direction/DIRECTION_v5.md) that repo is going we have switched over to using this module directly, however due to this issue things are broken for our use case due to this race condition.

### LINKED ISSUE #2635: Deprecated warning with new hashicorp/aws v5.0.1 provider

### LINKED ISSUE #2733: Karpenter IRSA IAM policy to be like blueprints one
We are porting Karpenter module from the old blueprints v4 to this module submodule, but the IRSA IAM policy is remarkably different to make the same infra fail to run when moving between modules

For karpenter IRSA IAM policy in this module be more like blueprints one

- we temporarily added the missing permissions to a new extra policy we pass to the module under `irsa_additional`

## Additional context
https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/80bd3b9/modules/kubernetes-addons/karpenter/data.tf
vs
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/771465be/modules/karpenter/main.tf#L70-L163

### LINKED ISSUE #2809: Karpenter roles to follow new policy guide from new release v0.32.x
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - No 🛑: please wait to file a request until the functionality is avaialble in the AWS provider
  - Yes ✅: please list the AWS provider version which introduced this functionality

The new Karpenter version moved the product to beta and changed how the IAM policies for the controller and nodes should be configured. 

Follow Karpenter provided policy: https://github.com/aws/karpenter/blob/main/website/content/en/docs/upgrading/v1beta1-controller-policy.json

You could provide a new additional policy, but I believe the default behavior of the module should be aligned with the latest Karpenter version

## Additional context
I would like to work on this feature if there is no active work yet

### LINKED ISSUE #2816: AccessDeniedException kms:DescribeKey
## Description

Following up on https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2678 as this just effectively locked us out of all our clusters.

Having [kms_key_enable_default_policy](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/0a17f655fb7da00640627ed9255f1d96e42fcfd7/variables.tf#L161C11-L161C40) set to `false` by default can cause permanent lockouts if the [kms_key_owners](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/0a17f655fb7da00640627ed9255f1d96e42fcfd7/variables.tf#L167C11-L167C25) or [kms_key_administrators](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/0a17f655fb7da00640627ed9255f1d96e42fcfd7/variables.tf#L173C11-L173C33) variables are not set to something static (like the account root).

We are using policies created from AWS Identity Center at the management account level. A change in these policies can cause all inherited roles to get recreated, leaving users with the same policies but a different assume-role ARN.

As KMS policies don't consider global `*` IAM policies as valid, we have now lost the ability to decrypt our clusters and even `terraform destroy` fails (without some state-tinkering).

I would strongly suggest changing the default back to `true`, add the root account by default to one of the KMS admin variables or add a _(very)_ big disclaimer on the README telling users to consider the potentional downside of managing the KMS policies manually.

### LINKED ISSUE #2850: Add EKS Pod Identity support to the module
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: please list the AWS provider version which introduced this functionality
  
 AWS Announced new EKS Pod Identity feature, and AWS SDK Go v2 has been released including this feature.

This New Resource was added to the AWS provider: [aws_eks_pod_identity_association](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_pod_identity_association)

This functionality has been released in [v5.29.0 of the Terraform AWS Provider](https://github.com/hashicorp/terraform-provider-aws/blob/v5.29.0/CHANGELOG.md). 

```
resource "aws_eks_pod_identity_association" "foo" {
  cluster_name = "my_cluster"
  namespace = "my_namespace"
  service_account_name = "foo-sa"
  iam_role_arn = "aws:......:foo"
}
```

The solution should allow users to specify the necessary parameters for the `aws_eks_pod_identity_association` resource within the Terraform module. Specifically, users should be able to define the `cluster_name`, `namespace`, `service_account_name`, and `iam_role_arn` as shown in the example below:

```hcl
module "eks_pod_identity" {
  source = "path/to/eks_pod_identity_module"

  cluster_name         = "my_cluster"
  namespace            = "my_namespace"
  service_account_name = "foo-sa"
  iam_role_arn         = "aws:......:foo"
}
```

I am currently using IRSA, which is ok but the new functionality is better

bryantbiggs:
## Motivation and Context

- Resolves #2525
- Resolves #2541
- Resolves #2635
- Resolves #2733
- Resolves #2809
- Resolves #2816
- Resolves #2848
- Resolves #2850

## How Has This Been Tested?


main.tf
```diff
@@ -239,7 +241,7 @@ resource "aws_iam_openid_connect_provider" "oidc_provider" {
   # Not available on outposts
   count = local.create_oidc_provider ? 1 : 0
 
-  client_id_list  = distinct(compact(concat(["sts.${local.dns_suffix}"], var.openid_connect_audiences)))
+  client_id_list  = distinct(compact(concat(["sts.amazonaws.com"], var.openid_connect_audiences)))
```
joshsizer, line 244: Curious why this is being changed to a hard-coded value. Does this affect other AWS partitions, like govcloud?

bryantbiggs: No need for it to be dynamic, and no it does not affect other partitions - https://github.com/boto/botocore/blob/7ee6eeed83ea1bd4c527197fffec8bb08b00c5d7/botocore/data/sts/2011-06-15/service-2.json#L6

kamirendawkins: This may be outside the scope of this PR but the only reason this does not have to be dynamic is due to the use of the STS Global Endpoint, which is US-East-1 in disguise. Would it be better to either switch to the Regional Endpoints(Not all are Enabled by Default so this could be a breaking change), or at least allow for users to decide via a feature toggle?

main.tf
```diff
@@ -16,6 +16,8 @@ locals {
 
   create_outposts_local_cluster    = length(var.outpost_config) > 0
   enable_cluster_encryption_config = length(var.cluster_encryption_config) > 0 && !local.create_outposts_local_cluster
+
+  tags = merge(var.tags, { terraform-aws-modules = "eks" })
```
kamirendawkins, line 20: Should the module be forcing tags like this? From a cursory search only the `alb` and `emr` modules are currently doing this and I am curious if this is going to be a standard practice moving forward?

modules/karpenter/variables.tf
```diff
@@ -17,92 +17,101 @@ variable "cluster_name" {
 }
 
 ################################################################################
-# IAM Role for Service Account (IRSA)
+# Pod Identity IAM Role
```
kamirendawkins, line 20: It does not appear this is the case but the naming comes across as confusing. IRSA and Pod Identity are two distinct forms of authentication, is the intent here that module consumers should switch to Pod Identity or is IRSA still the direction?

main.tf
bryantbiggs: It will be standard going forward to drive support for the modules

modules/karpenter/variables.tf
bryantbiggs: Correct, Pod Identity is the direction

main.tf
kamirendawkins: I am not sure how adding this tag will drive support but if we are going to have a tag forced onto the resources it would be great to somehow include the module version so that consumers may also be able to benefit from tracking module and module version usage.

I also have concern around consumers with strict tagging policies and having no way to adjust/remove this to align with organizational requirements.

modules/karpenter/variables.tf
kamirendawkins: If that is the case there are other references to IRSA in the Karpenter Module that may need to be addressed? I am assuming this is just the first step since the Assume Role Policy still appears to be IRSA Compatible. Is there a timeline on when consumers would have to additionally support EKS Pod Identity to continue using this component of the module?

docs/UPGRADE-20.0.md
```diff
@@ -0,0 +1,91 @@
...
+
+Please consult the `examples` directory for reference example configurations. If you find a bug, please open an issue with supporting configuration to reproduce.
+
+## List of backwards incompatible changes
+
+- Minium supported AWS provider version increased to `v5.0`
+- Minimum supported Terraform version increased to `v1.1` to support Terraform state `moved` blocks
+- The `resolve_conflicts` argument within the `cluster_addons` configuration has been replaced with `resolve_conflicts_on_create` and `resolve_conflicts_on_delete` now that `resolve_conflicts` is deprecated
+- The `cluster_addons` `preserve` argument default/fallback value is now set to `true`. This has shown to be useful for users deprovisioning clusters while avoiding the situation where the CNI is deleted too early and causes resources to be left orphaned which results in conflicts.
+- The Karpenter sub-module's use of the `irsa` naming convention has been replaced with `pod-identity` along with an update to the Karpenter controller IAM policy to align with the `v1beta1`/`v0.32` changes
```
FernandoMiguel, line 11: for those of us that want to keep using irsa, and not have pod-identity in the cluster, will there be an option to do so?

joshsizer: I have a similar question. This module should continue to support IRSA

bryantbiggs: I don't follow, can you elaborate? What is the use case or problem that you are solving for?

FernandoMiguel: we would like to continue using regular tradicional IRSA instead of moving to https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html using
```
"Principal": {
                "Service": "pods.eks.amazonaws.com"
            }
```

bryantbiggs: For what though?

FernandoMiguel: for karpenter pods, for the way karpenter auths against AWS IAM
if the language is changing here to refer to `pod-identity`, it does indicate it will use a different method of talking to IAM

bryantbiggs: Ah ok, yes - the Karpenter sub-module here will be using pod identify going forward starting in v20. The IAM policy is still TBD, but the intent is to follow pod identity

FernandoMiguel: can both not be supported?
that's the ask here

bryantbiggs: No. And most likely the role will be removed from this sub-module in favor of consolidating on the in-progress https://github.com/clowdhaus/terraform-aws-eks-pod-identity

Why: pod identity allows for roles to be shared while still permitting scoped permissions based on session tags. The instance termination resources (queue, event rules, etc.) are still cluster specific and follow a 1:1 mapping.

kamirendawkins: This relates to my question that was marked resolved: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2858#discussion_r1434694155

Pod-Identity Requires another DaemonSet which, amongst other things I am sure I am missing, is more Node overhead and security reviews(for some) to even be able to use. Locking a Karpenter Update(v0.32+) behind an authentication change with no overlap seems like it would do the opposite of drive the adoption of this module.

Since Karpenter supports both methods it would stand to reason that the preferred module would also support both?

bryantbiggs: IRSA does not support session tags so you cannot have both for this particular policy. Pod identity is the successor to IRSA and that is the direction we are moving - you can continue to use IRSA on the v19.x module versions

FernandoMiguel: we can keep using v19, but we wont have updated iam policies for karpenter v.32+, correct?
or were all the require changes done in that version?

bryantbiggs: Again, we need to be specific - what do you mean by "updated policies"? IRSA does not support session tags, so you cannot have a policy crafted for scoped re-use with session tags that pod identity supports and simply swap out the trust policy for IRSA

kamirendawkins: Few things, not everyone wants to share Policies between Clusters for a myriad of reasons, both directions are valid. The issue here is that currently, v19.x, the Karpernter Policy is incompatible with the new CRDs provided in preparation for 1.x of Karpenter. This is a problem for many trying to stay on top of updates etc. The Issues requesting v0.32+ support indicate that this is to be solved in v20.x however this comes at the cost of also having to support Pod Identity which for some will be a blocker.

So if I am understanding this correctly it sounds like for anyone that does not want to use Pod Identity for any reason(for Karpenter in this case) then they have until v20.x to fork and migrate off of this module and unblock themselves with reg. to a Karpenter Update.

They could also stop using the Karpenter Sub-Module but I feel that with a sudden change like this teams would question the sustainability of the core module as well.

joshsizer: Is this entire module removing support for IRSA? Unfortunately pod Identity is not supported in AWS GovCloud regions: https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html#pod-id-considerations

It’s a shame to me to lose out to other module updates due to this hard switch to pod identities

bryantbiggs: > share Policies between Clusters for a myriad of reasons

Reasons, such as? This is a common pattern, no? Service linked roles, the role used by Terraform to execute applies, etc. I'm not sure how this applies to the direction of Pod Identity (just because you can use a common role doesn't dictate that you have to - we tend to want to move in the direction of more functionality and flexibility, not less - no?)

> The issue here is that currently, v19.x, the Karpernter Policy is incompatible with the new CRDs provided in preparation for 1.x of Karpenter.

Is this true? We have a number of folks using this sub-module with the latest Karpenter version today

> This is a problem for many trying to stay on top of updates etc. 

How does this relate to this module? This module is trying to move forward to support the latest features and functionality provided by Amazon EKS. Users are free to update/upgrade as they see fit, but this project is not held back by the pace at which users update/upgrade.

I feel like we're just fighting change here for the sake of fighting change. If you do not wish to use pod identity, you do not have to - nobody is forcing you. However, we are not going to hold back the development of this module because a few folks do not want to change. We can't satisfy all users - we still have users using very old versions of this module and thats perfectly fine. But this module will always strive to move forward with the latest features and advancements provided by Amazon EKS. 

Lastly, if this is all in reference to just the Karpenter sub-module, I would suggest you turn your attention here https://github.com/aws/containers-roadmap/issues/1792

bryantbiggs: > Is this entire module removing support for IRSA? Unfortunately pod Identity is not supported in AWS GovCloud regions: [docs.aws.amazon.com/eks/latest/userguide/pod-identities.html#pod-id-considerations](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html#pod-id-considerations)
> It’s a shame to me to lose out to other module updates due to this hard switch to pod identities

Again, please please please be specific. The root module will still, and as long as EKS supports it, and always support creating an OIDC identity provider which is used for IRSA authentication. Thats staying and not going anywhere, unless EKS drops support for that

If we are just talking about Karpenter, I will revert to say - the Karpenter sub-module current implementation is all TBD currently. Will it or will it not create an IAM role - I don't know. Does that prevent you from using this module?  Again, I don't know without more context. You can use the v19 version which does work

FernandoMiguel: > Lastly, if this is all in reference to just the Karpenter sub-module, I would suggest you turn your attention here https://github.com/aws/containers-roadmap/issues/1792
sure we all want that.
but until it's actually here, we still have to manage it ourselves

kamirendawkins: > Reasons, such as? This is a common pattern, no? Service linked roles, the role used by Terraform to execute applies, etc. I'm not sure how this applies to the direction of Pod Identity (just because you can use a common role doesn't dictate that you have to - we tend to want to move in the direction of more functionality and flexibility, not less - no?)

Like I said, both patterns are valid and no one is debating that. I feel like flexibility would be supporting both authentication mechanisms but perhaps I am wrong.

> Is this true? We have a number of folks using this sub-module with the latest Karpenter version today

Based on #2733 it is hard to tell and we felt that waiting for v20.x made the most sense until now. We will happily test and confirm either way.

> How does this relate to this module? This module is trying to move forward to support the latest features and functionality provided by Amazon EKS. Users are free to update/upgrade as they see fit, but this project is not held back by the pace at which users update/upgrade.

If we use the module to create the IAM Policy that Karpenter is using and it is not compatible with the newer CRDs than we cannot update Karpenter no? That would mean that this module becomes blocking based on that functionality. You indicated that v19.x works with the new CRDs so again, we will happily test to confirm. No one is asking to hold the module back, we are just asking for consideration with regard to how changes are rolled out and impact the end user.

> I feel like we're just fighting change here for the sake of fighting change. If you do not wish to use pod identity, you do not have to - nobody is forcing you. However, we are not going to hold back the development of this module because a few folks do not want to change. We can't satisfy all users - we still have users using very old versions of this module and thats perfectly fine. But this module will always strive to move forward with the latest features and advancements provided by Amazon EKS.

I am sorry if you feel I am fighting a change just to fight a change, as someone who has been happily using this module for years now I am just fighting against breaking the ability for consumers to use the new latest features and advancements provided by the maintainers. If we want to fight a change just to fight a change we can argue about the arbitrary tag that is added in this PR that I still do not understand but we are not, we are discussing a functionality change that impacts the ability for some to use this module.

That being said, if we wish to continue you to use the Karpenter Sub-Module, in its projected state, then technically you are forcing us to either use Pod Identity or change the way in which we manage Karpenter or consume this module in general.

> Lastly, if this is all in reference to just the Karpenter sub-module, I would suggest you turn your attention here https://github.com/aws/containers-roadmap/issues/1792

As it was stated already, until this lands we manage it ourselves so it does not solve this.

bryantbiggs:
awesome, thanks @jbronn - theres a few parts we're waiting on for this release, and one of those is https://github.com/aws/karpenter-provider-aws/issues/5195. I'd prefer to follow the policy set by the upstream project instead of crafting something custom here so we'll see what the Karpenter team comes back with

jbronn:
@bryantbiggs I think we're on the same page, as [my changes](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/e4c6a715ce703a08b7f1696632aabdd59b770720...radiant-maxar:terraform-aws-eks:v20.0.0-alpha) were an attempt to sync with latest [0.33.1 policy](https://raw.githubusercontent.com/aws/karpenter-provider-aws/v0.33.1/website/content/en/preview/upgrading/v1beta1-controller-policy.json) rather than some custom one I made myself.  In particular, there are significant issues with the current policy here that prevent Karpenter working at all:
* Typos in `AllowScopedResourceTagging` and `AllowScopedDeletion` policy statements: they should be using `ResourceTag` instead of `RequestTag` in conditions.
* The `AllowScopedEC2InstanceActionsWithTags` has wrong conditions specified.
* The `AllowInterruptionQueueActions` and `AllowPassingInstanceRole` statements should be referring to the SQS queue and role ARNs instead of hard-coding strings that don't match the managed terraform resources.

variables.tf
```diff
@@ -161,7 +171,7 @@ variable "enable_kms_key_rotation" {
 variable "kms_key_enable_default_policy" {
   description = "Specifies whether to enable the default key policy. Defaults to `false`"
```
pmcgrath, line 172: Tiny one, as the default has changed
Defaults to `false` -> Defaults to `true`

docs/UPGRADE-20.0.md
giedriuskilcauskas: Pod identity is not supported on Fargate, so if using Karpenter with controllers on Fargate, need to continue using IRSA

main.tf
```diff
@@ -468,111 +567,3 @@ resource "aws_eks_identity_provider_config" "this" {
...
-        }
-      ],
-      var.aws_auth_roles
-    ))
-    mapUsers    = yamlencode(var.aws_auth_users)
-    mapAccounts = yamlencode(var.aws_auth_accounts)
-  }
-}
-
-resource "kubernetes_config_map" "aws_auth" {
```
dtscssap, line 545: any reason to not use `moved` blocks here so folks have an easier time upgrading the new major version release?

bryantbiggs: yes - what am I moving it to? 😉 

unfortunately, move blocks are good for refactoring type changes that are all kept within the module, but theres a lot of rough edges around trying to move things that have for_each, moving things in/out of a module, etc. 

we will have guidance around upgrading - fortunately, nothing is control plane impacting, but there will be some one time handling of changing from `aws-auth` to cluster access entries. TBD what those options will be

FernandoMiguel: The upgrade path recommended by Amazon docs is to go with both configMap and api and only then move to api only.

variables.tf
bryantbiggs: thank you! fixed in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2858/commits/570065efe5e960321702f022c969d449c47717f6

main.tf
bryantbiggs: correct - we are defaulting the access entry to `API_AND_CONFIG_MAP` on this change and then we'll be changing that default to `API` only on the next release in the future. Its a one way op so, this is the safest middle group while folks migrate

docs/UPGRADE-20.0.md
bryantbiggs: in v20.x, both pod identity and irsa will be supported - we don't have clear guidance yet on a policy with session tags so the policy will match the upstream project (as its currently written) and I'll keep working with the service teams on different avenues of improving policy support. 

This is one example I am exploring where the module would pull the policy from the projects repository based on the project's version:
- [Example of a project that contains the policy that will be pulled directly from the source](https://github.com/clowdhaus/example-external-policies)
- [Example of consuming that policy in a module that aligns the policy to the projects version](https://github.com/clowdhaus/example-external-policies-consumer)

IAM policies do not have a concept of versions, but with these projects that have a clear and direct link between their API functionality and the permissions required, I firmly believe there needs to be a direct link between the version of the software deployed and the version of the policy used. So if we take this situation where there were a number of API changes and permission changes going from Karpenter 0.28 to say 0.31 or later, when you update your Karpenter controller version, you would be updating the policy in lock-step. You could upgrade/downgrade to any version you like, the the API and the permissions are kept in sync. We also experienced this with the AWS load balancer controller and I'm sure it will keep happening unless there is a way to synchronize these two artifacts

modules/self-managed-node-group/main.tf
dtscssap, line 1: Within this file you have the following logic:

```
resource "aws_iam_role_policy_attachment" "this" {
  for_each = { for k, v in toset(compact([
    "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy",
    "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly",
    var.iam_role_attach_cni_policy ? local.cni_policy : "",
  ])) : k => v if var.create && var.create_iam_instance_profile }
  policy_arn = each.value
  role       = aws_iam_role.this[0].name
}
```

Any chance you could update it to something like the following?
```
resource "aws_iam_role_policy_attachment" "this" {
  for_each = var.create && var.create_iam_instance_profile ? merge(
    {
      AmazonEKSWorkerNodePolicy          = "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy"
      AmazonEC2ContainerRegistryReadOnly = "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly"
    },
    !var.iam_role_attach_cni_policy ? {} : {
      CNI_Policy = local.cni_policy
    }
  ) : {}
  
  policy_arn = each.value
  role       = aws_iam_role.this[0].name
}
```

The use of the `data "aws_partition" "current" {}` resource to assemble the policy ARNs means that if one were to put a `depends_on`  around the entire node-group module invocation you now have dictionary keys not "known til apply" which throws an error. 

e.g. 
```
module "node_group_management" {                                      
  depends_on = [module.some_other_resource_known_after_apply]          
                                                           
  source  = "terraform-aws-modules/eks/aws//modules/self-managed-node-group"
  version = "~> 19.13.1" 
  
  ...
}
```

will result in the following:
```
│ 
│   on .terraform/modules/base_layer_00.eks.node_group_management/modules/self-managed-node-group/main.tf line 750, in resource "aws_iam_role_policy_attachment" "this":
│  750:   for_each = { for k, v in toset(compact([
│  751:     "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy",
│  752:     "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly",
│  753:     var.iam_role_attach_cni_policy ? local.cni_policy : "",
│  754:   ])) : k => v if var.create && var.create_iam_instance_profile }
│     ├────────────────
│     │ local.iam_role_policy_prefix is a string, known only after apply
│     │ var.create is true
│     │ var.create_iam_instance_profile is true
│ 
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will
│ identify the instances of this resource.
│ 
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place apply-time results only in the map values.
│ 
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends on, and then apply a second time to fully
│ converge.
```

bryantbiggs: Why are you using `depends_on`? Using `depends_on` across modules is not recommended and should be avoided https://github.com/hashicorp/terraform/issues/30340#issuecomment-1010202582

dtscssap: i believe the verbiage used in that linked comment was "this can be a convenient shortcut in some cases, but we generally discourage using it." Doesn't sounds as forbidden as you are making it out to be IMO.

I can think of a number of examples where one might wait for a Service Quota request to process or some DNS verification step to complete something or other before bringing up an EKS cluster and/or its node groups and wish to simplify this process with explicit resource dependencies at the module level.

Regardless of the use of `depends_on`, it could also be said that it's generally not recommended and should be avoided to use content potentially "known after apply" as keys in dictionaries if you don't have to ... as is true in this scenario where we do not functionally need to leverage dynamically created policy ARNs as the keys in the `for_each` loop for any reason.

either way, fair enough. closed thread & move on... it was worth a shot 🤷 thanks anyways, keep doing what your doing, love this module! ❤️  looking forward to the next release

README.md
```diff
@@ -416,7 +287,7 @@ We are grateful to the community for contributing bugfixes and improvements! Ple
 
 ## License
 
-Apache 2 Licensed. See [LICENSE](https://github.com/terraform-aws-modules/terraform-aws-rds-aurora/tree/master/LICENSE) for full details.
```
bryantbiggs, line 419: Looks like I messed this up over 3 years ago 😅  https://github.com/terraform-aws-modules/terraform-aws-eks/commit/ee9f0c646a45ca9baa6174a036d1e09bcccb87b1

modules/karpenter/main.tf
```diff
@@ -1,207 +1,406 @@
...
+        variable = "${local.irsa_oidc_provider_url}:aud"
+        values   = ["sts.amazonaws.com"]
+      }
     }
   }
 }
 
-resource "aws_iam_role" "irsa" {
-  count = local.create_irsa ? 1 : 0
+resource "aws_iam_role" "controller" {
```
antonbabenko, line 65: Can we keep previous names for these resources in the submodule? `controller` sounds rather generic and it is not mentioned anywhere in the AWS Premium Support docs AFAIK.

WDYT?

docs/UPGRADE-20.0.md
```diff
@@ -0,0 +1,262 @@
...
+   source  = "terraform-aws-modules/eks/aws//modules/karpenter"
+-  version = "~> 19.21"
++  version = "~> 20.0"
+
+# If you wish to maintain the current default behavior of v19.x
++  enable_irsa             = true
++  create_instance_profile = true
+
+# To avoid any resource re-creation
++  iam_role_name          = "KarpenterIRSA-${module.eks.cluster_name}"
```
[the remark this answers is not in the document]
bryantbiggs: this is the Karpenter sub-module, and it matches the current behavior used today (its just moved up from within the module to here to show how users can avoid re-creating the resources). We already [require the cluster name](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/2cb1fac31b0fc2dd6a236b0c0678df75819c5a3b/examples/karpenter/main.tf#L160) to be passed in today so the behavior should match and work as intended (we need the cluster components created before this sub-module)

docs/UPGRADE-20.0.md
```diff
@@ -0,0 +1,262 @@
...
+Once the `authentication_mode` has been updated, next you will need to remove the configmap from the statefile to avoid any disruptions:
+
+```sh
+terraform state rm 'module.eks.kubernetes_config_map_v1_data.aws_auth[0]'
+terraform state rm 'module.eks.kubernetes_config_map.aws_auth[0]' # include if Terraform created the original configmap
+```
+
+#### ℹ️ Terraform 1.7+ users
+
+If you are using Terraform `v1.7+`, you can utilize the [`remove`](https://developer.hashicorp.com/terraform/language/resources/syntax#removing-resources) to facilitate both the removal of the configmap through code. You can create a fork/clone of the provided [migration module](https://github.com/clowdhaus/terraform-aws-eks-migrate-v19-to-v20) and add the `remove` blocks and apply those changes before proceeding. We do not want to force users onto the bleeding edge with this module, so we have not included `remove` support at this time.
```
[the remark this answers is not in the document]
bryantbiggs: I picked 1.3 due to the `moved` blocks now being used to help with migrating during the upgrade. 1.1 added the `moved` block, but there were a number of improvements to them in 1.3.  It also moves us forward so that one day we can start using some of the other features such as the `remove` block (in this migration we are still providing the CLI command, it would be nice to use this in code for users but that requires 1.7), and maybe one day the testing features. But for now, 1.3 was just picked for the `moved` blocks

examples/karpenter/README.md
```diff
@@ -51,38 +54,37 @@ Note that this example may create resources which cost money. Run `terraform des
...
 ## Resources
 
 | Name | Type |
 |------|------|
 | [helm_release.karpenter](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) | resource |
-| [kubectl_manifest.karpenter_example_deployment](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/manifest) | resource |
-| [kubectl_manifest.karpenter_node_class](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/manifest) | resource |
-| [kubectl_manifest.karpenter_node_pool](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/manifest) | resource |
+| [kubectl_manifest.karpenter_example_deployment](https://registry.terraform.io/providers/alekc/kubectl/latest/docs/resources/manifest) | resource |
+| [kubectl_manifest.karpenter_node_class](https://registry.terraform.io/providers/alekc/kubectl/latest/docs/resources/manifest) | resource |
```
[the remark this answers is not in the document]
bryantbiggs: hmm, you're right - not sure how I fix this, I'll have to see about opening an issue either on the provider or `terraform-docs`, not sure which yet

modules/karpenter/main.tf
bryantbiggs: this I was going to name `this` (per usual), but the problem I ran into is that I can't have a migration of `this` -> `node`, and `irsa` -> `this` - they would just all converge to conflict on `node` eventually. The module is primarily creating resources for the Karpenter controller, hence the name change to that - and theres one additional role that Karpenter requires which is the node IAM role. So thats where the names came from

And controller is a Kubernetes construct - in the Karpenter docs you'll see the term `Karpenter controller` used quite a bit - 
> As someone using Karpenter, once your Kubernetes cluster and the Karpenter controller are up and running

https://karpenter.sh/docs/

antonbabenko: Thanks for the explanation. I didn't think about the case that such 2 step renaming can be a problem for Terraform.

antonbabenko:
This PR is included in version 20.0.0 :tada:

PR #2870, opened 2024-01-11
fix: Don't create a launch template if a custom one is used

alexandre-bordiere:
## Motivation and Context

If `use_custom_launch_template` is set to `true`, then the module should not attempt to create a launch template regardless if the `create_launch_template` is also `true`.

## Breaking Changes

N/A

## How Has This Been Tested?

I am note sure if I need to update one of the examples but I did test locally and we can see that the plan is going from 83 resources to add to 82:
<p align="center"></p>
<p align="center">Before</p>

<p align="center"></p>
<p align="center">After</p>

Another way to see the result is in the outputs: The `launch_template_id` and other properties are `null` because the module is not creating the resource anymore.

<p align="center"></p>
<p align="center">Before</p>

<p align="center"></p>
<p align="center">After</p>

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -34,7 +34,7 @@ locals {
 }
 
 resource "aws_launch_template" "this" {
-  count = var.create && var.create_launch_template && var.use_custom_launch_template ? 1 : 0
+  count = var.create && var.create_launch_template && !var.use_custom_launch_template ? 1 : 0
 
   dynamic "block_device_mappings" {
     for_each = var.block_device_mappings


modules/eks-managed-node-group/main.tf
```diff
@@ -34,7 +34,7 @@ locals {
 }
 
 resource "aws_launch_template" "this" {
-  count = var.create && var.create_launch_template && var.use_custom_launch_template ? 1 : 0
+  count = var.create && var.create_launch_template && !var.use_custom_launch_template ? 1 : 0
```
bryantbiggs, line 37: thats not quite what that variable means - if you want to disable the creation of the launch template, then you set `create_launch_template = false`

Issue #2881, opened 2024-01-19
Add support for configuring access_config to support aws_eks_access_policy_association 


sidewinder12s:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - No 🛑: please wait to file a request until the functionality is avaialble in the AWS provider
  - Yes ✅: please list the AWS provider version which introduced this functionality

Yes: v5.33.0 

## Is your request related to a problem? Please describe.

Cannot use the new eks_access_policy_associations without modifying the cluster config to support them.

## Describe the solution you'd like.

Add support for the `access_config` block on the cluster resource.

## Describe alternatives you've considered.

Can't configure it otherwise

## Additional context

New PR for support in the provider was just landed: https://github.com/hashicorp/terraform-provider-aws/pull/35037

bryantbiggs:
We'll have v20 shortly (should be by Monday)

sidewinder12s:
> We'll have v20 shortly (should be by Monday)

Figured! Just didn't see an issue for this bit of change.

bryantbiggs:
I started adding this but there were some issues that have been patched and will go out on 5.34.0 - so we're going to be a bit longer waiting for that release - apologies
- https://github.com/hashicorp/terraform-provider-aws/pull/35381
- https://github.com/hashicorp/terraform-provider-aws/pull/35391

Bharathkumarraju:
@bryantbiggs  thanks currently we are using eks cluster access management using aws-auth configmap like below we need to migrate to new cluster access management API... can we migrate existing cluster using the terraform module.. all apps go down right?

```
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "~> 19.16.0"

  cluster_name                          = format("%s-${local.region_prefix}-${local.env}-eks", local.cluster_prefix)
  cluster_endpoint_public_access        = false
  cluster_additional_security_group_ids = [data.terraform_remote_state.network-sg.outputs.eks_cluster_sg_id_sin]
  cluster_version                       = local.cluster_version
  cloudwatch_log_group_kms_key_id       = data.terraform_remote_state.kms-cloudwatch-logs-sin.outputs.kms_key_cloudwatch_logs_sin.arn
[...]

PR #2892, opened 2024-02-02
fix: Correct cluster access entry to create multiple policy associations per access entry

bryantbiggs:
## Description
- Correct cluster access entry to create one entry and multiple policy associations. Currently, it is trying to use the flattened logic that was intended for supporting multiple policy associations
	- Want: Per access entry, support multiple policy associations
	- Current: One access entry per policy association, and if multiple policies are provided, it will fail since entry already exists

## Motivation and Context
- Provide code to validate the functionality as well as document the schema for users
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2881#issuecomment-1924502935

## Breaking Changes
- Technically, yes - the looping logic of the access entry has changed. However, this is not working as it was intended and we are coming off a breaking change so I am inclined to add this as a bugfix since it was never working as intended from the last major release

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -168,28 +168,22 @@ locals {
       for pol_key, pol_val in lookup(entry_val, "policy_associations", {}) :
       merge(
         {
-          principal_arn     = entry_val.principal_arn
-          kubernetes_groups = lookup(entry_val, "kubernetes_groups", [])
-          tags              = lookup(entry_val, "tags", {})
-          type              = lookup(entry_val, "type", "STANDARD")
-          user_name         = lookup(entry_val, "user_name", null)
+          principal_arn = entry_val.principal_arn
+          entry_key     = entry_key
+          pol_key       = pol_key
         },
         { for k, v in {
           association_policy_arn              = pol_val.policy_arn
           association_access_scope_type       = pol_val.access_scope.type
           association_access_scope_namespaces = lookup(pol_val.access_scope, "namespaces", [])
         } : k => v if !contains(["EC2_LINUX", "EC2_WINDOWS", "FARGATE_LINUX"], lookup(entry_val, "type", "STANDARD")) },
-        {
-          entry_key = entry_key
-          pol_key   = pol_key
-        }
       )
     ]
   ])
 }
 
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -168,28 +168,22 @@ locals {
       for pol_key, pol_val in lookup(entry_val, "policy_associations", {}) :
       merge(
         {
-          principal_arn     = entry_val.principal_arn
-          kubernetes_groups = lookup(entry_val, "kubernetes_groups", [])
```
bryantbiggs, line 172: `kubernetes_groups`, `tags`, `type`, and `user_name` are used in the access entry, not in the policy association. So these are not used here and can be removed

main.tf
```diff
@@ -168,28 +168,22 @@ locals {
...
-          entry_key = entry_key
-          pol_key   = pol_key
-        }
       )
     ]
   ])
 }
 
 resource "aws_eks_access_entry" "this" {
-  for_each = { for k, v in local.flattened_access_entries : "${v.entry_key}_${v.pol_key}" => v if local.create }
```
bryantbiggs, line 192: this is the issue thats technically breaking, but its not working as intended on v20

antonbabenko: This is acceptable for a quick-fix release like "5 second rule"

https://en.wikipedia.org/wiki/Five-second_rule 🤣

antonbabenko (approved):

antonbabenko:
This PR is included in version 20.0.1 :tada:

hmih:
I am already using this in production lol. Here's an example of a map that I pass to `access_entries` to generate the right resources:

```
  eks_access_entries = {
    for ix, arn in [module.iam.eks_admin_role_arn, module.iam.argocd_role_arn] :
    ix => {
      principal_arn = arn
      policy_associations = {
        admin = {
          policy_arn = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
          access_scope = {
            type = "cluster"
          }
        }
      }
    }
```

It would be nice to type the variable, so I don't have to read the source the understand the shape of the map.

PR #2922, opened 2024-02-08
docs: Lowercase Fargate compute type annotation value

stefanozanella:
## Description
Change the custom annotation on coredns pods to `fargate` instead of `Fargate`.

## Motivation and Context
I just set up a cluster using Fargate for the `kube-system` namespace, and the additional configuration for coredns as expressed by the example subject of this PR failed, because the node taint is against an annotation with value `fargate`.


bryantbiggs (approved):

antonbabenko:
This PR is included in version 20.2.2 :tada:

PR #2940, opened 2024-02-21
docs: Update output for user data example to capture rendered output

bryantbiggs:
## Description
- Update output for user data example to capture rendered output

## Motivation and Context
- Most examples spit output to stdout on the terminal, and that works for 99.99999% of the use cases. Here, the user data sub-module is really used as a template rendering engine. Changing the output of user data can potentially be quite disruptive since it will affect ever instance under management by this module (its a lot!). To help ensure that changes to user data are not introducing unintended rendered result changes, we are adding the rendered output to source control so that we can visualize any resulting changes more easily. I am adding this in its own PR since support for AL2023 is coming in a [follow up PR](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/master...bryantbiggs:terraform-aws-eks:feat/al2023?expand=1) and I want to show that the current user data behavior is unmodified. So this only modifies the example and captures the output in source control which from a module perspective is a no-op

## Breaking Changes
- No

## How Has This Been Tested?


antonbabenko (approved):
🙌

antonbabenko:
This PR is included in version 20.3.0 :tada:

PR #2960, opened 2024-03-08
fix: Update AWS provider version to support `AL2023_*` AMI types; ensure AL2023 user data receives cluster service CIDR

bryantbiggs:
## Description
- Update AWS provider version to support AL2023_* AMI types
	- Provider https://github.com/hashicorp/terraform-provider-aws/pull/36046
	- SDK [aws/aws-sdk-go-v2@release-2024-02-29/service/eks/CHANGELOG.md#v1410-2024-02-29](https://github.com/aws/aws-sdk-go-v2/blob/release-2024-02-29/service/eks/CHANGELOG.md?rgh-link-date=2024-03-01T23%3A24%3A48Z#v1410-2024-02-29)
- Ensure AL2023 user data receives cluster service CIDR
- Fix a bug where setting `create = false` on a managed nodegroup would return an error when looking up the instance type used for EFA

## Motivation and Context
- Resolves #2954
- Relates https://github.com/awslabs/amazon-eks-ami/issues/1597#issuecomment-1973512095
- On AL2 AMI, the service CIDR was optional but starting with the AL2023 AMI, its [now required](https://github.com/awslabs/amazon-eks-ami/issues/1678#issuecomment-1979794606)

## Breaking Changes
- No

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects
- [x] I have executed `pre-commit run -a` on my pull request

image

Issue #2955, opened 2024-03-05
 User data was not in the MIME multipart format with module >20.2.1


maruina:
## Description

Creating a new EKS cluster with a Managed NodeGroup fails with the following error
```
│ Error: waiting for EKS Node Group (green:kube_system-20240305170030320200000001) create: unexpected state 'CREATE_FAILED', wanted target 'ACTIVE'. last error: : Ec2LaunchTemplateInvalidConfiguration: User data was not in the MIME multipart format.
```

- [ ] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: `>20.2.1`

- Terraform version:
```
➜ terraform -version
Terraform v1.6.1
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v5.39.1
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.3
+ provider registry.terraform.io/hashicorp/helm v2.12.1
+ provider registry.terraform.io/hashicorp/kubernetes v2.26.0
+ provider registry.terraform.io/hashicorp/time v0.10.0
+ provider registry.terraform.io/hashicorp/tls v4.0.5
```

## Reproduction Code [Required]
```
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "~> 20.0"

  cluster_name    = "green"
  cluster_version = "1.29"
  tags = {
    "karpenter.sh/discovery" : "green",
  }

  # Access config
  authentication_mode = "API_AND_CONFIG_MAP"
  # Allow account-admin to be cluster admin
  access_entries = {
    account_admin = {
      principal_arn = "arn:aws:iam::111111111111:role/aws-reserved/sso.amazonaws.com/foo"
      policy_associations = {
        compute_admin = {
          policy_arn = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
          access_scope = {
            type = "cluster"
          }
        }
      }
    }
    compute_admin = {
      principal_arn = "arn:aws:iam::111111111:role/aws-reserved/sso.amazonaws.com/bar"
      policy_associations = {
        cluster_admin = {
          policy_arn = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
          access_scope = {
            type = "cluster"
          }
        }
      }
    }
  }

  enable_irsa = true

  vpc_id     = module.vpc.vpc_id
[...]

Issue #2747, opened 2023-09-14
invalid for_each argument


mmerickel:
The issue is exactly [Issue #2681](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2681#issuecomment-1653056775). The dynamic resource names for `local.iam_role_policy_prefix` and `local.eni_policy` are dynamic and are used in the KEY of the for_each statement.

This should be changed (bw-incompat) to statically defined keys for the 3 resources being created.

bryantbiggs:
please provide the details requested/required on the issue template

bryantbiggs:
just looking at the error isn't enough detail - we need to see a reproduction to know what you are doing (or trying to do)

mmerickel:
You are free to ignore this - I put significant effort into trying to make a simple repro and am failing.

Here is what I will leave you with - the terraform does not pass the smell-test - the keys are dynamic based on the value of the partition data source. This is a pitfall of using `toset()` in terraform. I was able to get past the issue in my env by running a `terraform apply -target 'module.eks.module.eks_managed_node_group["foo"].data.aws_partition.current'`. After this, the partition info was loaded successfully and the full plan/apply worked on the next run.

Obviously this doesn't happen every time, it's possible to get past it, but it could clearly be done better in this module such that the resource keys are not dependent on the partition data source.

A proper fix (albeit bw-incompat) is this:

```diff
diff --git a/modules/eks-managed-node-group/main.tf b/modules/eks-managed-node-group/main.tf
index 1d3fe81..3725176 100644
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -287,7 +287,9 @@ resource "aws_launch_template" "this" {
   # Prevent premature access of policies by pods that
   # require permissions on create/destroy that depend on nodes
   depends_on = [
-    aws_iam_role_policy_attachment.this,
+    aws_iam_role_policy_attachment.worker_node,
+    aws_iam_role_policy_attachment.ecr,
+    aws_iam_role_policy_attachment.cni,
   ]
 
   lifecycle {
@@ -430,14 +432,24 @@ resource "aws_iam_role" "this" {
 }
 
 # Policies attached ref https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group
-resource "aws_iam_role_policy_attachment" "this" {
-  for_each = { for k, v in toset(compact([
[...]

Issue #2681, opened 2023-07-12
Error: Invalid for_each argument


taimax13:
## Description

Couurnt v.19.15.3 Users may encounter an error such as Error: Invalid for_each argument - The "for_each" value depends on resource attributes that cannot be determined until applied, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument first to apply - also seen in v 18.8.0. This behavior is not seen in the v18.31.2

- [repetitive from other versions] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

- Module version [Required]:
current: 19.15.3
- Terraform version:
Terraform v1.4.5
- Provider version(s):

## Reproduction Code [Required]

module "node_group" {
  source                            = "terraform-aws-modules/eks/aws//modules/eks-managed-node-group"
  version                           = "19.15.3"
  #version                           = "18.31.2" ###https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2337on - to avoid this issue roll back to version <--
  for_each                          = var.node_groups
  create                            = each.value.create
  name                              = "${each.value.name}-${var.owner}"
  create_iam_role                   = false
  iam_role_arn                      = aws_iam_role.eks_nodes.arn
  cluster_name                      = module.eks.cluster_name
  cluster_primary_security_group_id = module.eks.cluster_primary_security_group_id
  subnet_ids                        = each.value.subnet_ids 
  ## launch tempate version for node group. valid values: any version number. leave blank "" for latest
  launch_template_version = ""
  min_size                = each.value.min_size
  max_size                = each.value.max_size
  desired_size            = each.value.des_size
  disable_api_termination = false
  instance_types          = each.value.instance_types
  key_name                = module.tls_keys_node_groups.key_name
  ## Type of capacity associated with the EKS Node Group. Valid values: ON_DEMAND, SPOT
  vpc_security_group_ids            = [module.eks.node_security_group_id]
  capacity_type = each.value.capacity_type
  ebs_optimized = true
  block_device_mappings = {
    xvda = {
      device_name = "/dev/xvda"
      ebs = {
        volume_size           = 80
        volume_type           = "gp3"
        iops                  = 3000
        throughput            = 150
        encrypted             = true
        kms_key_id            = module.kms_key_ebs.key_arn
        delete_on_termination = true
      }
    }
  }
  metadata_options = {
    http_endpoint               = "enabled"
    http_put_response_hop_limit = "2"
[...]

PR #2962, opened 2024-03-10
feat: Replace the use of `toset()` with static keys for node IAM role policy attachment

bryantbiggs:
## Description
- Use static keys for IAM role policy attachment

## Motivation and Context
The use of `toset()` in `for_each` loops is known to be problematic. In hindsight, this should have been corrected during v20.0 but it must have been an oversight on my part. The use of `toset()` on the static policy attachment (custom/additional policy attachment was added awhile back with static keys) has worked without issue except for scenarios where users try to use an explicit `depends_on` (which is [not recommended](https://github.com/hashicorp/terraform/issues/30340#issuecomment-1010202582) across modules for a number of reasons). 

Why is this an issue now? We have an issue with our user data where the cluster service CIDR is not being passed properly, causing nodes to have the wrong internal IP. Oddly enough, we haven't seen user reports for this prior to #2955, and I first noticed it on https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2960#issue-2176398616 (see screen shot - some internal IPs are IPv6, others are IPv4 - they all should be IPv6). While [drafting a PR to resolve the service CIDR issue](https://github.com/terraform-aws-modules/terraform-aws-eks/compare/master...bryantbiggs:fix/service-cidr?expand=1), the dreaded `Error: Invalid for_each argument` appeared for reasons I can't quite understand (yet):

```
│
│   on ../../modules/eks-managed-node-group/main.tf line 506, in resource "aws_iam_role_policy_attachment" "this":
│  506:   for_each = var.create && var.create_iam_role ? toset(compact([
│  507:     "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy",
│  508:     "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly",
│  509:     var.iam_role_attach_cni_policy ? local.cni_policy : "",
│  510:   ])) : object()
│     ├────────────────
│     │ local.iam_role_policy_prefix is "arn:aws:iam::aws:policy"
│     │ var.create is true
│     │ var.create_iam_role is true
│
│ The "for_each" set includes values derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will identify the instances
│ of this resource.
```

## Breaking Changes

From an API perspective, no. However, there will be a very brief (< 1s-2s) period where permission(s) will be removed and re-attached to the IAM role used by nodes. This has been minimized with the use of Terraform state `move` blocks, but due to limitations of the current `move` functionality, they are unable to support moving from multiple sources to a single target (i.e. - the source will vary based on the partition - `aws`, `aws-cn`, `aws-us-gov`) so only the commercial `aws` region (majority) is supported. In addition, `move` blocks do not support variables/dynamic values (which would also resolve the prior issue), which means that the IPv6 CNI policy attachment, which is not a managed policy and therefore user created and its ARN will contain the account ID, cannot be moved either. However, the majority use case for the commercial `aws` partition, using the managed IPv4 CNI policy, is supported without disruption

The potential disruption while unfortunate, is quite minimal and could be equated to a small blip in network connectivity. Therefore, I am proposing this change be made under the current v20 major version so that we can continue on to resolve the service CIDR issue and avoid introducing further, larger disruptions that would be captured under v21 at a later date. 

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -504,11 +504,12 @@ resource "aws_iam_role" "this" {
 
 # Policies attached ref https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group
 resource "aws_iam_role_policy_attachment" "this" {
-  for_each = { for k, v in toset(compact([
-    "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy",
-    "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly",
-    var.iam_role_attach_cni_policy ? local.cni_policy : "",
-  ])) : k => v if var.create && var.create_iam_role }
+  for_each = { for k, v in {
+    AmazonEKSWorkerNodePolicy          = "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodePolicy"
+    AmazonEC2ContainerRegistryReadOnly = "${local.iam_role_policy_prefix}/AmazonEC2ContainerRegistryReadOnly"
+    AmazonEKS_CNI_IPv6_Policy          = var.iam_role_attach_cni_policy && var.cluster_ip_family == "ipv6" ? local.cni_policy : ""
+    AmazonEKS_CNI_Policy               = var.iam_role_attach_cni_policy && var.cluster_ip_family == "ipv4" ? local.cni_policy : ""
+  } : k => v if var.create && var.create_iam_role && v != "" }
 
   policy_arn = each.value
   role       = aws_iam_role.this[0].name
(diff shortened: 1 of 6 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 20.8.0 :tada:

mmerickel:
Given the comment that this hasn't been reported prior to #2955 and #2960 I'll just note that this is the exact fix requested in #2681 and #2747. Thank you for releasing it.

bryantbiggs:
My comments wasn't around not having seen this type of error reported before on the attach policies, it was that we didn't have a reported issue with a valid configuration. Meaning, the explicit `depends_on` was not a valid configuration. But regardless, it was something that should have been corrected in v20 so hopefully this puts these type of issues to bed 🤞🏽

PR #2963, opened 2024-03-10
fix: Ensure the correct service CIDR and IP family is used in the rendered user data

### LINKED ISSUE #2943: Example Karpenter does not work, karpenter pods are pending (version 20.4.0)
## Description

When the karpenter example is applied, karpenter pods are pending because of taints.
 
When changed the karpenter helm values, it works:
```
resource "helm_release" "karpenter" {

  # values ommited

  values = [
    <<-EOT
    # values ommited
    tolerations:
      - effect: NoSchedule
        operator: "Equal"
        key: eks.amazonaws.com/compute-type
        value: fargate
    EOT
  ]
}
```

## Versions

- Module version: 20.4.0

- Terraform version: 
```
Terraform v1.6.6
```
- Provider version(s):
```
Terraform v1.6.6
on linux_amd64
+ provider registry.terraform.io/gavinbunney/kubectl v1.14.0
+ provider registry.terraform.io/hashicorp/aws v5.38.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.3
+ provider registry.terraform.io/hashicorp/helm v2.12.1
+ provider registry.terraform.io/hashicorp/time v0.10.0
+ provider registry.terraform.io/hashicorp/tls v4.0.5
```

```
> cd examples/karpenter/
> terraform apply -approve
# get kubectl credentials
> k get pods - n karpenter
karpenter     karpenter-b8796757b-7gpr2   0/1     Pending   0   17m
karpenter     karpenter-b8796757b-nvw6k   0/1    Pending  0   17m
```

## Expected behavior

karpenter pods running.

## Actual behavior

karpenter pods in pending.

### LINKED ISSUE #2955: User data was not in the MIME multipart format with module >20.2.1

bryantbiggs:
## Description
- Ensure the correct service CIDR and IP family is used in the rendered user data
- Make `cluster_service_cidr` a hard requirement - because of this change in behavior, users who use disconnected nodegroups (those created outside the root module, using the sub-module directly) could be caught off guard by nodes not joining if they do not add this value. For users of the root module approach (most common), this is a no-op on their part since the module ensures the necessary values are passed through to the child modules correctly
- Fix Fargate profile sub-module to add changes introduced in #2962 - this was missed in that PR
- Bump the Terraform MSV from `1.3.0` to `1.3.2` to pick up patch for nested `precondition` (inside `for_each`) https://github.com/hashicorp/terraform/issues/31846

## Motivation and Context
- Resolves #2943 
- Resolves #2955 
- During #2960, I noticed that not all nodes had an IPv6 internal IP - this is odd given that the cluster is an IPv6 cluster
- With these changes, the user data is now more closely aligned with what EKS managed nodegroups do on their own

Before (#2960):
image

After (this PR):
image

## Breaking Changes
- No

## How Has This Been Tested?


examples/eks_managed_node_group/main.tf
```diff
@@ -388,6 +389,9 @@ module "disabled_eks_managed_node_group" {
   source = "../../modules/eks-managed-node-group"
 
   create = false
+
+  # Hard requirement
+  cluster_service_cidr = ""
```
antonbabenko, line 394: This is the first time we require something to be set for "disabled" modules, as far as I see. Can we still make it optional to be consistent?

bryantbiggs: I've come up with a way to make this only required when `create = true` using `precondition` on a `null_resource` that does nothing (but enforce the check) https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2963/files#diff-87747f89f95b121dfc7a5f6e40ee0a56f1f98449581c02bd34f2171210c2f14dR1-R12

When `create = true` and you do not provide a `cluster_service_cidr` you now get this error which is what we (I) want since it is required when creating a nodegroup:
image

I did have to bump the Terraform MSV from `1.3.0` to `1.3.2` to pick up this patch https://github.com/hashicorp/terraform/issues/31846

modules/_user_data/main.tf
```diff
@@ -1,3 +1,15 @@
+# The `cluster_service_cidr` is required when `create == true`
+# This is a hacky way to make that logic work, otherwise Terraform always wants a value
+# and supplying any old value like `""` or `null` is not valid and will silently
+# fail to join nodes to the cluster
+resource "null_resource" "validate_cluster_service_cidr" {
+  lifecycle {
+    precondition {
+      condition     = var.create ? length(local.cluster_service_cidr) > 6 : true # length(local.cluster_service_cidr) == 0
```
antonbabenko, line 8: Where does `6` come from?

antonbabenko (approved):
Why 6? and LGTM :)

modules/_user_data/main.tf
bryantbiggs: Theres probably a better way, but the gist of the check is that something needs to be provided when `create = true`

bryantbiggs: Cleaned up the commented out code and added a note on the arbitrary value of 6 in https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2963/commits/1f2ef2acfc0fe9762f16a9c47d0810b1996391b7

antonbabenko:
This PR is included in version 20.8.3 :tada:

modules/_user_data/main.tf
antonbabenko: Thanks for the explanation! Something like this is a bit more flexible (probably): `try(cidrhost(local.cluster_service_cidr, 0) != "::", false)`

PR #2975, opened 2024-03-19
fix(karpenter): Allow only https in sqs

ksbde:
## Description
I would like to deny access to AWS SQS through HTTP by AWS SQS access policy due to security reasons.

## Motivation and Context
[SQS security best practices](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-security-best-practices.html)

## Breaking Changes
No

## How Has This Been Tested?


bryantbiggs:
if the upstream Karpenter project supports this, then we can support it here as well

bryantbiggs:
if the upstream project states that it is supported, or its shown in their [example](https://github.com/aws/karpenter-provider-aws/blob/main/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml), then we can support it. but we do not want to diverge - thats caused pain in the past with things like IAM policies

bryantbiggs:
closing for now until upstream guidance is updated

PR #3000, opened 2024-04-08
fix: Forces cluster outputs to wait until access entries are complete

### LINKED ISSUE #2923: Access Entries are deleted before Helm charts so helm charts stays orphan in the terraform state
## Description

We deploy a couple of helm charts inside the same terraform state where the eks module is called. It works like a charm with access entries while creating the resources but it cannot delete the helm resources properly in terraform destroy. The reason is that terraform destroy deletes the access entries before the helm charts.

It was working great in version 19.21.0 with aws-auth configmap

- [x ] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 20.2.1

- Terraform version:1.6.6

- Provider version(s): 1.6.6

- Create the eks cluster without giving any access entries.
- Deploy a dummy helm chart
- Run terrafrom destroy

## Expected behavior

Terraform destroy should run successfully

## Actual behavior

It deletes the access entries and cannot delete the helm charts so helm charts stays idle at the terraform sate

lorengordon:
## Description
This changeset adds `depends_on` to a few cluster outputs. This will tell terraform that these cluster outputs should not be made "known" and available until all the EKS access entries have completed. The outputs selected in this PR are limited to those used for authenticating to the EKS cluster. If it turns out more are needed, it is easy enough to add them, and I am not aware of any further repercussions to this particular usage of `depends_on`.

## Motivation and Context
- Resolves #2923

Without this change, it can be difficult to order the terraform lifecycle on create and destroy when also using the helm, kubectl, and/or kubernetes providers. It requires using `state rm` and `-target` to get things to complete successfully. With this change, you can just apply and destroy normally. See also the comment and issue thread here:

https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2923#issuecomment-2043123795

## Breaking Changes

None

## How Has This Been Tested?

I also ran `terraform apply -auto-approve && terraform destroy -auto-approve` on the ["complete" example in the blueprints-addons project](https://github.com/aws-ia/terraform-aws-eks-blueprints-addons/tree/main/tests/complete), three times successfully, without using any state rm or target steps.

### DIFF
--- a/examples/karpenter/README.md
+++ b/examples/karpenter/README.md
@@ -41,11 +41,6 @@ kubectl delete node -l karpenter.sh/provisioner-name=default
 2. Remove the resources created by Terraform
 
 ```bash
-# Necessary to avoid removing Terraform's permissions too soon before its finished
-# cleaning up the resources it deployed inside the cluster
-terraform state rm 'module.eks.aws_eks_access_entry.this["cluster_creator"]' || true
-terraform state rm 'module.eks.aws_eks_access_policy_association.this["cluster_creator_admin"]' || true
-
 terraform destroy
 ```
 
--- a/examples/outposts/README.md
+++ b/examples/outposts/README.md
@@ -37,9 +37,6 @@ $ terraform apply
 Note that this example may create resources which cost money. Run `terraform destroy` when you don't need these resources.
 
 ```bash
-# Necessary to avoid removing Terraform's permissions too soon before its finished
-# cleaning up the resources it deployed inside the clsuter
-terraform state rm 'module.eks.aws_eks_access_entry.this["cluster_creator_admin"]' || true
 terraform destroy
 ```
 
--- a/outputs.tf
+++ b/outputs.tf
@@ -5,16 +5,31 @@
 output "cluster_arn" {
   description = "The Amazon Resource Name (ARN) of the cluster"
   value       = try(aws_eks_cluster.this[0].arn, null)
+
+  depends_on = [
+    aws_eks_access_entry.this,
+    aws_eks_access_policy_association.this,
+  ]
 }
 
 output "cluster_certificate_authority_data" {
   description = "Base64 encoded certificate data required to communicate with the cluster"
   value       = try(aws_eks_cluster.this[0].certificate_authority[0].data, null)
+
+  depends_on = [
+    aws_eks_access_entry.this,
+    aws_eks_access_policy_association.this,
+  ]
 }
 
 output "cluster_endpoint" {
   description = "Endpoint for your Kubernetes API server"
   value       = try(aws_eks_cluster.this[0].endpoint, null)
+
+  depends_on = [
+    aws_eks_access_entry.this,
+    aws_eks_access_policy_association.this,
+  ]
 }
 
 output "cluster_id" {
@@ -25,6 +40,11 @@ output "cluster_id" {
 output "cluster_name" {
   description = "The name of the EKS cluster"
   value       = try(aws_eks_cluster.this[0].name, "")
+
+  depends_on = [
+    aws_eks_access_entry.this,
+    aws_eks_access_policy_association.this,
+  ]
 }
 
 output "cluster_oidc_issuer_url" {


lorengordon:
Bloody pre-commit hooks, I hates them so much. Fine, hold on, waiting on a test of the change to finish up (EKS is so slow!)...

bryantbiggs:
could you also remove these commands from the example READMEs since they aren't needed anymore (thank goodness)
```
# Necessary to avoid removing Terraform's permissions too soon before its finished
# cleaning up the resources it deployed inside the cluster
terraform state rm 'module.eks.aws_eks_access_entry.this["cluster_creator"]' || true
terraform state rm 'module.eks.aws_eks_access_policy_association.this["cluster_creator_admin"]' || true
```

bryantbiggs (approved):
thank you!

antonbabenko:
This PR is included in version 20.8.5 :tada:

lorengordon:
@llavaud I had seen some funny business on apply when using this eks module in conjunction with the eks-blueprints-addons module, and specifically the `helm_chart` resources. I'm not sure the cause is necessarily the same problem as what this patch fixed though. I think that might be a separate issue, though perhaps revealed by fixing this one. Since destroy is now clean, but somehow apply sometimes requires a second try. Not consistently though. Some kind of race condition, maybe. Or perhaps the helm provider is mishandling its own attributes and authenticating waaaaay early? Or perhaps some attribute is not being threaded through the config as needed for ordering correctly.

Module-level `depends_on` should be avoided at all cost, since it will cause the dependency to be recreated on _any_ change to the module. Better to just apply twice, until we figure out what is really happening...

bryantbiggs:
access entry creation is asynchronous and therefore should be treated as eventually consistent. However, Terraform operates in very much a synchronous manner. Deploying applications via Terraform is not recommended, but the 2nd best alternative is for the providers to implement retries similar to the kubectl provider

lorengordon:
@bryantbiggs Any idea if there is an issue for that on the `helm` provider? Do you have a ref to how the kubectl provider is doing it?

bryantbiggs:
Most likely, and here is the argument https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs#apply_retry_count

lorengordon:
Ahhh, _that's_ what that argument is doing for me (in this config I inherited)! 😁

lorengordon:
can the kubectl provider apply helm charts? hmm....

lorengordon:
Here's an issue to follow on retry support in the helm provider... https://github.com/hashicorp/terraform-provider-helm/issues/1058

lorengordon:
Ok, I think the problem is that the cluster outputs are still becoming available before fargate and node group resources are complete, and providers that deploy resources into those node groups are starting waaay too early. I've got a branch I'm testing that just adds those modules to the same outputs we did in this PR. If that works cleanly a few times, I'll open a PR. Probably get to it Monday...

bryantbiggs:
> becoming available before fargate and node group resources are complete

Thats fine - Kubernetes handles that. Terraform, however, does not when  you have things like `wait = true` on the Helm resource. We most likely will not accept changes related to this because:
1. Hashicorp advises against chaining providers
2. Terraform is not an app deployment tool and will always fall short here
3. Too many eggs in one basket (cluster and apps) is not recommended - large blast radius and non-ideal workflow

lorengordon:
That's all well and good and technically correct I suppose. Hard to argue with. However, the eks-blueprints project, and the examples in this project, both really need to discuss _how_ to actually do that then. It feels like there's a lot of configs that are really misleading about how they work and whether they're even a good idea. Or we could just fix it so it actually works.

Anyway, while my other hypothesis may be one problem. There is another race condition causing the node group to be created before the ENI Config is ready, which results in this error:

```
container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
```

That's even with the vpc-cni configured for "before_compute"...

```
  cluster_addons = {
    vpc-cni = {
      addon_version  = "v1.18.0-eksbuild.1"
      before_compute = true

        env = {
        }
      })
    }
  }
```

bryantbiggs:
If you are using the ENIConfig for secondary CIDR, we just shipped https://aws.amazon.com/about-aws/whats-new/2024/04/amazon-vpc-cni-automatic-subnet-discovery/ which I believe removes the need to set/use the ENIConfig

lorengordon:
I'm reading that article, and the blog post, and trying to understand how to make it actionable in the context of this module, but I think I'm really too green with EKS/k8s to understand what it's saying. I inherited this EKS config, and I'm just trying to make it work more reliably. I know terraform really quite well, and know how to manipulate the DAG to make these implicit dependencies more explicit. But EKS/k8s I do not know at all. :(

bryantbiggs:
tl;dr - before you might create ENIConfigs like:

```yaml
apiVersion: crd.k8s.amazonaws.com/v1alpha1
kind: ENIConfig
metadata: 
  name: $az_1
spec: 
  securityGroups: 
    - $cluster_security_group_id
  subnet: $new_subnet_id_1
```

But now, all you need to do is tag your secondary subnets with:

```hcl
  tags = {
    "kubernetes.io/role/cni" = 1
  }
```

And the VPC CNI will discover these subnets and start utilizing them

lorengordon:
Aha, that was the clue I needed. Ok, thanks @bryantbiggs, sorry I was being so dense! Now I see what is causing the dependency ordering issue. The config I inherited is using the kubectl provider to create that ENIConfig, and the change I was proposing is preventing it from starting until _after_ the node group is created. But the node group config requires that resource _before_ it starts in order to complete successfully. Something like this, then:

1. Create Cluster (aws)
2. Create CNI addon (aws)
3. Create ENIconfig (kubectl)
4. Create Node Group (aws)
5. Create everything else, including remaining addons and helm releases from the blueprints-addons module (aws/helm/kubernetes)

I'll see what I can do using the `terraform_data` resource in my own config to create the dependencies on the provider config inputs, instead of modifying the output dependencies here. I'm pretty sure I can force it to order everything correctly. Only explicit dependency missing then would be some way to create the graph edge from the kubectl resource to the node group resource. It seems currently that is handled just with a waiter (`dataplane_wait_duration`), rather than an attribute of some kind. But I think I have enough to get myself unblocked at least!

bryantbiggs:
Just to be clear, you do not need the ENIconfig if you use the recently released subnet discoverability feature as defined in [this](https://aws.amazon.com/about-aws/whats-new/2024/04/amazon-vpc-cni-automatic-subnet-discovery/) and its subsequent links

lorengordon:
Yep understood. It may be a little academic at this point, but now I just want to see if I can influence the timing of the providers and their resources using `terraform_data` lol. And perhaps I can also create an explicit graph edge for the node group by threading a tag value from the kubectl resource to the node group tags...

lorengordon:
Nifty, yeah this seems to work just fine... For the `kubectl` provider, just use the module outputs since they will be available before the node group. And for other providers, thread the inputs through `terraform_data` in a way that links them to the compute option used in the config (or just use all of them, as I did below). Then all resources that depend on the node group will be created after the node group is ready. And if you need some things to kick off before compute and others after compute, for the same provider type, just use two providers with a provider alias.

```hcl
# Use terraform_data inputs to create resources after compute
provider "kubernetes" {
  host                   = terraform_data.eks_cluster_after_compute.input.cluster_endpoint
  cluster_ca_certificate = base64decode(terraform_data.eks_cluster_after_compute.input.cluster_certificate_authority_data)
  token                  = data.aws_eks_cluster_auth.this.token
}

# Use terraform_data inputs to create resources after compute
provider "helm" {
  kubernetes {
    host                   = terraform_data.eks_cluster_after_compute.input.cluster_endpoint
    cluster_ca_certificate = base64decode(terraform_data.eks_cluster_after_compute.input.cluster_certificate_authority_data)
    token                  = data.aws_eks_cluster_auth.this.token
  }
}

# Use module outputs directly to create resources before compute
provider "kubectl" {
  apply_retry_count      = 10
  host                   = module.eks.cluster_endpoint
  cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
  token                  = data.aws_eks_cluster_auth.this.token
}

# Force a dependency between the EKS cluster auth inputs, and the compute resources
resource "terraform_data" "eks_cluster_after_compute" {
  input = {
    cluster_endpoint                   = module.eks.cluster_endpoint
    cluster_certificate_authority_data = module.eks.cluster_certificate_authority_data

    fargate_profiles         = module.eks.fargate_profiles
    eks_managed_node_groups  = module.eks.eks_managed_node_groups
    self_managed_node_groups = module.eks.self_managed_node_groups
  }
}
```

And it does actually work to thread tags back through from the `kubectl_manifest` to the node group. This will create the graph edge between the kubectl resource and the node group, so terraform can order actions correctly.

```hcl
module "eks" {
  version = "~> 20.8.5"
  # ...
  eks_managed_node_groups = {
    foo = {
      # ...
      tags = merge(
        { for label, config in kubectl_manifest.eni_config : "eni-config/${label}" => config.id },
        local.tags
      )
    }
  }
}
```

Though, for the moment, I chose not to thread the tags through that way because of the potential for recursive references, instead just relying on `dataplane_wait_duration`. If I really wanted to establish this graph edge, I would probably not use the top-level node-group/compute options in the module, and instead separately call the node-group submodule. Makes the dependencies easier to visualize and manipulate.

PR #3004, opened 2024-04-10
fix(karpenter) include primary_security_group_tags variable

gritzkoo:
## Description

Following the [karpenter docs](https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/#5-create-nodepool) the primary security group needs a tag `karpenter.sh/discovery` and with this new `variable` is possible to include only on this security group.

## Motivation and Context

Using this module to create eks cluster and karpenter, the nodepool is not finding the security group because it`s not possible to include only on this security group.
## How Has This Been Tested?


bryantbiggs:
we already have `cluster_tags` if you wish to add this there, but the recommendation is to use `node_security_group_tags` and use that security group which provides the ability to control and restrict security group rules

PR #3028, opened 2024-05-07
chore: Update locals computed names for consistency

xakraz:
## Description

Replace suffix '-cluster' with '-eks' for consistency like it is already used by 'launch_template_name', 'managed_node_groups''s 'iam_role_name', 'aws_iam_openid_connect_provider', and more

## Motivation and Context

Resources naming consistency

## Breaking Changes

I don't think so

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -255,7 +255,7 @@ module "kms" {
 ################################################################################
 
 locals {
-  cluster_sg_name   = coalesce(var.cluster_security_group_name, "${var.cluster_name}-cluster")
+  cluster_sg_name   = coalesce(var.cluster_security_group_name, "${var.cluster_name}-eks")
   create_cluster_sg = local.create && var.create_cluster_security_group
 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
this would be a breaking change and we have named it as such since there are two security groups created in the module so there is a distinction. for now, we don't plan on making any change related to this

xakraz:
I can understand your point, nevertheless naming is somewhat confusing.
I would prefer `-eks-cluster` rather than just `cluster` and `-eks-node` rather than `-node` as suffixes.

Otherwise, from what I can observe, people using this module tend to add the `eks` word in the cluster name and we end with some resources named `****-eks-eks` 🙃 

Regarding the breaking change, "maybe" it can be acceptable if the resources are deleted and recreated with "just" a different name.

bryantbiggs:
you have the variables such as `var.cluster_security_group_name` to fully control that name, thats why they are provided.

PR #3030, opened 2024-05-07
feat: Propagate `ami_type` to self-managed node group; allow using `ami_type` only

bryantbiggs:
## Description
- Propagate the `ami_type` down to the self-managed nodegroup. This allows users to specify just the `ami_type` instead of needing to specify the `ami_type` and the `platform` (which is redundantly annoying). The `platform` functionality is still preserved for backwards compatibility until its removed in the next major version release
- Fix the `cluster_service_cidr` output; previously it would output either the IPv4 CIDR or `null`, which is incorrect for IPv6 clusters

## Motivation and Context
- This simplifies and unifies the user experience across both EKS managed node groups and self-managed node groups, it also allow users to opt into the v21 functionality now instead of waiting which will future proof their configurations

## Breaking Changes
- No; validated by deploying EKS managed node group and self-managed node group examples on `master` and then switching to this branch to check the diff

## How Has This Been Tested?

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -1,16 +1,56 @@
 data "aws_partition" "current" {}
 data "aws_caller_identity" "current" {}
 
-data "aws_ami" "eks_default" {
-  count = var.create && var.create_launch_template ? 1 : 0
+################################################################################
+# AMI SSM Parameter
+################################################################################
 
-  filter {
-    name   = "name"
-    values = ["amazon-eks-node-${var.cluster_version}-v*"]
+locals {
+  # Just to ensure templating doesn't fail when values are not provided
+  ssm_cluster_version = var.cluster_version != null ? var.cluster_version : ""
+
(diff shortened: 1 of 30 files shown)


modules/self-managed-node-group/main.tf
```diff
@@ -1,16 +1,56 @@
...
 
-  filter {
-    name   = "name"
-    values = ["amazon-eks-node-${var.cluster_version}-v*"]
+locals {
+  # Just to ensure templating doesn't fail when values are not provided
+  ssm_cluster_version = var.cluster_version != null ? var.cluster_version : ""
+
+  # TODO - Temporary stopgap for backwards compatibility until v21.0
+  ami_type_to_user_data_type = {
```
bryantbiggs, line 13: in v21, we can look at consolidating these repeated lookup maps in another helper sub-module like the user-data. but for now, a little bit of duplication is ok while we roll it out and ensure we maintain backwards compatibility

antonbabenko (approved):

antonbabenko:
This PR is included in version 20.9.0 :tada:

PR #3041, opened 2024-05-20
feat(eks): Add availability zone filter for nodegroup modules to enable update in place

Josephuss:
## Description

Additional dynamic availability zone filters for eks-managed and self managed nodegroup modules to enable update in place during runs.

The following changes makes this behaviour more streamlined, so that:

- placement groups automatically restrict the subnets to a single A/Z
- the A/Z is specifically defined by the user.

## Motivation and Context

Deployment for nodegroups with placement groups are working, but they are subsequently difficult to update.

This is because the placement group restricts the available A/Z to a single one, and a full multiple A/Z group of subnets is still passed to the autoscaling group, leading to a conflict and messages like

_│ Error: updating EKS Node Group version: operation error EKS: UpdateNodegroupVersion, https response error StatusCode: 400, RequestID: 58562857-****-****-****-************, InvalidRequestException: Instances in the <placement-group-name> Placement Group must be launched in the eu-west-1a Availability Zone. Specify the eu-west-1a Availability Zone and try again.
│_ 

Once implemented, there’s no need to put a specific overridden subnet references in our config like this:

        create_placement_group: true
        subnet_ids:
         - subnet-0fa7d************
         
Config changes can now be filtered like this:
     
        create_placement_group: true
        subnet_az_filter: "eu-west-1a"
       

## Breaking Changes

None

## How Has This Been Tested?


bryantbiggs:
please start with an issue and a reproduction - its not clear what the issue is or what this solution is intended to provide

Issue #3025, opened 2024-05-02
Terraform tries to recreate clusters previously using ConfigMap auth 


igorbrites:
## Description

I'm trying to upgrade my clusters to use v20, which all use the `aws-auth` configmap. But now, as `bootstrap_cluster_creator_admin_permissions` is fixed to `false`, even if I try setting `authentication_mode = "API_AND_CONFIG_MAP"`, Terraform says it needs to recreate the whole cluster.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. ✅ Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. ✅ Re-initialize the project root to pull down modules: `terraform init`
3. ✅ Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: `git@github.com:clowdhaus/terraform-aws-eks-migrate-v19-to-v20.git?ref=3f626cc493606881f38684fc366688c36571c5c5`

- Terraform version: 1.7.4

- Provider version(s): 

```bash
+ provider registry.terraform.io/cloudposse/utils v1.22.0
+ provider registry.terraform.io/gavinbunney/kubectl v1.14.0
+ provider registry.terraform.io/hashicorp/aws v5.47.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.4
+ provider registry.terraform.io/hashicorp/helm v2.13.1
+ provider registry.terraform.io/hashicorp/kubernetes v2.29.0
+ provider registry.terraform.io/hashicorp/random v3.6.1
+ provider registry.terraform.io/hashicorp/time v0.11.1
+ provider registry.terraform.io/hashicorp/tls v4.0.5
```

## Reproduction Code [Required]

```hcl
module "eks" {
  # source  = "terraform-aws-modules/eks/aws"
  # version = "~> 19.0"
  source  = "git@github.com:clowdhaus/terraform-aws-eks-migrate-v19-to-v20.git?ref=3f626cc493606881f38684fc366688c36571c5c5"

  cluster_name    = local.cluster_name
  cluster_version = var.cluster_version

  cluster_endpoint_private_access = true
  cluster_endpoint_public_access  = true

  create_kms_key = true
  cluster_encryption_config = {
    resources = ["secrets"]
  }

  vpc_id     = var.vpc_id
[...]

PR #3042, opened 2024-05-21
fix: Ignore changes to `bootstrap_cluster_creator_admin_permissions` which is disabled by default 

wevanscfi:
## Description
Ignore changes to bootstrap_cluster_creator_admin_permissions to allow upgrading module and importing existing clusters

## Motivation and Context
Fixes: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3025

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -102,6 +102,12 @@ resource "aws_eks_cluster" "this" {
     aws_cloudwatch_log_group.this,
     aws_iam_policy.cni_ipv6_policy,
   ]
+
+  lifecycle {
+    ignore_changes = [
+      accecss_config.bootstrap_cluster_creator_admin_permissions
+    ]
+  }
 }
 
 resource "aws_ec2_tag" "cluster_primary_security_group" {
--- a/main.tf
+++ b/main.tf
@@ -105,7 +105,7 @@ resource "aws_eks_cluster" "this" {
 
   lifecycle {
     ignore_changes = [
-      access_config.bootstrap_cluster_creator_admin_permissions
+      access_config["bootstrap_cluster_creator_admin_permissions"]
     ]
   }
 }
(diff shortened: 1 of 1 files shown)


igorbrites:
This fix is way simpler than I was anticipating! Thanks @wevanscfi!

main.tf
```diff
@@ -102,6 +102,12 @@ resource "aws_eks_cluster" "this" {
     aws_cloudwatch_log_group.this,
     aws_iam_policy.cni_ipv6_policy,
   ]
+
+  lifecycle {
+    ignore_changes = [
+      accecss_config.bootstrap_cluster_creator_admin_permissions
```
igorbrites, line 108: Just a typo
```suggestion
      access_config.bootstrap_cluster_creator_admin_permissions
```

main.tf
```diff
@@ -102,6 +102,12 @@ resource "aws_eks_cluster" "this" {
     aws_cloudwatch_log_group.this,
     aws_iam_policy.cni_ipv6_policy,
   ]
+
+  lifecycle {
+    ignore_changes = [
+      access_config["bootstrap_cluster_creator_admin_permissions"]
```
bryantbiggs, line 108:
```suggestion
      access_config.bootstrap_cluster_creator_admin_permissions
```

antlsung:
❤️

main.tf
wevanscfi: Since access_config is a block, bootstrap_cluster_creator_admin_permissions must be referenced using the index, not an attribute.

I assumed your suggested syntax initially as well, however it results in the following error. 
```
Block type "access_config" is represented by a list of objects, so it must be indexed using a numeric key, like .access_config[0].
```

bryantbiggs: oh, weird. ok, will go with what you've got then!

bryantbiggs (approved):
thank you!

PR #3069, opened 2024-06-13
docs: Move examples that are more like test cases to the new `tests/` directory; add better example configurations

bryantbiggs:
## Description
- Move examples that are more like test cases to the new `tests/` directory
- Replace "test" configurations with better example configurations
- Change `nodegroup` to `node group`

## Motivation and Context
- The use of the `examples/` directory has been historically a mix of both demonstrating the various ways in which the provided module(s) can be configured, but also as a means of testing and validating changes. For smaller, more simple module this works ok - but for a larger, complex module such as EKS with all of the various ways one could configure a cluster, its a nightmare for users looking to get started. This PR is the start of splitting apart those two concepts:
	- Test configurations will be stored under `tests/` directory, and in the future we can explore the use of utilizing the native Terraform testing facilities. These configurations are quite elaborate and far from what you should ever expect to see in the wild - we are aiming to cover the full breadth of functionality that the module(s) provide in this directory in order to test and validate that provided functionality
	- Example configurations will be stored under the `examples/` directory. Now, when users look to the example configurations provided, those examples should less complex than what was previously provided and therefore more realistic in terms of a starting point for a cluster

There will be subsequent PRs that will expand on this change as we look to improve the documentation around the module's usability. For example, the next PR will look to introduce a simple mkdocs website where we can start to build out better and more comprehensive documentation around things such as:
- EKS addons and their configuration, how the version is inferred, etc.
- Configuring storage for nodes
- Explain the use of custom launch template versus the default launch template with EKS managed node group
- The process and behavior of upgrading a cluster's Kubernetes version with this module
- Autoscaling of nodes and how that affects/interacts with this module (or vice versa)

## Breaking Changes
- No

## How Has This Been Tested?


antonbabenko (approved):
Looks good to me if you think that such separation helps you manage it.

antonbabenko:
This PR is included in version 20.15.0 :tada:

PR #3072, opened 2024-06-16
fix: Allow desired_capacity_type be defined in self managed module

sergsoares:
fix: Allow desired_capacity_type be defined in self managed module

## Description
Allow the [desired_capacity_type](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#desired_capacity_type) be defined in the self managed variables. 

## Motivation and Context
Terraform apply are failing for self managed groups when the autoscaling desired state are changed outside terraform because the desired_capacity_type become null and we can't enforce it be "units".

```
~ resource "aws_autoscaling_group" "this" {
      - desired_capacity_type            = "units" -> null
        id                               = "eks-nodes-20240612022550573900000003"
      ~ max_size                         = 2 -> 3
        name                             = "eks-nodes-20240612022550573900000003"
        # (26 unchanged attributes hidden)

        # (10 unchanged blocks hidden)
    }
```
Leaves default value of desired_capacity_type for _units_ for autoscaling groups.

It fixed the issue we had with https://github.com/hashicorp/terraform-provider-aws/issues/37523

## Breaking Changes
None

## How Has This Been Tested?


bryantbiggs:
this looks like a provider level issue, not a module level issue

sergsoares:
I will study open a PR in provider code too then.

But has any improvement that can be done yet ? because [desired_capacity_type in autoscaling group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#desired_capacity_type) has multiples values that can be provided ("units", "vcpu", "memory-mib").

It can benefits too who want to customize node groups with self managed too, if not we can close the PR.

bryantbiggs:
Does cluster autoscaler know how to use those other units? I don't believe so, so it doesn't make sense to add those here

bryantbiggs:
I'm going to close this for now since it doesn't make sense in the context of EKS today

Issue #3044, opened 2024-05-22
Unable to update Node Groups in place with cluster placement group strategy without EFA


Josephuss:
## Description

When using a node group *without EFA enabled* and a placement group with cluster strategy, updates of the node group fail because the auto scaling group does not restrict the list of availability zones.

The node group with cluster placement successfully gets deployed into a single AZ of the 3 subnets configured with no errors, However, replacing or upgrading the node group fails as listed below due to the availability zone is not filtered and the update does not take place unless the subnet id is overridden in the configs.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]:  20.8.5

- Terraform version: 1.5.7

- Provider version(s): 5.40.0

## Reproduction Code 

This is a copy of the managed node group example with a placement group created. The default placement group strategy is cluster.

```hcl
module "eks_managed_cluster_node_group" {
  source = "../../modules/eks-managed-node-group"

  name                 = "managed-cluster-node-group"
  cluster_name         = module.eks.cluster_name
  cluster_ip_family    = module.eks.cluster_ip_family
  cluster_service_cidr = module.eks.cluster_service_cidr

  subnet_ids                        = module.vpc.private_subnets
  cluster_primary_security_group_id = module.eks.cluster_primary_security_group_id
  vpc_security_group_ids = [
    module.eks.node_security_group_id,
  ]
  
  ami_type = "BOTTLEROCKET_x86_64"
  platform = "bottlerocket"

  create_placement_group = true

  # this will get added to what AWS provides
  bootstrap_extra_args = <<-EOT
    # extra args added
    [settings.kernel]
    lockdown = "integrity"

    [settings.kubernetes.node-labels]
    "label1" = "foo"
    "label2" = "bar"
  EOT

  tags = merge(local.tags, { Separate = "eks-managed-node-group" })
}
```

1.  Create a managed node group of size 1 and create a placement group using cluster strategy (default).
2. Increase the size to 2.

## Expected behavior

Node group is increased as requested.

## Actual behavior

Node group update may fail with:
> Error: updating EKS Node Group version: operation error EKS: UpdateNodegroupVersion, https response error StatusCode: 400, RequestID: 58562857----********, InvalidRequestException: Instances in the Placement Group must be launched in the eu-west-1c Availability Zone. Specify the eu-west-1c Availability Zone and try again.

## Additional context

* EFA is not enabled
* Placement group is created
[...]

PR #3045, opened 2024-05-23
feat: Enable update in place for node groups with cluster placement group strategy

Josephuss:
## Description

When using a node group without EFA enabled and a placement group with cluster strategy, sometimes updates of the node group fail because the auto scaling group does not restrict the list of availability zones.

The following changes fixes this behaviour so that:

* placement groups automatically restrict the subnets to a single A/Z
* the A/Z is specifically defined by the user.

## Motivation and Context

**This fixes the issue mentioned in https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3044**

Deployment for nodegroups with placement groups are working, but they are subsequently difficult to update.

This is because the placement group restricts the available A/Z to a single one, and a full multiple A/Z group of subnets is still passed to the autoscaling group, leading to a conflict and messages like

_│ Error: updating EKS Node Group version: operation error EKS: UpdateNodegroupVersion, https response error StatusCode: 400, RequestID: 58562857-****-****-****-************, InvalidRequestException: Instances in the <placement-group-name> Placement Group must be launched in the eu-west-1a Availability Zone. Specify the eu-west-1a Availability Zone and try again.
│_ 

Once implemented, there’s no need to put a specific overridden subnet references in our config like this:

        create_placement_group: true
        subnet_ids:
         - subnet-0fa7d************
         
Config changes can now be filtered like this:
     
        create_placement_group: true
        placement_group_az_filter: "eu-west-1a"
       

## Breaking Changes
None

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #3090, opened 2024-07-05
feat: Add subnet AZ filter

bersalazar:
## Description

Allows deployment of instances in a node group to be restricted to a list of availability zones, specified as a "subnet AZ filter".

Applies to both EKS-managed and self-managed node groups.

An example of the proposed interface for both cases:

```
 eks_managed_node_groups = {
    my_nodegroup = {
      name            = "my-nodegroup"
      subnet_az_filter = ["eu-west-1a"]
    }
  }
```

Resolves https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3044

Co-Authored-By: @Josephuss

## Motivation and Context

When using a node group with placement group and 'cluster' strategy, node group updates via autoscaling group (such as scale out and/or scale up) fail because the autoscaling group does not restrict the subnet to the AZ where the cluster placement group is set.

Example:

    Error: updating EKS Node Group version: operation error EKS: UpdateNodegroupVersion, https response error StatusCode: 400, RequestID: 58562857----********,
    InvalidRequestException: Instances in the Placement Group must be launched in the eu-west-1a Availability Zone.
    Specify the eu-west-1a Availability Zone and try again.

This is currently manageable by [overriding the subnet ids in the node group configuration](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L308), for example:

```
eks_managed_node_groups = {
  my_custom_nodegroup = {
    name            = "customer1"
    subnet_ids = ["subnet-abc12345"]
  }
}
```

However, this is fragile because subnet IDs **change** after full destroy/create cycles that include re-creating the VPC, which is common in development environments. This requires manual intervention to update the subnet ID in the node group configuration.

Also, specifying subnets by their logical name is significantly easier than trying to use the raw subnet ID.

Placement group update is the one current use case for this feature, but subnet AZ filter is not limited to usage of placement groups.
It can be used without placement groups, in cases where the user wants to restrict the node group to a specific AZ.

## Breaking Changes

This doesn't break backwards compatibility with the current major version.

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/README.md
+++ b/modules/eks-managed-node-group/README.md
@@ -97,7 +97,7 @@ module "eks_managed_node_group" {
 | [aws_iam_policy_document.role](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
 | [aws_partition.current](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition) | data source |
 | [aws_ssm_parameter.ami](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ssm_parameter) | data source |
-| [aws_subnets.efa](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnets) | data source |
+| [aws_subnets.subnets](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnets) | data source |
 
 ## Inputs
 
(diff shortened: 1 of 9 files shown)


bdellegrazie:
This PR supersedes #3045

bryantbiggs:
thank you but this is not correct - this is a breaking change and conflicts with the EFA support. we will investigate issues with updating node groups when a placement group is utilized in a separate PR

bdellegrazie:
@bryantbiggs could you please explain how this breaks EFA support?

The "EFA" data source is renamed but that's a data source so not a breaking change, the filters are conditional upon whether EFA or cluster placement group is in use and your existing test case persists unmodified. If _no_ filters are in use, the default set of all subnets is returned and everything works as normal.

james-masson:
> we will investigate issues with updating node groups when a placement group is utilized in a separate PR

Can you expand on this please?

We've already reported the issues with Cluster placement-groups and multi-AZ ASGs to AWS support and our partnership contacts.

AFAIK what we're seeing is expected behaviour currently for AWS, and should be fixed client-side in the config. Hence the PR trying to help this code do the right thing.

PR #3111, opened 2024-08-05
Ignore desired_capacity_type

angr1ff:
Module ignore change desired_capacity_type from "unit" to null


bryantbiggs:
thank you but this this is not something we are going support at this time

Issue #2760, opened 2023-10-03
Tags with values known only at apply time fail to plan/apply


archoversight:
## Description

When using a mapping of tags that contains a value that is only known at apply time (such as `time_static` for things like `creation_date` as required by our auditors), `terraform plan`/`terraform apply` both fail.

If your request is for a new feature, please use the `Feature request` template.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 19.16.0
- Terraform version: Terraform v1.5.7
- Provider version(s):
  + provider registry.terraform.io/hashicorp/aws v5.19.0
  + provider registry.terraform.io/hashicorp/cloudinit v2.3.2
  + provider registry.terraform.io/hashicorp/kubernetes v2.23.0
  + provider registry.terraform.io/hashicorp/time v0.9.1
  + provider registry.terraform.io/hashicorp/tls v4.0.4

## Reproduction Code [Required]

```
provider "aws" {
  region = local.region
}

data "aws_eks_cluster_auth" "this" {
  name = module.eks.cluster_name
}

provider "kubernetes" {
  host                   = module.eks.cluster_endpoint
  cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
  token                  = data.aws_eks_cluster_auth.this.token
}

data "aws_caller_identity" "current" {}
data "aws_availability_zones" "available" {}

resource "time_static" "creation_date" {}

locals {
  name            = "ex-${replace(basename(path.cwd), "_", "-")}"
  cluster_version = "1.27"
  region          = "us-east-2"

  vpc_cidr = "10.0.0.0/16"
  azs      = slice(data.aws_availability_zones.available.names, 0, 3)

  tags = {
    Example    = local.name
    GithubRepo = "terraform-aws-eks"
    GithubOrg  = "terraform-aws-modules"
    CreationDate = time_static.creation_date.rfc3339
  }
}

################################################################################
# EKS Module
################################################################################

module "eks" {
  source = "terraform-aws-modules/eks/aws"

  cluster_name                   = local.name
  cluster_version                = local.cluster_version
  cluster_endpoint_public_access = true

  # IPV6
  cluster_ip_family = "ipv6"

  cluster_addons = {
  }

  vpc_id                   = module.vpc.vpc_id
  subnet_ids               = module.vpc.private_subnets
  control_plane_subnet_ids = module.vpc.intra_subnets

[...]

PR #3119, opened 2024-08-09
fix: Eliminates null check on tag values to fix for_each error about unknown *keys*

### LINKED ISSUE #3118: Using terraform <1.6.0, `aws_ec2_tag` with dynamic tag *values* results in for_each error about unknown *keys*
## Description

When using Terraform <1.6.0, a tag value that includes a dynamically interpolated function, like `timestamp()`, will result in an error of the form:

```
│ 
│   on .terraform/modules/eks.cluster/main.tf line 111, in resource "aws_ec2_tag" "cluster_primary_security_group":
│  111:   for_each = { for k, v in merge(var.tags, var.cluster_tags) :
│  112:     k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags && v != null
│  113:   }
│     ├────────────────
│     │ local.create is true
│     │ var.cluster_tags is empty map of string
│     │ var.create_cluster_primary_security_group_tags is true
│     │ var.tags is map of string with 7 elements
│ 
│ The "for_each" map includes keys derived from resource attributes that
│ cannot be determined until apply, and so Terraform cannot determine the
│ full set of keys that will identify the instances of this resource.
│ 
│ When working with unknown values in for_each, it's better to define the map
│ keys statically in your configuration and place apply-time results only in
│ the map values.
│ 
│ Alternatively, you could use the -target planning option to first apply
│ only the resources that the for_each value depends on, and then apply a
│ second time to fully converge.
```

Example setup for tags. We add an expiration to all resources, sorta like the below, but a little different to avoid a persistent diff. This does reproduce though:

```
  tags = {
    expiration = formatdate("YYYY-MM-DD", timeadd(timestamp(), "24h"))
  }
```

Actually, if you [search for `aws_ec2_tag`](https://github.com/terraform-aws-modules/terraform-aws-eks/issues?q=sort%3Aupdated-desc+is%3Aissue+aws_ec2_tag), there are _lots_ of issues all saying the same thing. Difference here is that I know why it is occurring! :D 

Valid forms of this error pertain only to _keys_. But this is a value. Dynamic _values_ should be fine in any for_each expression, so obviously the error is bogus, and represents a bug in Terraform. And this is actually fixed in Terraform 1.6.0 and newer, but we're still pinned to 1.5.7 for now due to the license change.

Still, I was curious, why was this happening at all, and can we modify the module to avoid the bug? Looking at the module code:

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/5fe865e860c4cc8506c639f2e63bc25e21a31b37/main.tf#L126-L128

I couldn't understand why the check was there for `v != null`. And since this is happening for _values_, it made me suspicious. So I removed the check. And the error went away. Hurrah!

I wanted to understand why the check was there, so I looked into the blame for the file, and saw the check was introduced in [PR 2250](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2250). I see [@antonbabenko commented](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2250#discussion_r1026533675) in the review, suggesting the check, but no particular explanation. Just maybe a sense that `null` would be invalid anyway, so might as well remove it? I can't think of any valid reason a tag value would be `null`. So to me it just seems like user error. Since the check causes problems (for many folks, going off the number of issues in the search), I'm hoping a PR would be accepted to remove the check.

lorengordon:
## Description
See linked issue #3118

## Motivation and Context

Fixes #3118

Related #2760 
Related #2681 
Related #2337

## Breaking Changes

None

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -124,7 +124,7 @@ resource "aws_ec2_tag" "cluster_primary_security_group" {
   # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2006
   # Ref: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2008
   for_each = { for k, v in merge(var.tags, var.cluster_tags) :
-    k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags && v != null
+    k => v if local.create && k != "Name" && var.create_cluster_primary_security_group_tags
   }
 
   resource_id = aws_eks_cluster.this[0].vpc_config[0].cluster_security_group_id


bryantbiggs (approved):
seems ok to try this

antonbabenko:
This PR is included in version 20.22.1 :tada:

lorengordon:
Confirmed everything is working for us, using Terraform 1.5.7! Thanks!

PR #3142, opened 2024-08-28
Ignore `platform_version` on EKS Cluster

jolexa:
## Description
This PR ignores the `platform_version` value of an EKS Cluster. Per [documentation](https://docs.aws.amazon.com/eks/latest/userguide/platform-versions.html), AWS automatically updates existing clusters to the latest version for the corresponding Kubernetes version.

## Motivation and Context
At first glance, a proposed TF change for `platform_version` in a plan is a moment of concern. When you later find out that AWS automatically updates the version, then the concern is minimimized (no-op), I suggest ignoring this output from the AWS API. 

Note: I considered submitting this idea to `terraform-provider-aws` but deemed it was inappropriate because _this_ module is opinionated by the fact that it does **not** allow consumers to set the platform_version. This signals to me that consumers of this module are 'ok' with just letting AWS do the automatic upgrades.

For example, this is the output of a TF plan
image

## Testing
I am unsure how to test this change because new clusters are automatically at the latest version and I no longer have a test case.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -114,7 +114,8 @@ resource "aws_eks_cluster" "this" {
 
   lifecycle {
     ignore_changes = [
-      access_config[0].bootstrap_cluster_creator_admin_permissions
+      access_config[0].bootstrap_cluster_creator_admin_permissions,
+      platform_version
     ]
   }
 }


bryantbiggs:
this is a computed only attribute - we do not ignore those. I would suggest revisiting your configurations to identify why its showing up in your diffs

jolexa:
@bryantbiggs understood, no problem! Quick followup if you have a moment?

> this is a computed only attribute

I believe I understand, you are referring to arguments vs attributes and you are suggesting that ignore_changes is only applicable to arguments.

> I would suggest revisiting your configurations to identify why its showing up in your diffs

I'm not sure what you mean by "configurations" - any hints? My understanding as to why it shows up in diffs is because terraform records `platform_version` in the statefile (confirmed this is true via `terraform state show 'module.cluster.module.cluster.aws_eks_cluster.this[0]'`). Is my understanding how how a diff is generated not accurate? 

I'm very curious now, [terraform documentation](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) does specify that ignore_changes operates on attribute names (not arguments, specifically)

bryantbiggs:
> I believe I understand, you are referring to arguments vs attributes and you are suggesting that ignore_changes is only applicable to arguments.

Correct

> I'm not sure what you mean by "configurations" - any hints?

Based on how you have configured your Terraform, it may have an effect on showing spurrious diffs like this. I can't say though because I don't have a reproduction

jolexa:
> Based on how you have configured your Terraform

Right! That's what I'm asking, I have no idea how to configure this. As far as I know, this is default behavior that I'm showing. Maybe, just maybe, this is a bug in the terraform-aws-provider, afterall??

ajax-ryzhyi-r:
@bryantbiggs We face the same issue. Every time AWS updates the  platform version for EKS clusters, we encounter drift in Terraform:
Screenshot 2024-08-30 at 11 21 53

It would be nice to see it merged

heshdotcc:
We're facing the very same situation here...

Would being able to fix such `platform_version` through a new module property be an acceptable fix?

Either way, we still wonder what configuration doesn't show this behavior.

Thanks,

bryantbiggs:
> Would being able to fix such platform_version through a new module property be an acceptable fix?

This is not something configurable. 

The image shown here https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3142#issuecomment-2320467888 shows that you are using the `cluster_platform_version` in some way - how are you using it? are you re-wrapping this entire module?

PR #3166, opened 2024-09-26
feat: Add support for `desired_capacity_type` (named `desired_size_type`) on self-managed node group

dtscssap:
## Description
* sets the `desired_capacity_type` for the `aws_autoscaling_group` resource to AWS default value `"units"`

## Motivation and Context
* current implementation succeeds with leaving this attribute undefined which AWS will set to its default `"units"` on their end, but evidently subsequent runs of Terraform don't see this during a refresh of state
  * if a user goes into the AWS Console and edits the Autoscaling Group (changes nothing, just clicks the `Save` button), Terraform will suddenly see this attribute when it refreshes state 🤦 
  * if this happens, you will have no choice but to destroy the autoscaling group and recreate via the Terraform because without this change there is no way to tell the current module to not try to set the `desired_capacity_type = "units" -> null`

    ```
      # module.eks_al2.module.self_managed_node_group["example"].aws_autoscaling_group.this will be updated in-place
      ~ resource "aws_autoscaling_group" "this" {
          - desired_capacity_type            = "units" -> null
            id                               = "example"
            name                             = "example"
            # (26 unchanged attributes hidden)

            # (39 unchanged blocks hidden)
        }

    Plan: 0 to add, 1 to change, 0 to destroy.

    Do you want to perform these actions?
      Terraform will perform the actions described above.
      Only 'yes' will be accepted to approve.

      Enter a value: yes

    module.eks_al2.module.self_managed_node_group["example"].aws_autoscaling_group.this: Modifying... [id=<redacted>]
    ╷
    │ Error: updating Auto Scaling Group (example): InvalidParameter: 1 validation error(s) found.
    │ - minimum field size of 1, UpdateAutoScalingGroupInput.DesiredCapacityType.
    │ 
    ```

## Breaking Changes
* no, `"units"` is the default value

## How Has This Been Tested?
  - I have not as there is nothing to show, this is a value that is being set under the hood


modules/self-managed-node-group/main.tf
```diff
@@ -497,6 +497,7 @@ resource "aws_autoscaling_group" "this" {
   default_cooldown          = var.default_cooldown
   default_instance_warmup   = var.default_instance_warmup
   desired_capacity          = var.desired_size
+  desired_capacity_type     = "units"
```
ftieben, line 500: Please create var.desired_capacity_type and make "units" as a default value

bryantbiggs (approved):

antonbabenko:
This PR is included in version 20.26.0 :tada:

PR #3167, opened 2024-09-27
Add support for al2023 self-managed nodes

mpetrowi:
Add support for al2023 node user data.  This is enabled by adding `platform = al2023` to the worker config.


mpetrowi:
My apologies, I didn't intend to open this against the upstream repo

PR #3171, opened 2024-10-05
fix: The `create_launch_template` variable description says the opposite to what it's actually doing.

fhisamoto:
## Description
Provides a concise description to what the variable is actually doing.

## Motivation and Context
While implementing my own nodegroup `launch_template` I realized the variable description is the opposite to the implementation.

## Breaking Changes
Do no apply

## How Has This Been Tested?


bryantbiggs:
This is not a valid change, it is currently written correctly

PR #3201, opened 2024-11-06
Dev 5769/custom cluster admin arn

ksoviero-liscio:
## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -156,7 +156,7 @@ locals {
   # better controlled by users through Terraform
   bootstrap_cluster_creator_admin_permissions = {
     cluster_creator = {
-      principal_arn = data.aws_iam_session_context.current.issuer_arn
+      principal_arn = var.custom_cluster_creator_admin_arn ? var.custom_cluster_creator_admin_arn : data.aws_iam_session_context.current.issuer_arn
       type          = "STANDARD"
 
       policy_associations = {
--- a/main.tf
+++ b/main.tf
@@ -156,7 +156,7 @@ locals {
   # better controlled by users through Terraform
   bootstrap_cluster_creator_admin_permissions = {
     cluster_creator = {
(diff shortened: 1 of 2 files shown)


main.tf
```diff
@@ -156,7 +156,7 @@ locals {
   # better controlled by users through Terraform
   bootstrap_cluster_creator_admin_permissions = {
     cluster_creator = {
-      principal_arn = data.aws_iam_session_context.current.issuer_arn
+      principal_arn = var.custom_cluster_creator_admin_arn != "" ? var.custom_cluster_creator_admin_arn : data.aws_iam_session_context.current.issuer_arn
```
bryantbiggs, line 159: you shouldn't do this - you should just pass in the intended role via the cluster access entry. this was just a helper utility to make that easier for users to carry over behavior prior to cluster access entries

bryantbiggs: just FYI

ksoviero-liscio: I totally didn't mean to create this PR, but since I did, @bryantbiggs are you saying to disable enable_cluster_creator_admin_permissions and create a aws_eks_access_entry resource independently of the module, or do you mean something else?

bryantbiggs: you can do that, or you can do it through the module like https://github.com/terraform-aws-modules/terraform-aws-eks/blob/d2e6262188aa706c8315ddcd1cc23ae2749edca9/tests/eks-managed-node-group/main.tf#L366-L381

`enable_cluster_creator_admin_permissions` is just mimicking the behavior of the past - for folks who rely on Terraform going into the cluster to provision additional resources, this makes that easier. but if you aren't using the same Terraform IAM entity to go into the cluster, don't use `enable_cluster_creator_admin_permissions` and instead just define those entities and what permissions they should have via `access_entries`

ksoviero-liscio: Oh, I see... That's very helpful, thanks for the help!

PR #3209, opened 2024-11-18
feat: Add variable for disabling dynamic outputs

ajax-ryzhyi-r:
## Description

Add `disable_dynamic_outputs` variable for disabling outputs that can cause drifts

## Motivation and Context

We use this module directly from terragrunt, and occasionally, configuration drift occurs due to changes in the `cluster_platform_version` and `cluster_addons` outputs. Since these changes are made on the AWS side, these drifts are false positives. It would be beneficial to have an option to disable them.

363097684-5a6b5e8c-b734-4a35-b473-221a08c8e44b
Screenshot 2024-11-14 at 08 58 50
Screenshot 2024-11-18 at 09 36 43

## Breaking Changes

All changes are backward compatible
## How Has This Been Tested?


bryantbiggs:
This is not valid - these are computed values and expected to change. Most likely this is a user configuration error

ajax-ryzhyi-r:
> This is not valid - these are computed values and expected to change. Most likely this is a user configuration error

I understand that these values are expected to change, which is why we want to ignore them. As mentioned in the PR description, we're using this module directly from terragrunt, so changes in these outputs cause false-positive drifts.

Any chance of getting this merged?)

bryantbiggs:
> Any chance of getting this merged?)

No - I don't see why this would cause any conflicts. if you don't use these outputs in your implementation, then don't specify them.

ajax-ryzhyi-r:
> > Any chance of getting this merged?)
> 
> No - I don't see why this would cause any conflicts. if you don't use these outputs in your implementation, then don't specify them.

We don't use these outputs but we can't disable them without modifying this module because we're using it directly from terragrunt, which turns this module into the root module

bryantbiggs:
I'm not a terragrunt user but that sounds like a terragrunt issue - @antonbabenko are you familiar with this behavior in terragrunt?

antonbabenko:
I agree that this is not a Terragrunt issue but rather the expected behavior of Terraform (it shows what has changed).

PR #3210, opened 2024-11-19
feat: Added additional conditions into Cluster- and Node IAM Role Assumption Policy

fschykowski:
## Description
Allows modifications to the assume role policy of the eks cluster role and node role. Simply translates a condition object to a condition block, so different combinations for `test`,`variable` and `values` can be passed.

## Motivation and Context
The recommended action by AWS to prevent confused deputy attacks is to add sourceArn, sourceAccount or sourceOrg as conditions as condition to the role assumption. https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html

This has already been provided for sourceArn Conditon for the fargate module: https://github.com/terraform-aws-modules/terraform-aws-eks/commit/a070d7b2bd92866b91e0963a0f819eec9839ed03

I raised a similar PR to VPC module as well which got approved: https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1138

## Breaking Changes
Should not break anything, as the default variable is empty and will simply not render the dynamic block.

## How Has This Been Tested?
- [X] I have validated locally that the config is rendered correctly from terraform plan for both adding sourceAccount as condition and not adding conditions.


bryantbiggs:
thank you, but we will not be accepting this change at this time. if the service docs mention additional configuration that can be set on the trust policy for confused deputy or other guidance, then we will typically support that here. Right now I don't see that in the EKS docs

PR #3214, opened 2024-11-25
feat: Added configurable delay before addons creation

vchepkov:
## Description

Currently, even though the cluster status is active, the addon API is not completely ready yet. This causes retries and significant delays during addon creation, potentially negating the before_compute delay.

## Motivation and Context
when implementing vpc-cni custom networking, addon fails to create in the configured dataplane_wait_duration time interval, causing pods using host network. I opened a case with AWS support and they informed me that there is a 3 minutes delay between cluster marked as ACTIVE and addon VPI to be ready, without delay vpc-cni creation (or any add-on, but vpc-cni is critical) can take up to 20 minutes.

I left default to 0s, for those who do not use custom networking, but in my test suite I set it to 3 minutes and each time addon was created under 20 seconds

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ We are grateful to the community for contributing bugfixes and improvements! Ple
 | [aws_security_group.node](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group) | resource |
 | [aws_security_group_rule.cluster](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule) | resource |
 | [aws_security_group_rule.node](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule) | resource |
+| [time_sleep.addon_delay](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep) | resource |
 | [time_sleep.this](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep) | resource |
 | [aws_caller_identity.current](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity) | data source |
 | [aws_eks_addon_version.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/eks_addon_version) | data source |
@@ -232,6 +233,7 @@ We are grateful to the community for contributing bugfixes and improvements! Ple
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | <a name="input_access_entries"></a> [access\_entries](#input\_access\_entries) | Map of access entries to add to the cluster | `any` | `{}` | no |
+| <a name="input_addon_delay_duration"></a> [addon\_delay\_duration](#input\_addon\_delay\_duration) | Duration to wait after the EKS cluster has become active before creating the addons | `string` | `"0s"` | no |
 | <a name="input_attach_cluster_encryption_policy"></a> [attach\_cluster\_encryption\_policy](#input\_attach\_cluster\_encryption\_policy) | Indicates whether or not to attach an additional policy for the cluster IAM role to utilize the encryption key provided | `bool` | `true` | no |
 | <a name="input_authentication_mode"></a> [authentication\_mode](#input\_authentication\_mode) | The authentication mode for the cluster. Valid values are `CONFIG_MAP`, `API` or `API_AND_CONFIG_MAP` | `string` | `"API_AND_CONFIG_MAP"` | no |
 | <a name="input_bootstrap_self_managed_addons"></a> [bootstrap\_self\_managed\_addons](#input\_bootstrap\_self\_managed\_addons) | Indicates whether or not to bootstrap self-managed addons after the cluster has been created | `bool` | `null` | no |
--- a/main.tf
+++ b/main.tf
@@ -489,6 +489,14 @@ data "aws_eks_addon_version" "this" {
   most_recent        = try(each.value.most_recent, null)
 }
 
+resource "time_sleep" "addon_delay" {
+  count           = length(var.cluster_addons) > 0 ? 1 : 0
+  create_duration = var.addon_delay_duration
+  triggers = {
+    cluster_name = aws_eks_cluster.this[0].name
+  }
+}
+
 resource "aws_eks_addon" "this" {
   # Not supported on outposts
   for_each = { for k, v in var.cluster_addons : k => v if !try(v.before_compute, false) && local.create && !local.create_outposts_local_cluster }
@@ -523,6 +531,7 @@ resource "aws_eks_addon" "this" {
     module.fargate_profile,
     module.eks_managed_node_group,
     module.self_managed_node_group,
+    time_sleep.addon_delay
   ]
 
   tags = merge(var.tags, try(each.value.tags, {}))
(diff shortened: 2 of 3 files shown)


bryantbiggs:
I'm not following - this is at cluster creation or after the cluster has been created that you are facing the issue? 

For cluster creation, simply set:  `before_compute = true` in the VPC CNI addon with the relevant configuration values for custom networking

For existing clusters, you will need to set the appropriate values on the VPC CNI and then cycle the nodes

vchepkov:
It is after cluster creation. several months ago some changes were introduced and addons take up to 20 minutes to create.
I was chasing this problem ever since.

vchepkov:
I attached PR to my support case, hopefully AWS engineers will chime in

bryantbiggs:
> At the cluster creation.

This is how you would do that today 

```hcl
module "eks" {

  # Truncated for brevity
  ...

  cluster_addons = {
     vpc-cni = {
        before_compute       = true
        env = {
           AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG = "true"
           ENI_CONFIG_LABEL_DEF               = "topology.kubernetes.io/zone"
        })
     }
  }

...
}
```

> Several months ago some changes were introduced and addons take up to 20 minutes to create.
> I was chasing this problem ever since.

Can you share a reproduction?

antonbabenko:
This issue has been resolved in version 20.30.0 :tada:

bryantbiggs:
@vchepkov would you mind updating to the latest, `v20.30.0` that was just released, and reporting back on if that resolves the issues you are seeing?

bryantbiggs:
also, do check out the test case I added which shows configurations that should be used to help improve this situation. In the next breaking change of the module, these will be baked in as the defaults, but for now users can set them to reach the desired outcome:

- https://github.com/terraform-aws-modules/terraform-aws-eks/blob/b468ead1d97d3c9a83fad227884b3cc689f2b200/tests/fast-addons/main.tf#L30-L31
- https://github.com/terraform-aws-modules/terraform-aws-eks/blob/b468ead1d97d3c9a83fad227884b3cc689f2b200/tests/fast-addons/main.tf#L50-L55
- https://github.com/terraform-aws-modules/terraform-aws-eks/blob/b468ead1d97d3c9a83fad227884b3cc689f2b200/tests/fast-addons/main.tf#L139-L140

vchepkov:
Definitely, will test. But I am a bit skeptical, because in my tests it matches what support folks said, you need a 3 minute delay and timing in your MRs is way different. Maybe combination with disabling bootstrap addons? I will report back shortly

bryantbiggs:
I have the internal ticket between support and the service team 😉 and am quite optimistic that this will help the situation

vchepkov:
oh, I don't need ENIConfig anymore? that is nice and 🤞

vchepkov:
@bryantbiggs , I tried new v20.30.1 on Wednesday and it didn't work, so, but I haven't implemented subnet discovery yet
So today I added it and result is the same - none of the addons are being created - so they will timeout in 20 minutes :
```
module.eks.module.eks-cluster.aws_eks_addon.this["coredns"]: Still creating... [3m20s elapsed]
module.eks.module.eks-cluster.aws_eks_addon.before_compute["vpc-cni"]: Still creating... [4m30s elapsed]
module.eks.module.eks-cluster.aws_eks_addon.this["kube-proxy"]: Still creating... [3m30s elapsed]
```

From what I can tell, the only difference in our code - we use self_managed_node_groups. Is it possible that disabling bootstrap_self_managed_addons require   eks managed nodes?

bryantbiggs:
Can you show more of the apply sequence of events

vchepkov:
[clusters-output.txt](https://github.com/user-attachments/files/17960005/clusters-output.txt)

vchepkov:
Removed gitlab's coloring

[clusters-output-cleaned.txt](https://github.com/user-attachments/files/17960060/clusters-output-cleaned.txt)

bryantbiggs:
thank you for sharing, still digging through it but wanted to comment and better understand a few things:
1. I don't believe it would have any influence on the issue you are experiencing, but why the use of self-managed node groups instead of EKS managed node groups?
2. We do support EKS Pod Identity on EKS addons today, and we have a module for this to make that a bit easier https://github.com/terraform-aws-modules/terraform-aws-eks-pod-identity instead of using IRSA. (Note: VPC CNI IRSA doesn't make a lot of sense unless you are removing the permissions needed by VPC CNI after the node group has been created and nodes join the cluster. You can create clusters without the VPC CNI IRSA role - they inherit the permissions of the node - but you can't create clusters without the permissions on the node IAM role when using VPC CNI w/ IRSA) 
3. Any reason for continuing to use `aws-auth` ConfigMap instead of switching to cluster access entries? This helps avoid chaining providers due to Terraform needing to authenticate and go into the cluster to create/update the configmap. Right now it looks like you are adding the same roles in both the ConfigMap and as access entries
4. Are you defining your node groups within the EKS module itself, like [this](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/6866b40becaddeca60638096da919732b8b52902/tests/self-managed-node-group/main.tf#L81) - or are you defining them as standalone, external modules, like [this](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/6866b40becaddeca60638096da919732b8b52902/tests/eks-managed-node-group/main.tf#L425-L452)? 
5. Did you get a chance to look at the logs for the VPC CNI and CoreDNS on why they aren't reaching a  healthy state?

vchepkov:
It's my please, thank you for looking into it. I can upload all my code to support ticket if you like, if that helps
1. There are some security tools that we need to install on host (security wants them)
2. we have been using that irsa module for quite awhile, I think it was just following principle of least privileges. If that's causing issue we can refactor it
3. some other apps we use still rely on classic config-map, so we enable it. Technically, nothing relies on it in this module and I can move it elsewhere, but I don't think it's causing an issue.
4. we using the same module and set self_managed_node_group_defaults  and self_managed_node_groups

bryantbiggs:
hi @vchepkov - can you please share a minimum reproduction of this issue. Here is my minimum reproduction that demonstrates the functionality is working as intended https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/tests/fast-addons

vchepkov:
I will try to use your example with our roles and deployment pipeline, I am at loss now. 
I have stripped out irsa and aws-auth and it didn't make any difference

vchepkov:
@bryantbiggs , I wasn't able to exactly reproduce the issue yet, but, two changes makes addons behave unexpectedly.
```
$ git diff
diff --git a/tests/fast-addons/main.tf b/tests/fast-addons/main.tf
index 12d2a41..cd1242b 100644
--- a/tests/fast-addons/main.tf
+++ b/tests/fast-addons/main.tf
@@ -3,9 +3,10 @@ provider "aws" {
 }
 
 locals {
-  name            = "ex-${basename(path.cwd)}"
+  name            = "vchepkov-${basename(path.cwd)}"
   cluster_version = "1.31"
-  region          = "eu-west-1"
+  region          = "us-east-1"
+  cni_version     = "v1.19.0-eksbuild.1"
 
   tags = {
     Test       = local.name
@@ -43,7 +44,7 @@ module "eks" {
       most_recent = true
     }
     vpc-cni = {
-      most_recent    = true
+      addon_version  = local.cni_version
       before_compute = true
         env = {
@@ -61,7 +62,7 @@ module "eks" {
 
-  eks_managed_node_groups = {
+  self_managed_node_groups = {
     example = {
       instance_types = ["m6i.large"]
```

As I mentioned before, we do use self_managed_node_groups and we pin addon_version, because we had an outage in the past with an addon upgrade. These two changes make coredns to consistently take 15 minutes to create.

bryantbiggs:
sorry, which two changes make the addons behave differently?

vchepkov:
I posted a diff in the comment? 
```
-      most_recent    = true
+      addon_version  = local.cni_version
...
-  eks_managed_node_groups = {
+  self_managed_node_groups = {
```

bryantbiggs:
Ah ok those two - wasn't sure about the changes to region, etc

vchepkov:
Ah sorry, yes, those are not important

PR #3233, opened 2024-12-08
chore: Add `TODO` to turn off detailed monitoring by default at next breaking change

aiell0:
## Description
Turn off detailed monitoring of EKS managed node groups by default.

## Motivation and Context
Detailed monitoring gets expensive, and users who are new to the module may not know this setting exists. If someone discovers they want this setting, they could turn it on.

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/variables.tf
+++ b/modules/eks-managed-node-group/variables.tf
@@ -273,10 +273,11 @@ variable "metadata_options" {
   }
 }
 
+# TODO - make this false by default at next breaking change
 variable "enable_monitoring" {
   description = "Enables/disables detailed monitoring"
   type        = bool
-  default     = false
+  default     = true
 }
 
 variable "enable_efa_support" {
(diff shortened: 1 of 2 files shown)


modules/eks-managed-node-group/variables.tf
```diff
@@ -276,7 +276,7 @@ variable "metadata_options" {
 variable "enable_monitoring" {
   description = "Enables/disables detailed monitoring"
   type        = bool
-  default     = true
+  default     = false
```
bryantbiggs, line 279: we can't change this since that would be a breaking change, but we can add a note to change the default at the next breaking change. For example https://github.com/terraform-aws-modules/terraform-aws-eks/blob/c63a57908d7b4d51895d3f8502d60daae758b761/variables.tf#L496

aiell0: Great just made the change, thanks!

bryantbiggs (approved):

antonbabenko:
This PR is included in version 20.31.1 :tada:

PR #3242, opened 2024-12-13
fix: Auto Mode custom tag policy should apply to cluster role, not node role

bryantbiggs:
## Description
- Auto Mode custom tag policy should apply to cluster role, not node role

## Motivation and Context
- Currently this is incorrect per https://docs.aws.amazon.com/eks/latest/userguide/auto-learn-iam.html#tag-prop and therefore instances fail to provision when custom tags are applied to a node class

## Breaking Changes
- Claiming this as a no since it never worked as intended. Yes, the policy will be destroyed and permissions removed from the node but those permissions were not being used for anything - they need to be applied to the cluster role

## How Has This Been Tested?


antonbabenko (approved):

antonbabenko:
This PR is included in version 20.31.4 :tada:

PR #3243, opened 2024-12-17
feat: Add an AMI version output to modules/eks-managed-node-group

jkt-signal:
## Description

Adds an output to the `eks-managed-node-group` module with the release version used for the node group. 

## Motivation and Context

The release version being used (when not supplied from the outside with the `ami_release_version` or `ami_id` variables) is only available in a resource and locals private to this module. Users of the module can therefore not determine what AMI release version is being used if they use `use_latest_ami_release_version`; this change adds an output with that information, which makes it possible for users of the module to determine whether an update is happening (by comparing to an `aws_eks_managed_node_group` data resource, for example).

## Breaking Changes

No, this only adds an output and makes no breaking changes.

## How Has This Been Tested?

The existing `examples/eks-managed-node-group` applies with no changes in the plan after this commit. The examples do not seem to include examples of _using_ the outputs of the module, and I did not add one.

### DIFF
--- a/modules/eks-managed-node-group/outputs.tf
+++ b/modules/eks-managed-node-group/outputs.tf
@@ -36,6 +36,11 @@ output "node_group_id" {
   value       = try(aws_eks_node_group.this[0].id, null)
 }
 
+output "node_group_release_version" {
+  description = "AMI version of the EKS node group"
+  value       = try(aws_eks_node_group.this[0].release_version, null)
+}
+
 output "node_group_resources" {
   description = "List of objects containing information about underlying resources"
   value       = try(aws_eks_node_group.this[0].resources, null)


bryantbiggs:
I don't see why this is useful, could you elaborate? if you want to know the release version used by a node group, you can look in the console or use the API https://docs.aws.amazon.com/cli/latest/reference/eks/describe-nodegroup.html

jkt-signal:
> I don't see why this is useful, could you elaborate? if you want to know the release version used by a node group, you can look in the console or use the API https://docs.aws.amazon.com/cli/latest/reference/eks/describe-nodegroup.html

This isn't for looking at the version manually; I'd like to use it to set up a Terraform precondition block to require manual confirmation if a `terraform apply` is about to upgrade the Kubernetes version on our managed node groups.

bryantbiggs:
that still doesn't make sense. you can set the Kubernetes version that should be used by the node group and control when and to what version it is upgraded

jkt-signal:
Sorry, that was a typo; we do fix the Kubernetes version, I meant the AMI version.

bryantbiggs:
still not following - we support tracking latest, otherwise you can specify the AMI version if you'd like. nothing to do with outputs

Issue #3247, opened 2024-12-19
create eks with authentication_mode = "CONFIG_MAP"


consooo:
Hi Team,

this is a duplicate of https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3206 because we encountered the same issue.

`authenticationMode` has to be set to `API_AND_CONFIG_MAP` for creating new clusters. So I think option `CONFIG_MAP` should not be an allowed value due to the hardcoding of `BootstrapClusterCreatorAdminPermissions: false`.

we now set additionally `enable_cluster_creator_admin_permissions: true` as well. Is this configuration similar to the previous `CONFIG_MAP` setting? There should be some mention of this in the upgrade guide: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/UPGRADE-20.0.md

bryantbiggs:
sorry, what exactly are you asking for?

sbets:
@bryantbiggs In case of hardcoded BootstrapClusterCreatorAdminPermissions: false no backwards compatibility for existing clusters. Now it is forcing me to recreate my existing cluster. It was already fixed on aws provider level here https://github.com/hashicorp/terraform-provider-aws/pull/38295, so it would be nice to have the same here.

Or may be you can help here and share some approach to upgrade existing cluster without recreation?

Thanks!

bryantbiggs:
You're commenting on an issue that's not related to that, and you haven't provided any context/reproduction so it's nearly impossible to offer any guidance

consooo:
@bryantbiggs sorry for the late reply. Let me try to clarify. For context, we have eks clusters with `authentication_mode =  "CONFIG_MAP"`. This was possible when using a version prior `v20.29.0`. Nowhere in the documentation is mentioned that this authentication mode is no longer supported and "CONFIG_MAP" is still a valid value although it shouldn't be
-> [variables.tf](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/a713f6f464eb579a39918f60f130a5fbb77a6b30/variables.tf#L41C11-L41C30)

For upgrading clusters, it's still possible to keep this value but it isn't for creating new clusters.

So my question is: what is the recommended change here? Do I have to switch from "CONFIG_MAP" to "API_AND_CONFIG_MAP"? What does this change? There is no documentation as, imo, this change shouldn't happen in a minor version.

bryantbiggs:
[...]

PR #3270, opened 2025-01-10
fix: Use variable to enable EKS cluster creation with CONFIG_MAP authentication mode

hsalluri259:
## Description
Using `enable_cluster_creator_admin_permissions` variable for `bootstrap_cluster_creator_admin_permissions` in place of hard coded `false` value to support EKS cluster creation when authentication mode is `CONFIG_MAP` only. I think alternatively we can use API_AND_CONFIG_MAP, but this will take time for us to migrate our clusters and customers clusters. 
## Motivation and Context
It allows a user to create an EKS cluster when authentication mode is `CONFIG_MAP` only by using `bootstrap_cluster_creator_admin_permissions=true` . 
Without this value, the user will be facing the below error.
```
Error: creating EKS Cluster (k8s-dev-eks-xyz: operation error EKS: CreateCluster, https response error StatusCode: 400, RequestID: eec24135-0159-484f-b0d7-b5c5518b3cd5, InvalidParameterException: bootstrapClusterCreatorAdminPermissions must be true if cluster authentication mode is set to CONFIG_MAP
```
It fixes the following open issue. 
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3247 
## Breaking Changes
The default value for `bootstrap_cluster_creator_admin_permissions` variable is still `false`. So it doesn't break anything. 

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -44,12 +44,8 @@ resource "aws_eks_cluster" "this" {
   access_config {
     authentication_mode = var.authentication_mode
 
-    # See access entries below - this is a one time operation from the EKS API.
-    # Instead, we are hardcoding this to false and if users wish to achieve this
-    # same functionality, we will do that through an access entry which can be
-    # enabled or disabled at any time of their choosing using the variable
-    # var.enable_cluster_creator_admin_permissions
-    bootstrap_cluster_creator_admin_permissions = false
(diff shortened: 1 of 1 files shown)


bryantbiggs:
Thank you, however this is not a valid change and we have no plans to modify this value

PR #3276, opened 2025-01-13
fix: Added Name Tag for EKS Nodes if using NOT Custom Launch Template

adv4000:
Added Name tag to EKS Nodes if using EKS Native Launch Template otherwise no tag Name exist

## Description
Added Name tag to EKS Nodes if using EKS Native Launch Template otherwise no tag Name exist

## Motivation and Context
If  `var.use_custom_launch_templat=false`  EKS nodes doesn't have any Name Tags
In AWS Console nodes seen with empty Name.

## Breaking Changes
NO

## How Has This Been Tested?

image


bryantbiggs:
this is one of the primary reasons you should use the custom launch template the module provides (default) - EKS MNG doesn't support this without it

PR #3280, opened 2025-01-20
fix : Removed EC2 type from exception condition so policy attachment for access entry can be created

vibhamsharma:
…y can be done for EC2 type for creating access entry for Node role.

While Creating Access entry for Self managed node role, we were getting below error as it seems EC2 type is included in exception for policy attachment. Once I removed EC2 from exception condition, it started picking up Policy attachment and Access Entry got created.

│ Error: Unsupported attribute
│ 
│   on .terraform/modules/eks/main.tf line 289, in resource "aws_eks_access_policy_association" "this":
│  289:   policy_arn    = each.value.association_policy_arn
│     ├────────────────
│     │ each.value is object with 3 attributes
│ 
│ This object does not have an attribute named "association_policy_arn".

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -258,7 +258,7 @@ locals {
           association_policy_arn              = pol_val.policy_arn
           association_access_scope_type       = pol_val.access_scope.type
           association_access_scope_namespaces = lookup(pol_val.access_scope, "namespaces", [])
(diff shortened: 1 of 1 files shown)


vibhamsharma (commented):
fix : Removed EC2 type from exception condition so policy attachment for access entry can be created.

Issue #3273, opened 2025-01-10
Disable EKS auto mode fails


project0:
## Description

Disabling eks auto mode on an existing cluster fails. Apparently there are some related changes  and issues that claimed to solve and reverted it again. As all related issues are closed i am have decided to raise a new issues. 

* https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3268
* https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3253
 * https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3252
 * https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3236

## Versions

- Module version [Required]:

- Terraform version:
Terraform v1.10.2

- Provider version(s):
```
14:44:13.210 STDOUT terraform: Providers required by configuration:
14:44:13.210 STDOUT terraform: .
14:44:13.210 STDOUT terraform: ├── provider[registry.terraform.io/hashicorp/aws] >= 5.9.0
14:44:13.210 STDOUT terraform: ├── module.irsa_external-dns
14:44:13.210 STDOUT terraform: │   └── provider[registry.terraform.io/hashicorp/aws] >= 4.0.0
14:44:13.210 STDOUT terraform: ├── module.karpenter
14:44:13.210 STDOUT terraform: │   └── provider[registry.terraform.io/hashicorp/aws] >= 5.81.0
14:44:13.210 STDOUT terraform: ├── module.vpc
14:44:13.210 STDOUT terraform: │   └── provider[registry.terraform.io/hashicorp/aws] >= 5.46.0
14:44:13.210 STDOUT terraform: ├── module.eks
14:44:13.210 STDOUT terraform: │   ├── provider[registry.terraform.io/hashicorp/aws] >= 5.81.0
14:44:13.210 STDOUT terraform: │   ├── provider[registry.terraform.io/hashicorp/tls] >= 3.0.0
14:44:13.210 STDOUT terraform: │   ├── provider[registry.terraform.io/hashicorp/time] >= 0.9.0
14:44:13.210 STDOUT terraform: │   ├── module.eks_managed_node_group
14:44:13.210 STDOUT terraform: │       ├── provider[registry.terraform.io/hashicorp/aws] >= 5.81.0
14:44:13.210 STDOUT terraform: │       └── module.user_data
[...]

PR #3308, opened 2025-02-22
fix: Supports setting `cluster_compute_config = null` or `cluster_compute_config.enabled = false`

### LINKED ISSUE #3307: Error if `null` value assigned to `cluster_config_mode`, or if `cluster_config_mode.enabled = false`
## Description

There is one logic error around the `cluster_config_mode`, and another error around the API interaction.

The logic error is that if a user assigns a `null` value to `cluster_config_mode`, then the `length()` function throws an error, here:

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/1bfc10a5589afadf9c633e0417dbf67316e3e32b/main.tf#L56

```
╷
│ Error: Invalid function argument
│
│   on .terraform/modules/eks.cluster/main.tf line 56, in resource "aws_eks_cluster" "this":
│   56:     for_each = length(var.cluster_compute_config) > 0 ? [var.cluster_compute_config] : []
│     ├────────────────
│     │ while calling length(value)
│     │ var.cluster_compute_config is null
│
│ Invalid value for "value" parameter: argument must not be null.
╵
```

Since the default value is already an empty map, that error is most easily fixed by adding `nullable = false` to the variable definition.

---

The second error, around the API interaction, is that if you _only_ set `cluster_config_mode.enabled = false`, then the cluster will plan fine, but then fail to create with an error reported by the API during the apply:

```
│ Error: creating EKS Cluster (test-minimum-inputs-kuneic): operation error EKS: CreateCluster, https response error StatusCode: 400, RequestID: 573c664d-68c6-463c-b117-ba12722ebe96, InvalidParameterException: For EKS Auto Mode, please ensure that all required configs, including computeConfig, kubernetesNetworkConfig, and blockStorage are all either fully enabled or fully disabled.
```

In one sense, I think the API is doing the wrong thing, since we're _disabling_ auto mode and the module already empties all the other mentioned configs when disabling the compute config. But probably the easiest way to resolve the problem is to change the condition on the dynamic block to just be `local.auto_mode_enabled`, instead of testing the `length()` of the input...

Note that this was previously reported, but the issue was closed without resolution due to lack of details, https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3268. Hopefully I've provided enough to pursue/accept a fix.

I'll have a PR ready shortly to resolve this report...

## Versions

- Module version [Required]: 20.33.1

- Terraform version: 1.10.5

- Provider version(s): 

```
+ provider registry.terraform.io/hashicorp/aws v5.87.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.5
+ provider registry.terraform.io/hashicorp/helm v2.17.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.35.1
+ provider registry.terraform.io/hashicorp/null v3.2.3
+ provider registry.terraform.io/hashicorp/random v3.6.3
+ provider registry.terraform.io/hashicorp/time v0.12.1
+ provider registry.terraform.io/hashicorp/tls v4.0.6
```

```
  cluster_compute_config = null
```

```
  cluster_compute_config = {
    enabled = false
  }
```

## Expected behavior

1. Expected to be able to pass a `null` value to `cluster_compute_config` to get the default value.

2. Expected to be able to disable Auto Mode by setting `cluster_compute_config.enabled = false`.

## Actual behavior

Both scenarios fail, for reasons detailed above.

lorengordon:
## Description

Couple logic issues around `cluster_compute_config` that this is addressing:

1. The logic around enabling Auto Mode is based on the value never being `null`, otherwise the `length()` function fails, so we use `nullable = false` in the variable definition so that a null value will be interpreted as an empty map (the default value).
2. The API will return an error if `compute_config.enabled` is set to `false`, when every other required option for auto mode is also not provided. So we just condition the block on `local.auto_mode_enabled` so the entire block is absent.

## Motivation and Context

Fixes #3307

## Breaking Changes

None

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -53,7 +53,7 @@ resource "aws_eks_cluster" "this" {
   }
 
   dynamic "compute_config" {
-    for_each = length(var.cluster_compute_config) > 0 ? [var.cluster_compute_config] : []
+    for_each = local.auto_mode_enabled ? [var.cluster_compute_config] : []
 
     content {
       enabled       = local.auto_mode_enabled
--- a/variables.tf
+++ b/variables.tf
@@ -47,6 +47,7 @@ variable "authentication_mode" {
 variable "cluster_compute_config" {
   description = "Configuration block for the cluster compute configuration"
   type        = any
+  nullable    = false
   default     = {}
 }
 


bryantbiggs:
This doesn't solve it, has to be fixed upstream first https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3273

lorengordon:
The issue you linked to is about an existing cluster. These errors occur in a brand new cluster. Please reopen and reconsider a little further.

lorengordon:
Alright, I've figured out the issue with disabling Auto Mode. It's not really an upstream issue, as much as a difficult API. But this module *can* support disabling Auto Mode, with some limitations. I'm pushing the fix to this branch, the commit has the full explanation. If you'll please reopen and review, it would be much appreciated!

bryantbiggs:
we are not making any further changes to enable/disable of Auto Mode until https://github.com/hashicorp/terraform-provider-aws/pull/41155 lands upstream

lorengordon:
I don't think that upstream patch is actually a solution to the issue. It can't fix the CreateCluster API, and *that* is what is imposing the limitations. It's not a provider issue.

lorengordon:
Well I guess the PR doesn't update once closed, but the branch has the fix. You can review the changeset here, instead, https://github.com/terraform-aws-modules/terraform-aws-eks/compare/master...lorengordon:terraform-aws-eks:fix/compute-config-false-or-null

I'll leave the branch, so It'll be waiting for you whenever you get around to reviewing.

PR #3309, opened 2025-02-24
fix: Provides accurate type constraint for cluster_compute_config

lorengordon:
## Description

Adds type information to cluster_compute_config inpute. This aligns with Terraform's recommendation around type constraints, [stating the `any` type is "incorrect" as currently applied in the module](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#dynamic-types-the-any-constraint).

## Motivation and Context

Relates to https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3307

## Breaking Changes

None

## How Has This Been Tested?


god-of-biscuits:
This change would really solve an edge case that causes a 400 error and makes it near impossible to support auto mode and non auto mode clusters with the same module.

PR #3356, opened 2025-05-08
fix: Fallback to clusters primary sg during node sg rule creation

visit1985:
## Description

This PR fixes a situation, where Terraform tries to create node security group rules with an empty string as `source_security_group_id` when the module is used with input variable `create_cluster_security_group = false`.

## Motivation and Context

I'm migrating existing clusters to this module which do not use a secondary cluster security group.

## Breaking Changes

**None**

Although I had to remove `depends_on { aws_security_group_rule.node }` from `aws_eks_cluster.this` to prevent a circular dependency all node groups and addons still depend on it via attributes. So the order of applied resources is still intact.

## How Has This Been Tested?

  Tested with `examples/eks-auto-mode` and `examples/eks-managed-node-group` by adding `create_cluster_security_group = false`, running apply, re-apply (without changes) and destroy.


visit1985:
Hey @bryantbiggs, is there any thing I can help with, to speed up the review process for this PR?

Thanks

bryantbiggs:
this isn't a valid change - please start with an issue and a reproduction so we can take a look

PR #3376, opened 2025-06-08
fix: Ensure `additional_cluster_dns_ips` is passed through from root module

880831ian:
## Description

Add missing `additional_cluster_dns_ips` variable definition and forwarding in the main module file `node_group.tf`.

## Motivation and Context

The `additional_cluster_dns_ips` variable was not previously declared or passed in the main module (`node_group.tf`). This caused issues when attempting to use `self.additional_cluster_dns_ips` as an input to downstream modules such as `self-managed-node-group` and `_user_data`.

By explicitly defining and forwarding the variable, this fix enables proper DNS IP customization for NodeLocal DNS and similar use cases.

## Breaking Changes

No breaking changes. This PR only adds support for an optional input variable.

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -479,17 +479,18 @@ module "self_managed_node_group" {
   # User data
   platform = try(each.value.platform, var.self_managed_node_group_defaults.platform, null)
   # TODO - update this when `var.platform` is removed in v21.0
-  ami_type                 = try(each.value.ami_type, var.self_managed_node_group_defaults.ami_type, "AL2_x86_64")
-  cluster_endpoint         = try(time_sleep.this[0].triggers["cluster_endpoint"], "")
-  cluster_auth_base64      = try(time_sleep.this[0].triggers["cluster_certificate_authority_data"], "")
-  cluster_service_cidr     = try(time_sleep.this[0].triggers["cluster_service_cidr"], "")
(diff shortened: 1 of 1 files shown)


880831ian:
Hi, just a note for reviewers:

I’ve tested this change using the following example:
[examples/self-managed-node-group/eks-bottlerocket.tf](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/self-managed-node-group/eks-bottlerocket.tf)

Without explicitly adding the additional_cluster_dns_ips variable in the root node_group.tf, it’s not possible to pass this value correctly to the self-managed-node-group and _user_data modules. This causes the additional_cluster_dns_ips setting to be silently ignored.

This patch ensures that the value can be propagated properly through the module chain.

Thanks!

bryantbiggs:
what OS/AMI are you using?

880831ian:
> what OS/AMI are you using?

I’m using the BOTTLEROCKET_x86_64 AMI type in the node group configuration, as defined in the example:
[examples/self-managed-node-group/eks-bottlerocket.tf](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/self-managed-node-group/eks-bottlerocket.tf)

antonbabenko:
This PR is included in version 20.36.1 :tada:

Issue #3371, opened 2025-06-03
allow multiply roles to be passed to AllowPassingInstanceRole


ipleten:
### Description

I want to add Windows nodes to cluster with latest module version. 

I'm trying to add windows node to EKS cluster this requires to create new Access entity with new IAM role and EC2_WINDOWS type. I can't reuse the same role as access entity requites to have unique role name  (we don't use aws-auth configmap). and it raises error:
`     ResourceInUseException: The specified access entry resource is already in use on this cluster.`

If I create a new role there is no way to pass it into Karpenter module to be used  with `AllowPassingInstanceRole`.

I saw related issues but they all about using aws-auth config. It doesn't work with access entities or I missing something.  

Update: I believe i can use Additional policy to do this.

PR #3379, opened 2025-06-10
feat: Allow Karpenter access to IAM:PassRole more than one node role arn

toneill818:
## Description
Option to pass extra ARNs that can be used as instance profiles to Karpenter's policy.
Fixes Issue #3371

## Motivation and Context
When creating a cluster with a mix of Windows and Linux nodes and using Access Entries we need two IAM Roles. One for the Windows instances that has an access entry of type EC2_WINDOWS and one for the Linux instances with an access entry of type EC2_LINUX.

## How Has This Been Tested?

- [ X ] I have executed `pre-commit run -a` on my pull request


marcusramberg:
Not sure how you can use the 21 release with both windows and linux nodes without this change as an access entry for a role can't be both EC2_LINUX and EC2_WINDOWS. Could it be reopened please?

PR #3409, opened 2025-07-07
Add a comment to scaling_config desired_size for future readers

marcus-crane:
## Description

Add a code comment to the EKS managed node group scaling_config lifecycle policy hinting at why it exists

## Motivation and Context

Earlier this morning, I was caught out by the `desired_size` lifecycle issue. I had a skim through some issues, didn't see any immediately obvious issue titles and almost opened an issue despite thinking that surely the lifecycle ignore had to be intentional.

As a hint for future readers, I think it would be useful to leave a hint for anyone who runs into this caveat. Whether my wording is what you're after I'm happy to discuss but I think some sort of hint should be left there 🙂 

## Breaking Changes

None

## How Has This Been Tested?

N/A

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -476,6 +476,8 @@ resource "aws_eks_node_group" "this" {
   lifecycle {
     create_before_destroy = true
     ignore_changes = [
+      // If you are here, you may think this is strange or even a bug but it is intentional!
+      // For more context, check out https://github.com/bryantbiggs/eks-desired-size-hack
       scaling_config[0].desired_size,
     ]
   }


bryantbiggs:
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md

bryantbiggs:
https://github.com/bryantbiggs/eks-desired-size-hack

marcus-crane:
> https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/faq.md#why-are-there-no-changes-when-a-node-groups-desired_size-is-modified

This would probably a better thing to link.

I understand why the lifecycle exception exists but for users who have this module abstracted away, one discovery path is setting desired at a higher level, receiving an AWS error and then tracing the module reference down into the provider code, rather than setting up the module for the first time where you would be more likely to read over those FAQs.

I do think having a hint in the code can eliminate some number of redundant queries you likely get as a maintainer.

PR #3438, opened 2025-07-24
fix: Local enable_encryption_config still reads true even when toggled off

TheFLHurricane:
Because encryption_config is now strictly defined as an object, count will never be <1 even if empty.

## Description
Addendum to previous PR... Changes enable_encryption_config to react to actual value of encryption_config rather than count since strict type object will always be >0.

## Motivation and Context

This allows enable_encryption_config to become false when encryption_config is {} as in previous versions.

## Breaking Changes

No, restores backwards compatibility.

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -24,7 +24,7 @@ locals {
   role_arn = try(aws_iam_role.this[0].arn, var.iam_role_arn)
 
   create_outposts_local_cluster = var.outpost_config != null
-  enable_encryption_config      = length(var.encryption_config) > 0 && !local.create_outposts_local_cluster
+  enable_encryption_config      = var.encryption_config != {} && !local.create_outposts_local_cluster
 
   auto_mode_enabled = try(var.compute_config.enabled, false)
 }


main.tf
```diff
@@ -24,7 +24,7 @@ locals {
   role_arn = try(aws_iam_role.this[0].arn, var.iam_role_arn)
 
   create_outposts_local_cluster = var.outpost_config != null
-  enable_encryption_config      = length(var.encryption_config) > 0 && !local.create_outposts_local_cluster
```
bryantbiggs, line 27: do you have a reproduction of what you are trying to do? I'm not quite following, and this logic change looks incorrect

main.tf
```diff
@@ -24,7 +24,7 @@ locals {
...
+  enable_encryption_config      = var.encryption_config != {} && !local.create_outposts_local_cluster
```
gws, line 27: Something like this works for me, there's probably a better way to do it

```suggestion
  enable_encryption_config      = length(coalesce(var.encryption_config.resources, [])) > 0 && !local.create_outposts_local_cluster
```

antonbabenko:
This issue has been resolved in version 21.0.4 :tada:

PR #3436, opened 2025-07-24
fix: Move `encryption_config` default for `resources` out of type definition and to default variable value to allow disabling encryption

TheFLHurricane:
Fixes inability to provide default {} encryption_config

## Description

Updates how encryption_config default is defined to allow for empty {} override to turn off encryption.

## Motivation and Context
- Resolves #3435

## Breaking Changes

No, fixes a break

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

Issue #3435, opened 2025-07-24
v21: encryption cannot be turned off with encryption config.


TheFLHurricane:
## Description

encryption_config variable cannot be set to {} to disable encryption.  Although {} is the default, the type setting for resources provides a default of ["strings"] which will always override the empty {} even if explicitly set.  You can change this by declaring resources = [], however, because the local.enable_encryption_config relies on length(var.encryption_config) > 0, it still considers encryption to be turned on and creates all associated KMS resources.

## Versions

- Module version [Required]: v21.0

- Terraform version: v1.12.2
 
- Provider version(s): v6.4.0

## Reproduction Code [Required]

module "eks" {
  create = var.create_eks_cluster

  source  = "terraform-aws-modules/eks/aws"

  kubernetes_version = var.eks_cluster_version
  name               = "${var.cluster}-${var.release_name}-cluster"
  vpc_id             = var.vpc_id
  subnet_ids         = flatten(var.subnets)

  enable_irsa = true

  endpoint_private_access = true
  endpoint_public_access  = false

  create_security_group              = false
  create_primary_security_group_tags = false
  security_group_id                  = data.aws_security_group.eks_cluster_sg.id
  create_node_security_group         = false
  node_security_group_id             = data.aws_security_group.eks_all_workers_sg.id

  prefix_separator             = ""
  iam_role_name                = "${var.cluster}-${var.release_name}-cluster"
  iam_role_additional_policies = {
    AmazonEKSServicePolicy = "arn:aws:iam::aws:policy/AmazonEKSServicePolicy"
  }

  addons = {
    kube-proxy = {
      resolve_conflicts_on_create = "OVERWRITE"
    }
    eks-node-monitoring-agent = {}
  }

  encryption_config = {}
  
  timeouts = {
    create = "30m"
    delete = "15m"
    update = "60m"
  }

  enable_cluster_creator_admin_permissions = true

}

Set encryption_config to {}

## Expected behavior

Not create any resources related to encryption

## Actual behavior

All encryption resources are created and encryption added for secrets.

## Additional context

Solution is to remove ["secrets"] as default for resources under type and put resources = ["secrets"] into the default for the variable.  This will maintain the existing default functionality while allowing users to still optionally flag off the encryption as before.

TheFLHurricane:
Provided a fix if accepted - https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3436

antonbabenko:
[...]

PR #3439, opened 2025-07-24
fix: Correct encryption configuration enable logic; avoid creating Auto Mode policy when Auto Mode is not enabled

bryantbiggs:
## Description
- Correct encryption configuration enable logic to allow users to disable CMK encryption
- Avoid creating Auto Mode policy when Auto Mode is not enabled
- 
## Motivation and Context
- Closes #3438
- Ref #3436
- Ref #3435

## Breaking Changes
- No

## How Has This Been Tested?


antonbabenko:
This PR is included in version 21.0.4 :tada:

tluck:
something is not right here with enable_encryption_config
 
- if one wants to use an aws owned kms key and not a customer managed key...

```
locals {
  create = var.create && var.putin_khuylo

  partition  = try(data.aws_partition.current[0].partition, "")

  role_arn = try(aws_iam_role.this[0].arn, var.iam_role_arn)

  create_outposts_local_cluster = var.outpost_config != null
  enable_encryption_config      = var.encryption_config != null && !local.create_outposts_local_cluster

  auto_mode_enabled = try(var.compute_config.enabled, false)
}

```
with this line
    enable_encryption_config      = var.encryption_config != null && !local.create_outposts_local_cluster

enable_encryption seems to become **true** by default - since the default encryption_config (in variables.tf) is an empty object and not null

edit/setting default encryption_config = null vs {}  in module variables .tf (like outpost_config) - then it works as expected (and like v5.)

bryantbiggs:
We're not triaging unless there is a reproduction

PR #3456, opened 2025-07-30
fix: Add enable_data_source_optimizations variable

### LINKED ISSUE #3440: v21: adding depends_on to module breaks eks-managed-node-group
## Description

When deploying the module in a fresh workspace and adding a depends_on field to the module a plan or apply fails with `Error: Invalid count argument`

## Versions

- Module version [Required]: ~> 21.0

- Terraform version: v1.11.3

- Provider version(s): aws v6.5.0

```
# Edit examples/eks-managed-node-groups/eks-al2023.tf
module "eks_al2023" {
  depends_on = [module.vpc] # add this field
  ... # leave other fields unchanged
}

```

- try to apply examples/eks-managed-node-groups on a fresh workspace after the edit above

## Expected behavior
`terraform plan` should go through

## Actual behavior

```
╷
│ Error: Invalid count argument
│ 
│   on .terraform/modules/eks_al2023/modules/eks-managed-node-group/main.tf line 2, in data "aws_partition" "current":
│    2:   count = var.create && var.partition == "" ? 1 : 0
│ 
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the
│ resources that the count depends on.
╵
╷
│ Error: Invalid count argument
│ 
│   on .terraform/modules/eks_al2023/modules/eks-managed-node-group/main.tf line 5, in data "aws_caller_identity" "current":
│    5:   count = var.create && var.account_id == "" ? 1 : 0
│ 
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the
│ resources that the count depends on.

```

## Additional context
Issue seems to come from these 2 lines
```
# in main.tf -> locals block
partition  = try(data.aws_partition.current[0].partition, "")
```
## Workaround 
remove depends_on field

boost-ben:
## Description
This PR addresses a regression introduced in v21.0.0 where the terraform-aws-eks module became incompatible with `depends_on`, breaking existing v20 user configurations with 'Invalid count argument' errors.

Rather than force some users to refactor their working infrastructure (without a migration path) for optimization gains they may not need, it gives users the choice to reduce the number of GET requests for `partition` and `account_id` across submodules.

## Motivation and Context 
Resolves #3440

## Breaking Changes
No

## How Has This Been Tested?
Tested with `examples/eks-managed-node-group`

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -18,8 +18,8 @@ data "aws_iam_session_context" "current" {
 locals {
   create = var.create && var.putin_khuylo
 
-  account_id = try(data.aws_caller_identity.current[0].account_id, "")
-  partition  = try(data.aws_partition.current[0].partition, "")
+  account_id = var.enable_data_source_optimizations ? try(data.aws_caller_identity.current[0].account_id, "") : ""
+  partition  = var.enable_data_source_optimizations ? try(data.aws_partition.current[0].partition, "") : ""
 
   role_arn = try(aws_iam_role.this[0].arn, var.iam_role_arn)
 
--- a/variables.tf
+++ b/variables.tf
@@ -22,6 +22,12 @@ variable "tags" {
   default     = {}
 }
 
+variable "enable_data_source_optimizations" {
+  description = "Enable data source optimizations to reduce API calls."
+  type        = bool
+  default     = true
+}
+
 ################################################################################
 # Cluster
 ################################################################################


bryantbiggs:
again, this is not a valid change, nor will it solve the "problem" you are claiming

the authors of the language state that you should not do this https://github.com/hashicorp/terraform/issues/30340#issuecomment-1010202582

andreiboost:
I understand using `depends_on` is (heavily) discouraged, this whole situation reminds me of https://xkcd.com/1172/. However the truth is that something like this PR _does_ fix planning when `depends_on` is used on the EKS module.

Would you consider approving a PR making the two datasources optional, by adding `var.partition` and `var.account_id`?

PR #3490, opened 2025-08-24
fix: Correct metadata options loop condition due to variable definition defaults

### LINKED ISSUE #3482: "for_each" condition for dynamic "metadata_options" in aws_launch_template eks-managed-node-group module doesn't match variable parameter "nullable"
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/82acedc0593723979dc653d726617884cf7ab447/modules/eks-managed-node-group/main.tf#L227

`for_each = var.metadata_options != null ? [var.metadata_options] : []`

it doesn't make sense to check if var.metadata_options != null, because according to variables.tf it can't be 'null' -> nullable = false.
So, even if we pass 'null' or empty map {}, in any case the default values will be applied.

`variable "metadata_options" {
  description = "Customize the metadata options for the instance"
  type = object({
    http_endpoint               = optional(string, "enabled")
    http_protocol_ipv6          = optional(string)
    http_put_response_hop_limit = optional(number, 1)
    http_tokens                 = optional(string, "required")
    instance_metadata_tags      = optional(string)
  })
  default = {
    http_endpoint               = "enabled"
    http_put_response_hop_limit = 1
    http_tokens                 = "required"
  }
  nullable = false
}`

Please remove parameter 'nullable = false' for variable "metadata_options" so that we may have an opportunity not to set 'metadata_options' block.

Thanks!

bryantbiggs:
## Description
- Correct metadata options loop condition due to variable definition defaults

## Motivation and Context
- Resolves #3482

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -224,7 +224,7 @@ resource "aws_launch_template" "this" {
   }
 
   dynamic "metadata_options" {
-    for_each = var.metadata_options != null ? [var.metadata_options] : []
+    for_each = [var.metadata_options]
 
     content {
       http_endpoint               = metadata_options.value.http_endpoint
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -342,7 +342,7 @@ resource "aws_launch_template" "this" {
   }
 
   dynamic "metadata_options" {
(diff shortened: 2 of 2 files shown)


antonbabenko:
This PR is included in version 21.1.1 :tada:

PR #3515, opened 2025-09-16
fix: Disabling EKS auto-mode after it has been enabled

mirza-opsguru:
## Description

Changed the content in `eks_cluster` resource for dynamic blocks `kubernetes_network_config.elastic_load_balancing` and `storage_config` to be normal blocks instead of dynamic, so they are always resolved to either `true` or `false` (default).

## Motivation and Context

The problem is whenever you enable EKS auto-mode, if you want to disable it in the future, it won't be possible to do it.
The reason behind this is that the dynamic blocks mentioned above are managed by local variable called `auto_mode_enabled`, which gets its value from `try(var.compute_config.enabled, false)`, so it's either true or false.

```hcl
dynamic "kubernetes_network_config" {
    # Not valid on Outposts
    for_each = local.create_outposts_local_cluster ? [] : [1]

    content {
      dynamic "elastic_load_balancing" {
        for_each = local.auto_mode_enabled ? [1] : []

        content {
          enabled = local.auto_mode_enabled
        }
      }

      ip_family         = var.ip_family
      service_ipv4_cidr = var.service_ipv4_cidr
      service_ipv6_cidr = var.service_ipv6_cidr
    }
  }
  ...
  dynamic "storage_config" {
    for_each = local.auto_mode_enabled ? [1] : []

    content {
      block_storage {
        enabled = local.auto_mode_enabled
      }
    }
  }
```
As seen in the blocks, if the local variable resolves to false, it won't create the dynamic blocks, which is great if you never enabled eks auto-mode, but if you ever enabled it, it will try to resolve the values to `null` instead of `false`, which terraform and AWS provider won't allow. 
You'll get error like this:
```bash
  │ Error: compute_config.enabled, kubernetes_networking_config.elastic_load_balancing.enabled, and storage_config.block_storage.enabled must all be set to either true or false
  │ 
  │   with module.eks.aws_eks_cluster.this[0],
  │   on .terraform/modules/eks/main.tf line 35, in resource "aws_eks_cluster" "this":
  │   35: resource "aws_eks_cluster" "this" {
```

https://github.com/hashicorp/terraform-provider-aws/pull/42483
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3273
These issues are not opened, but they haven't been resolved before.

## Breaking Changes

It doesn't break anything, since the blocks for these values shouldn't be dynamic in the newer providers versions, they should always be set to either `true` or `false` (all 3 blocks).
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3273#issuecomment-2583220625

## How Has This Been Tested?

I have tested and validated these changes on a live project. (these changes were made for the project to get it working after the customer decided to disable EKS auto-mode after enabling it first)


antonbabenko:
This issue has been resolved in version 21.3.0 :tada:

Issue #3525, opened 2025-09-24
cluster_addons before_compute does not guarantee addons are created before compute resources


vincentmrg:
## Description

The `cluster_addons` feature in the module provides a `before_compute` attribute, which is described as a way to ensure that certain addons are created before compute resources (managed by `module.eks_managed_node_group` / `module.self_managed_node_group`).

In practice, however, this attribute does not guarantee that addons are created before compute resources. Instead, it only ensures that compute resources are created before addons with `before_compute = false` https://github.com/terraform-aws-modules/terraform-aws-eks/blob/de2aa10f25c7f2d2ab1264f6451f7cbf57f784c4/main.tf#L801

This behavior occurs because the modules responsible for creating compute resources currently do not have a `depends_on` referencing the addons with `before_compute = true`.

- [ X] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions
- Module version: `21.3.1`

## Expected behavior

Addons with `before_compute = true` should be applied before any compute resources.

The modules `module.eks_managed_node_group` and `module.self_managed_node_group` should include a `depends_on` on the `aws_eks_addon.before_compute` resources.

[Terraform documentation on depends_on for modules](https://developer.hashicorp.com/terraform/language/block/module#depends_on).

bryantbiggs:
please describe the underlying problem you are facing (i.e. - "I did this, and expected this, but got this")

vincentmrg:
I deployed an EKS cluster with some`cluster_addons` configured with `before_compute = true`. I expected them to be created before compute resources, they were not.

bryantbiggs:
well, that will depend. hard to say without reproductions

vincentmrg:
You're right, it's difficult to reproduce.

For my part, this happens sometimes when applying my infrastructure as code; it's not systematic depending on the order in which the resources are created.

I'm not going to share my entire setup here, but with a cluster_addons setup like this:

```
cluster_addons = {
    coredns    = {}
    kube-proxy = {}
    vpc-cni = {
      # CNI should be installed before worker nodes to prevent the following error on nodes:
      # KubeletNotReady              container runtime network not ready: NetworkReady=false reason: NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
      before_compute = true
    }
  }
```

I have a failure to apply my plan on node groups creation like this:
```
[...]

PR #3529, opened 2025-09-25
fix: Ensure addons with before_compute are created before compute resources

vincentmrg:
## Description

Fix issue described here: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3525

## Motivation and Context

Addons with before_compute = true should be applied before any compute resources.

The modules module.eks_managed_node_group and module.self_managed_node_group should include a depends_on on the aws_eks_addon.before_compute resources.

## Breaking Changes

No breaking change

## How Has This Been Tested?

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -382,6 +382,10 @@ module "eks_managed_node_group" {
     var.tags,
     each.value.tags,
   )
+
+  depends_on = [
+    aws_eks_addon.before_compute,
+  ]
 }
 
 ################################################################################
@@ -533,4 +537,8 @@ module "self_managed_node_group" {
     var.tags,
     each.value.tags,
   )
+
+  depends_on = [
+    aws_eks_addon.before_compute,
+  ]
 }


bryantbiggs:
This is incorrect and will cause disruptions for users

vincentmrg:
I tested it myself without finding any issues.
Perhaps I missed something. Is it possible to elaborate a little on how this proposal is incorrect and could cause disruptions for users?
The issue that led to this PR remains relevant; it would be nice to be able to move forward on the issue to help continue improving this module.

bryantbiggs:
> I tested it myself without finding any issues.

[here](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3525#issuecomment-3333294514) you said its difficult to reproduce, so how do you know this resolves anything? 

>  Is it possible to elaborate a little on how this proposal is incorrect and could cause disruptions for users?

See https://github.com/hashicorp/terraform/issues/30340

>  The issue that led to this PR remains relevant; it would be nice to be able to move forward on the issue to help continue improving this module.

Not saying the module is not without imperfections; there are scenarios where the Terraform language simply does not support the desired outcome. However, a module that is seeing over a million downloads a week I would expect to see more users reporting issues if there is truly an issue. I don't see that, I just see this one edge case that is not very well defined and eagerly jumping to a code solution that has known defects. and that is ultimately why the PR is not valid - it would cause more harm than good

vincentmrg:
I'm fully aware that this module is widely used and that caution is obviously required.
I opened this PR to illustrate a possible solution to the issue https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3525, which contains more details about the problem encountered.

The problem occurs when nodegroups are deployed before the vpc-cni cluster_addon.
When this happens, nodes from nodegroups never go into Ready, remaining in this state indefinitely:
```
 # CNI should be installed before worker nodes to prevent the following error on nodes:
KubeletNotReady              container runtime network not ready: NetworkReady=false reason: NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
```

> https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3525#issuecomment-3333294514 you said its difficult to reproduce, so how do you know this resolves anything?

This ensures that addons deployed with `before_compute = true` are deployed before any compute resource.

**Below** is a graph of the module before this commit

graph_before

**And** here the same graph after the commit

graph_after


We can clearly see from the graphs that `aws_eks_addon.before_compute` resources are not guaranteed to be deployed before compute resources (`module.self_managed_node_group` and `module.eks_managed_node_group`, where my proposal allows it.

bryantbiggs:
> This ensures that addons deployed with before_compute = true are deployed before any compute resource.

Yes, but it also guarantees that any changes to addons cause *ALL compute modules to be re-computed and cause disruptions. This is why we have taken more of a "best effort" route that is not disruptive https://github.com/terraform-aws-modules/terraform-aws-eks/blob/de2aa10f25c7f2d2ab1264f6451f7cbf57f784c4/main.tf#L799-L803

If you have compute already deployed, it will not work - its not designed for this scenario. If you are creating a new cluster from scratch, it should work 99/100 times

PR #3539, opened 2025-10-07
fix: removed use of dns_suffix

myerscf:
Using a Regions DNS Suffix is not an accurate way to determine the Service Principal Name for a service in different AWS Partitions. 

This MR removes its use, in preference of the TF `aws_service_principal` resource, that will better determine the name and not cause deployment issues in Partitions.

### DIFF
--- a/modules/karpenter/main.tf
+++ b/modules/karpenter/main.tf
@@ -12,7 +12,6 @@ data "aws_caller_identity" "current" {
 
 locals {
   account_id = try(data.aws_caller_identity.current[0].account_id, "")
-  dns_suffix = try(data.aws_partition.current[0].dns_suffix, "")
   partition  = try(data.aws_partition.current[0].partition, "")
   region     = try(data.aws_region.current[0].region, "")
 }
@@ -23,6 +22,10 @@ locals {
 
 locals {
   create_iam_role = var.create && var.create_iam_role
+}
+
+data "aws_service_principal" "ec2" {
+  service_name = "ec2"
 }
 
 data "aws_iam_policy_document" "controller_assume_role" {
@@ -271,7 +274,7 @@ data "aws_iam_policy_document" "node_assume_role" {
 
     principals {
       type        = "Service"
-      identifiers = ["ec2.${local.dns_suffix}"]
+      identifiers = [aws_service_principal.ec2.name]
     }
   }
 }
--- a/modules/karpenter/policy.tf
+++ b/modules/karpenter/policy.tf
@@ -234,7 +234,7 @@ data "aws_iam_policy_document" "controller" {
     condition {
       test     = "StringEquals"
       variable = "iam:PassedToService"
-      values   = distinct(["ec2.${local.dns_suffix}", "ec2.amazonaws.com"])
+      values   = [aws_service_principal.ec2.name]
     }
   }
 


bryantbiggs:
Are you facing an issue

myerscf:
Hi @bryantbiggs - yes, when trying to deploy into some AWS Partitions - using that Partitions domain to create the SPN, results in throwing an Invalid principal in policy validation error when deploying. Using the DNS suffix is an old, deprecated way of creating one.

[aws_service_principal](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/service_principal) solves this, by creating the correct, backwards compatible SPN for a service, depending on which region you are deploying to :)

bryantbiggs:
I'd like to learn more about this error you encountered, the error message, and what partition. the `aws_service_principal` data source is a relatively new source

myerscf:
Sure!

So in some of the old AWS partitions, using the DNS suffix was accepted as part of the SPN, but this was quickly deprecated as more Partitions where created in favour of a standardised `*.amazonaws.com`. In the regions where it was first allowed, they will allow either - but all services where updated in favour of the standardised naming convention. 

If you use the DNS suffix in the newer ones, where its not accepted you get an error like:

```
Error: creating IAM Role (<ROLE NAME>): operation error IAM: CreateRole, https response error StatusCode: 400,
RequestID: <REQUEST UUID>, MalformedPolicyDocument: Invalid principal in policy: "SERVICE":"ec2.c2s.ic.gov"
```

The `aws_service_prinicipal` resource was created, to match similar functionality in [CDK](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/region-info/lib/default.ts) - which allows users to create cross compatible SPNs in their policies, but enabling backwards compatibility for edge cases.
So for example if there is a region with an edge case say `logs` [in the CN partition](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/meta/service_principal_data_source.go#L109C6-L109C10) then it will return with the SPN with the DNS suffix - it not, the standardised one!


So, this change should keep things simple, and allow the modules to be easily deployed to different regions/partitions :D

bryantbiggs:
thanks for the information, we use this pattern quite a bit so I wanted to understand better, and we've also encountered issues with partitions (mostly because services don't do a great job at defining when the partition/suffix should change)

modules/karpenter/main.tf
```diff
@@ -12,7 +12,6 @@ data "aws_caller_identity" "current" {
 
 locals {
   account_id = try(data.aws_caller_identity.current[0].account_id, "")
-  dns_suffix = try(data.aws_partition.current[0].dns_suffix, "")
```
bryantbiggs, line 15: keep, just update with the new data source

myerscf: do you mean leave the `dns_suffix` as a local var, or update to have the `data.aws_service_principal.ec2.name` as its value? assuming the former, it would also leave the value unused 🤔 if the latter it would be wrong a different value not representing the dns_suffix!

myerscf:
> thanks for the information, we use this pattern quite a bit so I wanted to understand better, and we've also encountered issues with partitions (mostly because services don't do a great job at defining when the partition/suffix should change)

Updated to address the recommendations!

Will be happy to look through the other repos in `terraform-aws-modules` and PR similar changes to help with their use across AWS Partitions!

myerscf:
Leaving the `dns_suffix` prop leaves us with a linting issue :( do you still want it left in?

`Warning: [Fixable] local.dns_suffix is declared but not used (terraform_unused_declarations)`

bryantbiggs:
> Leaving the `dns_suffix` prop leaves us with a linting issue :( do you still want it left in?
> 
> `Warning: [Fixable] local.dns_suffix is declared but not used (terraform_unused_declarations)`

see [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3539#discussion_r2410861683) - the new local variable should have replaced this one

bryantbiggs (approved):

antonbabenko:
This PR is included in version 21.6.0 :tada:

PR #3562, opened 2025-10-27
feat: Add recommended security group rule for port `10251` to match EKS addon for `metrics-server`

### LINKED ISSUE #3557: metrics-server addon default port 10251 not covered by default security group rules
## Description

Right now the default EKS community addon for `metrics-server` requires a manual config for the port set up by `terraform-aws-eks` in the list of [default security group rules](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/96dbaa033bdc9e0a23421eb77c87b4533c44a225/node_groups.tf#L120) to match:

```hcl
addons = {
    ...
    metrics-server = {
      most_recent          = true
        containerPort = 4443
      })
    }
  }
```

**Alternatively**, one could add another security group rule covering port 1025**1**, the aws addon distribution's default ([documented here](https://docs.aws.amazon.com/eks/latest/userguide/metrics-server.html#:~:text=The%20Amazon%20EKS%20add%2Don%20version%20of%20Metrics%20Server%20is%20pre%2Dconfigured%20to%20use%20port%2010251)) and visible in the schema when checking it vie the aws web gui, with the same parameters as the one for port 10250 (`kubelet` default, settings copied over and port++):

```hcl
  node_security_group_additional_rules = {
    ingress_cluster_metrics_server = {
      description                   = "Cluster API to metrics-server"
      protocol                      = "tcp"
      from_port                     = 10251
      to_port                       = 10251
      type                          = "ingress"
      source_cluster_security_group = true
    }
  }
```

The helm package default is using port 10250 (see [here](https://artifacthub.io/packages/helm/metrics-server/metrics-server#:~:text=containerPort,10250)), which is also the general `metrics-server` default if no other port is configured (see [here](https://github.com/kubernetes-sigs/metrics-server/blob/master/charts/metrics-server/values.yaml#L68)). The AWS guys just had to reserve port 10250 for Fargate, for some reason...

I could not find the current default of port 4443, used by this terraform module, being used by the metrics-server in any install method (EKS addon or direct helm install, or manually deployed). (Edit: [this seemed to be the default port a few years back](https://github.com/kubernetes-sigs/metrics-server/issues/1026), so historically this used to be right port to cover with a security group rule, but isn't anymore)

I therefore propose to change it to 10251 by default (to match the EKS docs and default setting, since this is an EKS module), and also highlight this in the docs/description somewhere for people who want to deploy it manually via helm or otherwise. In either case, [the modules configured default port 4443 for the security group rule dedicated to the metrics-server](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/96dbaa033bdc9e0a23421eb77c87b4533c44a225/node_groups.tf#L120), with the dedication indicated by the comment in that line, seems unused by current installation methods for the metrics-server, and should be changed.

## Versions

- Module version [Required]:
terraform-aws-modules/eks/aws  ~> 21.6.1
- Terraform version:
Terraform v1.13.4
- Provider version(s):
registry.terraform.io/hashicorp/aws ~> 6.17

see above

Create an EKS cluster with the community addon `metrics-server` added to the list of addons, with no additional rules of init configs added.

## Expected behavior

`kubectl top nodes` and `kubectl top pods` to work, which require the metrics-server to be reachable by the API server.

Even if this is to expected to be required as a pre-setup or config step, I couldn't find it documented anywhere.

## Actual behavior

`metrics-server` fails to be registered and getting listed under list of `api-resources`, above `kubectl` commands and `horizontal pod autoscaler` fail and the list of apiservices in k9s shows `FailedDiscoveryCheck` for the metrics-server.

Once the rule for port 10251 is added (even manually for a cluster that was just deployed), the `metrics-server` is instantly registered and listed under `api-resources` . For fresh clusters, adding the above extra rule in terraform also solves the issue, and alternatively setting the initial config for the container port inside the addon's block to `4443` also works by itself.

pwntr:
and add TODO note to remove the metrics-server legacy port 4443 on the next breaking change

## Description
metrics-server uses port 10251 as the default nowadays with the EKS provided bundle that can be installed as an addon. Port 4443 is the legacy port used by it in the past, but not anymore in the releases of the last 3 years. Addresses #3557.

## Motivation and Context
- Resolves #3557 

## Breaking Changes
no breaking change introduced, backwards compatible and just adding another recommended port to be added as a security rule. TODO note added to be considered for the next planned breaking change version bump at v22 to remove the legacy port security group rule then.

## How Has This Been Tested?

tested with default metrics-server addon added to the list of addons, without passing any configuration changes

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -128,7 +128,7 @@ locals {
     }
     # metrics-server, current EKS default port
     ingress_cluster_10251_webhook = {
-      description                   = "Cluster API to node 4443/tcp webhook"
+      description                   = "Cluster API to node 10251/tcp webhook"
       protocol                      = "tcp"
       from_port                     = 10251
       to_port                       = 10251


node_groups.tf
```diff
@@ -126,6 +126,15 @@ locals {
       type                          = "ingress"
       source_cluster_security_group = true
     }
+    # metrics-server, current EKS default port
+    ingress_cluster_10251_webhook = {
+      description                   = "Cluster API to node 4443/tcp webhook"
```
bryantbiggs:
```suggestion
      description                   = "Cluster API to node 10251/tcp webhook"
```

pwntr: good catch :)

antonbabenko:
This PR is included in version 21.7.0 :tada:

Issue #3579, opened 2025-11-13
EKS destroy fails after creation


rishangbhavsar:
## Description

When running `terragrunt destroy -var-file=./vars/dev.tfvars` using the EKS module version **21.8.0**, Terraform fails with an **Invalid count argument** error originating from `eks-managed-node-group/main.tf`.
The destroy operation cannot proceed due to dependencies within the module that rely on computed resource attributes, which are not yet known during plan/destroy.

This appears to be a regression or missing conditional handling in how `count` is computed for the following data sources:

* `data.aws_partition.current`
* `data.aws_caller_identity.current`

The issue occurs even though the environment is fully provisioned and operational.

---

## ⚠️ Note

I have already performed the following checks:

1. Removed the local `.terraform` directory

   ```bash
   rm -rf .terraform/
   ```
2. Re-initialized the project root

   ```bash
   terraform init
   ```
3. Re-ran the destroy command — the issue persists.

---

## Versions

* **Module version:** 21.8.0
* **Terraform version:** 1.13.5
* **Terragrunt version:** 0.93.3
* **Provider version(s):**

  ```
  + provider registry.terraform.io/hashicorp/aws 6.20.0
  ```

---

## Reproduction Code [Required]

```
terraform apply -var-file=vars.tfvars
```

```bash
terragrunt destroy -var-file=vars.tfvars
```

EKS module reference:

```hcl

# ref:
# - https://docs.aws.amazon.com/eks/latest/userguide/network-reqs.html

locals {
  eks_cluster_name     = "${local.prefix}-eks"
  eks_iam_provider_aud = "${replace(module.eks.cluster_oidc_issuer_url, "https://", "")}:aud"
  eks_iam_provider_sub = "${replace(module.eks.cluster_oidc_issuer_url, "https://", "")}:sub"
  eks_iam_provider_arn = module.eks.oidc_provider_arn
}

module "eks" {
  source     = "terraform-aws-modules/eks/aws"
  version    = "21.8.0"
  depends_on = [module.vpc]

  name               = "test-eks"
  kubernetes_version = "1.34"
  # Kubernetes Upgrade Policy - https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-version-support-policy
  upgrade_policy = {
    support_type = "STANDARD"
  }
  # Network configuration
  vpc_id                   = module.vpc.vpc_id
[...]

PR #3580, opened 2025-11-13
refactor: Simplify data source usage and local variable assignments for node group module

Rishang:
This PR is the fix for issue: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3579

Purpose of the PR is mentioned in the issue itself


Rishang:
I have tested the changes based on the issue

bryantbiggs:
these changes are incorrect - the current source code is correct as written. we create and destroy the examples and test cases quite often, this is a heavily used module by many users - I don't see a reproduction that re-creates the issue

PR #3583, opened 2025-11-16
fix: Correct conditional logic for compute role association

lukzhang:
This PR addresses an issue where the EKS cluster's compute role (Node Instance Role) association was not reliably created when no explicit node groups (eks_managed_node_groups or self_managed_node_groups) were defined.

The logic has been updated to ensure that the node IAM role is associated if:

compute_config.enabled is explicitly set to true (or implicitly defaults to true).

Any form of compute capacity (managed, self-managed, or Fargate) is defined.

This ensures the cluster's base configuration is complete and ready to accept compute resources added after initial deployment, even when relying on the module's implicit defaults.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -63,7 +63,7 @@ resource "aws_eks_cluster" "this" {
     content {
       enabled       = compute_config.value.enabled
       node_pools    = compute_config.value.node_pools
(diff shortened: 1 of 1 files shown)


bryantbiggs:
This is incorrect - it is correct as it is written on `master`

PR #3589, opened 2025-11-20
feat: Add variable for configuring Karpenter Cloudwatch Event rule name prefix

Weastie:
## Description
Added variable `rule_name_use_terraform_prefix` to the Karpenter submodule that allows disabling the Terraform generated suffix to Cloudwatch Event rules generated by the submodule.

## Motivation and Context
When directly using the Karpenter submodule, there is no configuration present to disable the Terraform suffix applied to the Cloudwatch Event rules. I personally ran into the problem of hitting the AWS imposed length limit of 64. 

The Cloudwatch Event rule for the InstanceStateChange rule, for example, uses 46 characters for the "InstanceStateChange-" part plus the 26 characters of Terraform's auto-generated suffix, leaving only 18 characters left for the user inputted `rule_name_prefix`.

## Breaking Changes
Backwards compatibility is provided because the variable is optional and defaults to the previous behavior when omitted.

## How Has This Been Tested?
Because of how small this change is, I just tested it by replacing the source of the Karpenter module in my code to my PR repo, and tested that:
* Omitting the `rule_name_use_terraform_prefix` variable resulted in no changes and no errors
* Setting the `rule_name_use_terraform_prefix` variable to `true` (default) resulted in no changes and no errors
* Setting the `rule_name_use_terraform_prefix` variable to `false` resulted in renaming of the Cloudwatch Event rules and no errors

### DIFF
--- a/modules/karpenter/variables.tf
+++ b/modules/karpenter/variables.tf
@@ -349,7 +349,7 @@ variable "rule_name_prefix" {
 }
 
 variable "rule_name_use_terraform_prefix" {
-  description = "Determines whether the generated rule name is used by Terraform as a prefix"
+  description = "Determines whether the generated rule name (`rule_name_prefix` + internal rule suffix) is used by Terraform as a prefix"
   type        = bool
   default     = true
 }


bryantbiggs:
thank you but we do not plan on accepting this change - the prefix prevents several errors or situations that lead to errors

PR #3600, opened 2025-11-26
Allow overriding of prefix separator in self-managed node groups

james-menzies:
## Description
Add a `prefix_separator` variable into the `self_managed_node_groups` variable schema, and use it to override the hard-coded '-' separator when providing `name_prefix` values to underlying AWS resources. This is functionally identical to the `prefix_separator` variable on the top level-variable definition.

## Motivation and Context
I've found that when upgrading from v17 when using a self-managed node group, that the `aws_launch_template` resource can take a while to replace (around 7 minutes) and in that time, the cluster is not able to receive any incoming traffic.

In the v18 upgrade guide, the strategy around using the `prefix_separator` to replace '-' with '' works well with respect to ensuring that the EKS cluster itself can be updated in-place, however the same strategy cannot be applied to the AWS Launch template resource due to the hard coded hyphen separator.

## Breaking Changes
No breaking changes. I did consider propagating the `prefix_separator` directly from the top-level variable schema, however decided against that as it would potentially create breaking behaviour for existing users.

## How Has This Been Tested?

For testing:
* I first applied the self-managed node groups example from the `master` branch. 
* I then changed to the head of this branch and re-ran `terraform plan`. 
* There were no changes (expected). 
* Then I added an explicit `prefix_separator` of `-`. 
* Again, no changes (expected). 
* Then I changed the separator to an empty string. 
* All affected resources that use the prefix separator needed to be replaced (expected).

I didn't see any use of straying from using the default names in the `examples` directory, so I didn't add anything to it, but am happy to update it if desired.

### DIFF
--- a/node_groups.tf
+++ b/node_groups.tf
@@ -417,7 +417,7 @@ module "self_managed_node_group" {
 
   name             = coalesce(each.value.name, each.key)
   use_name_prefix  = each.value.use_name_prefix
-  prefix_separator = each.value.prefix_separator
+  prefix_separator = each.value.prefix_separator != null ? each.value.prefix_separator : "-"
 
   availability_zones = each.value.availability_zones
   subnet_ids         = coalesce(each.value.subnet_ids, var.subnet_ids)


bryantbiggs:
Thank you, but unfortunately this is not a change that we are looking to carry forward

Issue #3604, opened 2025-12-02
Trying to use v21 but fails due to terraform constraints


erezo9:
## Description
so this is a duplicate of this but the original was closed
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3558

the issue lies with when trying to spin up new clusters with eks_managed_node_groups - without that blocks it works just fine

from what i saw
the issues comes  from these line (in a chain of depends)
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/eks-managed-node-group/main.tf#L5
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L282
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/main.tf#L21

the datasource of the partition inside the managed node groups is calling a datasource which is also called before on the main of the eks

- [ V] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]:

- Terraform version:

- Provider version(s):

## Reproduction Code [Required]
module "eks" {
  source                          = "terraform-aws-modules/eks/aws"
  name                    = local.cluster_name
  kubernetes_version                 = var.cluster_version
  create_kms_key                  = true
  enable_kms_key_rotation         = true
  create_iam_role                 = true
  iam_role_name                   = "${local.cluster_name}"
  iam_role_use_name_prefix        = true
  vpc_id                          = local.vpc_id
  subnet_ids                      = data.aws_subnets.this.ids
  endpoint_public_access  = false
  deletion_protection     = var.deletion_protection != null ? var.deletion_protection : false

  eks_managed_node_groups = {
    one = {
      instance_types = [var.instance_types]
      min_size       = var.min_nodes
      create = true
      max_size       = var.max_nodes
      desired_size   = var.current_nodes
    }
  }
  
  authentication_mode                      = "API_AND_CONFIG_MAP"

}

 No
 yes

## Expected behavior
terraform creates the cluster

## Actual behavior

terraform is not able to compute resouces

### Terminal Output Screenshot(s)
│ Error: Invalid count argument
│
[...]

Issue #2612, opened 2023-05-12
Unable create new Cluster with version 19.13.1 (latest)


quyennguyenvan:
## Description

Please provide a clear and concise description of the issue you are encountering, and a reproduction of your configuration (see the `examples/*` directory for references that you can copy+paste and tailor to match your configs if you are unable to copy your exact configuration). The reproduction MUST be executable by running `terraform init && terraform apply` without any further changes.

If your request is for a new feature, please use the `Feature request` template.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

terraform -v
Terraform v1.3.9
on windows_amd64
+ provider registry.terraform.io/hashicorp/aws v4.66.1
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.2
+ provider registry.terraform.io/hashicorp/kubernetes v2.20.0
+ provider registry.terraform.io/hashicorp/random v3.5.1
+ provider registry.terraform.io/hashicorp/time v0.9.1
+ provider registry.terraform.io/hashicorp/tls v4.0.4

Your version of Terraform is out of date! The latest version
is 1.4.6. You can update by downloading from https://www.terraform.io/downloads.html

<on-windows-11>

## Reproduction Code [Required]

1. have using workspace 
2. Have cleared the local cached

3. step of produce

terraform init
terraform plan

Code Snippet to Reproduce

```
module "eks" {
  source          = "terraform-aws-modules/eks/aws"
  version         = "19.13.1"
  cluster_name    = var.cluster_name
  cluster_version = var.cluster_version

  enable_irsa = true

  create_iam_role = true
  iam_role_name   = "${var.cluster_name}-ClusterRole"

  cluster_addons = {
    coredns = {
      resolve_conflicts = "OVERWRITE"
    }
    kube-proxy = {
      most_recent = true
    }
    vpc-cni = {
      resolve_conflicts = "OVERWRITE"
    }
  }
  cluster_enabled_log_types = [
    "api",
    "audit",
    "authenticator",
    "controllerManager",
    "scheduler",
  ]

  vpc_id                               = var.vpc_id
[...]

PR #3605, opened 2025-12-02
fix: Change partition due to dependency in terraform

erezo9:
## Description
Due to dependency between the terraform of eks and the managed node groups
managed node groups should be able to run on its own - while allowing running from the eks module

it seems that the change to allow this was breaking and this should allow the terraform to run correctly
this is the behavior in hybrid_node_role and karpenter submodules

## Motivation and Context
fixes this issue
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3604

## Breaking Changes
this shouldnt break anything since it uses the same partition and accunt

## How Has This Been Tested?
- [V ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ V] I have tested and validated these changes using one or more of the provided `examples/*` projects

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -1,13 +1,13 @@
 data "aws_partition" "current" {
-  count = var.create && var.partition == "" ? 1 : 0
+  count = var.create ? 1 : 0
 }
 data "aws_caller_identity" "current" {
-  count = var.create && var.account_id == "" ? 1 : 0
+  count = var.create ? 1 : 0
 }
 
 locals {
-  partition  = try(data.aws_partition.current[0].partition, var.partition)
-  account_id = try(data.aws_caller_identity.current[0].account_id, var.account_id)
+  partition  = try(data.aws_partition.current[0].partition, "")
+  account_id = try(data.aws_caller_identity.current[0].account_id, "")
 }
 
 ################################################################################
--- a/modules/eks-managed-node-group/variables.tf
+++ b/modules/eks-managed-node-group/variables.tf
@@ -17,18 +17,6 @@ variable "region" {
   default     = null
 }
 
-variable "partition" {
-  description = "The AWS partition - pass through value to reduce number of GET requests from data sources"
-  type        = string
-  default     = ""
-}
-
-variable "account_id" {
-  description = "The AWS account ID - pass through value to reduce number of GET requests from data sources"
-  type        = string
-  default     = ""
-}
-
 ################################################################################
 # User Data
 ################################################################################
--- a/modules/fargate-profile/main.tf
+++ b/modules/fargate-profile/main.tf
@@ -4,15 +4,15 @@ data "aws_region" "current" {
   region = var.region
 }
 data "aws_partition" "current" {
-  count = var.create && var.partition == "" ? 1 : 0
+  count = var.create ? 1 : 0
 }
 data "aws_caller_identity" "current" {
-  count = var.create && var.account_id == "" ? 1 : 0
+  count = var.create ? 1 : 0
 }
 
 locals {
-  account_id = try(data.aws_caller_identity.current[0].account_id, var.account_id)
-  partition  = try(data.aws_partition.current[0].partition, var.partition)
+  account_id = try(data.aws_caller_identity.current[0].account_id, "")
+  partition  = try(data.aws_partition.current[0].partition, "")
   region     = try(data.aws_region.current[0].region, "")
 }
 
--- a/modules/fargate-profile/variables.tf
+++ b/modules/fargate-profile/variables.tf
@@ -18,18 +18,6 @@ variable "region" {
   default     = null
 }
 
-variable "partition" {
-  description = "The AWS partition - pass through value to reduce number of GET requests from data sources"
-  type        = string
-  default     = ""
-}
-
-variable "account_id" {
-  description = "The AWS account ID - pass through value to reduce number of GET requests from data sources"
-  type        = string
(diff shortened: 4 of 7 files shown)


erezo9:
@bryantbiggs @antonbabenko 
can you please go through this PR?
it should fix some issues when creating new clusters from scratch
it goes on the same behavior as other submodules

bryantbiggs:
thank you but this is not a valid change - we are not seeing issues with new cluster creation as the module is currently configured

erezo9:
@bryantbiggs 
its only clusters with managed node groups
if i dont have a node group it works fine

bryantbiggs:
we aren't seeing any issue with those - or any of our examples/tests that we provide

erezo9:
@bryantbiggs 
so I found when it happens exactly if you want to try and replicate the issue

If the module depends on another resource it will not make it after plan
Do you want me to provide reproduction code?

lestich:
We are experiencing the same issue. Creating a cluster from scratch with managed nodegroups fails due to "invalid count argument". Even with setting var.account_id and var.partition it still fails.

erezo9:
@lestich do you have depends_on on your module?

lestich:
@erezo9 yes we depend on one additonal resource (aws_vpc_security_group_egress_rule) i.e. nothing related to the eks module imho

erezo9:
@bryantbiggs
@antonbabenko 
this is from the examples

with explicit depends_on
```
module "eks_al2023" {
  version = "~> 21.0"

  name               = "${local.name}-al2023"
  kubernetes_version = "1.33"

  # EKS Addons
  addons = {
    coredns = {}
    eks-pod-identity-agent = {
      before_compute = true
    }
    kube-proxy = {}
    vpc-cni = {
      before_compute = true
    }
  }

  eks_managed_node_groups = {
    example = {
      # Starting on 1.30, AL2023 is the default AMI type for EKS managed node groups
      instance_types = ["m6i.large"]
      ami_type       = "AL2023_x86_64_STANDARD"

      min_size = 2
      max_size = 5
      # This value is ignored after the initial creation
      # https://github.com/bryantbiggs/eks-desired-size-hack
      desired_size = 2

      # This is not required - demonstrates how to pass additional configuration to nodeadm
      # Ref https://awslabs.github.io/amazon-eks-ami/nodeadm/doc/api/
      cloudinit_pre_nodeadm = [
        {
          content_type = "application/node.eks.aws"
          content      = <<-EOT
            ---
            apiVersion: node.eks.aws/v1alpha1
            kind: NodeConfig
            spec:
              kubelet:
                config:
                  shutdownGracePeriod: 30s
          EOT
        }
      ]
    }
  }
  depends_on = [ module.vpc ]
  tags = local.tags
}

```

returns 
image

without explicit depends_on

```
module "eks_al2023" {
  version = "~> 21.0"

  name               = "${local.name}-al2023"
  kubernetes_version = "1.33"

  # EKS Addons
  addons = {
    coredns = {}
    eks-pod-identity-agent = {
      before_compute = true
    }
    kube-proxy = {}
    vpc-cni = {
      before_compute = true
    }
  }

  eks_managed_node_groups = {
    example = {
      # Starting on 1.30, AL2023 is the default AMI type for EKS managed node groups
      instance_types = ["m6i.large"]
      ami_type       = "AL2023_x86_64_STANDARD"

      min_size = 2
      max_size = 5
      # This value is ignored after the initial creation
      # https://github.com/bryantbiggs/eks-desired-size-hack
      desired_size = 2

      # This is not required - demonstrates how to pass additional configuration to nodeadm
      # Ref https://awslabs.github.io/amazon-eks-ami/nodeadm/doc/api/
      cloudinit_pre_nodeadm = [
        {
          content_type = "application/node.eks.aws"
          content      = <<-EOT
            ---
            apiVersion: node.eks.aws/v1alpha1
            kind: NodeConfig
            spec:
              kubelet:
                config:
                  shutdownGracePeriod: 30s
          EOT
        }
      ]
    }
  }
  tags = local.tags
}

```
returns
image

bryantbiggs:
wow, a reproduction! ironic how your original code snippet (not reproduction) on the issue did not include this detail

and yes, you do not use `depends_on` on a module. You don't need to explicitly depend on the VPC module, you can see we do not. I don't know why you would define an egress rule outside of this module and then depend on it, but I don't see the full code to know more

either way, yes - you do not use `depends_on` on a module

erezo9:
@bryantbiggs  the depends_on vpc was just for an example
i have a null_resource and sqs queue which should run before the eks runs
and if i cant add depends_on it causes an issue

 why not just allow the eks_module to depend on other resources? 
if its by design then this should be added to the docs that the eks_module cannot depend explicitly on any other reource

also i cleaned part of the code so i can attach it, probably removed the depends_on on the way - my fault

bryantbiggs:
its not up to the module, this is just how Terraform works https://github.com/hashicorp/terraform/issues/30340#issuecomment-1010202582

99.999% of the time the problems are resolved by simply changing the way in which you are configuring your Terraform. Its not a module issue

erezo9:
but it was like that in version 20, this was simply added for version 21 for the managed node groups and self managed node groups,
the only possible way for anyone who had an eks with depends_on because he wanted to control the resource creation timing is to do
module eks
{}
module managed_nodegroup
{}

is that how i should migrate the code to?

bryantbiggs:
no, you don't know what you are talking about https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2612

locking because there is nothing further to discuss. you simply are using Terraform incorrectly

PR #3631, opened 2026-01-20
feat: Added link to AWS EKS workshops

antonbabenko:
Added it right above the "Usage" section. Using long static links to assets to make sure it works on all platforms, like HashiCorp and GitLab registries.

@bryantbiggs I think that if we put it right under the Ukrainian flag, it will seem very much connected together.


bryantbiggs (approved):

antonbabenko:
This PR is included in version 21.15.0 :tada:

Issue #3650, opened 2026-03-02
Karpenter module not compatible with AWS China


fitillo:
## Description

Policy created for karpenter controller doesn't work for AWS China. The problem is inside controller's policy: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v21.15.1/modules/karpenter/policy.tf#L237

Code contemplates adding ec2.amazonaws.com.cn when executing in China. The problem is that aws_service_principal is not working as expected. There's an open issue for it here: https://github.com/hashicorp/terraform-provider-aws/issues/46209 that I'm able to reproduce.

I propose reverting to previous code in v20. I've already tested it in my local setup and it works.

1. Add local `dns_suffix = data.aws_partition.current[0].dns_suffix` in main.tf
2. Modify line 237 in policy.tf to: `values   = distinct(["ec2.${local.dns_suffix}", "ec2.amazonaws.com"])`

I'd be happy to add the PR for this one.

fitillo:
Added PR: https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3651

bryantbiggs:
it is correct as written, the upstream provider needs to be corrected

fitillo:
@bryantbiggs Issue in the upstream provider was fixed https://github.com/hashicorp/terraform-provider-aws/issues/46209 I tested using the new version and it works fine. Thanks!

bryantbiggs:
Fantastic news. We should update the min supported provider version here so that users are guaranteed to have a working module on the other partitions

fitillo:
Sure, it is 6.39.0

PR #3651, opened 2026-03-03
fix: Fix karpenter policy to set proper conditions to ec2 service in China

fitillo:
Policy created for karpenter controller doesn't work for AWS China. The problem is inside controller's policy: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v21.15.1/modules/karpenter/policy.tf#L237

Code contemplates adding ec2.amazonaws.com.cn when executing in China. The problem is that aws_service_principal is not working as expected. There's an open issue for it here: https://github.com/hashicorp/terraform-provider-aws/issues/46209 that I'm able to reproduce.

This PR fixes above issue.


bryantbiggs:
https://github.com/terraform-aws-modules/terraform-aws-eks/issues/3650#issuecomment-3991103651

PR #3660, opened 2026-03-19
fix: Avoid attempting to retrieve the AMI ID from SSM parameter if a custom AMI ID is provided

samuelallan72:
## Description

The ssm param for the recommended ami is not used if a custom ami is used. Also, if one wishes to use a custom ami with the older AL2 ami_type (uses bootstrap.sh for user_data in launch templates), retrieving the ssm param fails with k8s >= 1.33 due to the AL2 images being removed:

    Error: reading SSM Parameter (/aws/service/eks/optimized-ami/1.34/amazon-linux-2/recommended/image_id): couldn't find resource

Private-ref: https://tasks.opencraft.com/browse/BB-10631

## Motivation and Context

This fixes using self managed node groups with Ubuntu AMIs.

## Breaking Changes

I don't believe this breaks backward compatibility.

## How Has This Been Tested?

I've tested this patch based on v20.37.2. Ideally we would backport it to the v20 release series, since we're using that at the moment and cannot update right now.

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -43,7 +43,7 @@ locals {
 
 data "aws_ssm_parameter" "ami" {
   # only pull the ami if we're creating the resource AND we're not already using a custom ami
-  count = var.create && length(var.ami_id) == 0 ? 1 : 0
+  count = var.create && var.ami_id == "" ? 1 : 0
 
   region = var.region
 
@@ -220,7 +220,7 @@ resource "aws_launch_template" "this" {
     arn = var.create_iam_instance_profile ? aws_iam_instance_profile.this[0].arn : var.iam_instance_profile_arn
   }
 
-  image_id                             = coalesce(var.ami_id, nonsensitive(try(data.aws_ssm_parameter.ami[0].value, "")))
+  image_id                             = var.ami_id == "" ? nonsensitive(data.aws_ssm_parameter.ami[0].value) : var.ami_id
   instance_initiated_shutdown_behavior = var.instance_initiated_shutdown_behavior
 
   dynamic "instance_market_options" {
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -42,7 +42,8 @@ locals {
 }
 
 data "aws_ssm_parameter" "ami" {
-  count = var.create ? 1 : 0
+  # only pull the ami if we're creating the resource AND we're not already using a custom ami
+  count = var.create && length(var.ami_id) == 0 ? 1 : 0
 
   region = var.region
 
@@ -219,7 +220,7 @@ resource "aws_launch_template" "this" {
     arn = var.create_iam_instance_profile ? aws_iam_instance_profile.this[0].arn : var.iam_instance_profile_arn
   }
 
-  image_id                             = coalesce(var.ami_id, nonsensitive(data.aws_ssm_parameter.ami[0].value))
+  image_id                             = coalesce(var.ami_id, nonsensitive(try(data.aws_ssm_parameter.ami[0].value, "")))
   instance_initiated_shutdown_behavior = var.instance_initiated_shutdown_behavior
 
   dynamic "instance_market_options" {


samuelallan72:
Please let me know what would be preferred regarding updating examples here. I could add a new one in examples/self-managed-node-group/ perhaps, with a module that deploys eks k8s 1.33 with ami_type AL2 and an ubuntu ami? I'm not sure if more examples are desired here though.

modules/self-managed-node-group/main.tf
```diff
@@ -219,7 +220,7 @@ resource "aws_launch_template" "this" {
     arn = var.create_iam_instance_profile ? aws_iam_instance_profile.this[0].arn : var.iam_instance_profile_arn
   }
 
-  image_id                             = coalesce(var.ami_id, nonsensitive(data.aws_ssm_parameter.ami[0].value))
+  image_id                             = coalesce(var.ami_id, nonsensitive(try(data.aws_ssm_parameter.ami[0].value, "")))
```
bryantbiggs:
```suggestion
  image_id                             = var.ami_id == null ? nonsensitive(data.aws_ssm_parameter.ami.value) : var.ami_id 
```

bryantbiggs (changes requested):

modules/self-managed-node-group/main.tf
samuelallan72: Updated!

modules/self-managed-node-group/main.tf
```diff
@@ -42,7 +42,8 @@ locals {
 }
 
 data "aws_ssm_parameter" "ami" {
-  count = var.create ? 1 : 0
+  # only pull the ami if we're creating the resource AND we're not already using a custom ami
+  count = var.create && length(var.ami_id) == 0 ? 1 : 0
```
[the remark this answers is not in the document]
samuelallan72: Thanks @bryantbiggs , updated! :)

bryantbiggs (approved):

antonbabenko:
This PR is included in version 21.17.1 :tada:

samuelallan72:
@antonbabenko @bryantbiggs excellent, thank you!

Would it be ok to backport this to v20? I'm planning to open a PR.

bryantbiggs:
unfortunately, we do not support back-porting since we only maintain and release from `master`, there aren't any other long living branches per major release version

samuelallan72:
@bryantbiggs ah I see, no worries, that makes sense! :slightly_smiling_face:

zied-jt:
@bryantbiggs can you please give more insight why this actual PR was accepted and not  #3545 with the same code ? 

It could be a useful feedback to have better contribution from community. thanks !

bryantbiggs:
1. It explained a valid use case and problem encountered (Ubuntu based AMIs)
2. The submitter was responsive (other PR was closed after no response)
3. The solution is semantically more correct - if an `ami_id` is provided, we don't need to perform the SSM parameter API call to find the AMI ID. The other PR was doing a string check against `"CUSTOM"` which is not great - its not clear what that means, the intent, etc. 


If you want to get changes reviewed and approved in open source projects - help maintainers understand so they can help you. If you look at a lot of the PRs and issues in our projects they are all the same "I hit a problem and I found it in your code - should be this". No context, no idea what the user is trying to do or why, nada - just one persons opinion that we are supposed to go off of and that simply isn't going to land very well the majority of the time

bryantbiggs:
also, the way that other PR was described, and what I responded - it is perfectly valid to have a custom AMI that is a derivative of an EKS AMI. and if you set the `ami_type` to what that base AMI was, this module will help bootstrap that node correctly with user data. so just simply specifying `"CUSTOM"` erases all information and its never clear what that means or why, so its not a path we actively encourage users. This is also why disabling the SSM param lookup if an `ami_id` is provided is the correct solution. If you have a custom AMI based of the EKS AL2023 x86 AMI, you can set the `ami_type` and `ami_id` to get the best scenario out of this module. If you were to follow the other PR and set `ami_type = "CUSTOM"`, you have quite a bit more work to do when you shouldn't need to

zied-jt:
@bryantbiggs thanks for taking time to explain.

PR #3665, opened 2026-04-01
feat: Add ECR Public permissions to EKS Auto Mode node IAM role

fletcherw:
## Description
Include permissions for authenticated container pulls from public ECR in the node roles used for EKS Auto clusters. 

## Motivation and Context
Without them, the pulls will still succeed, but they can be rate-limited, resulting in slow pod startup times.

## Breaking Changes
n/a

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -921,8 +921,9 @@ resource "aws_iam_role" "eks_auto" {
 # Policies attached ref https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html
 resource "aws_iam_role_policy_attachment" "eks_auto" {
   for_each = { for k, v in {
-    AmazonEKSWorkerNodeMinimalPolicy   = "${local.iam_role_policy_prefix}/AmazonEKSWorkerNodeMinimalPolicy",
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -921,8 +921,9 @@ resource "aws_iam_role" "eks_auto" {
 # Policies attached ref https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html
 resource "aws_iam_role_policy_attachment" "eks_auto" {
```
bryantbiggs: I think it's reasonable to add this. would it be possible to add this or call it out on the appropriate node roles? we follow what EKS specifies which the current code matches https://docs.aws.amazon.com/eks/latest/userguide/auto-create-node-role.html which does not have the public ECR permissions. 

cc @akestner / @tzneal

fletcherw: I have an internal PR out to update that doc page as well. If you'd like to wait to approve until that's complete, that's alright with me.

bryantbiggs (approved):
thank you

antonbabenko:
This PR is included in version 21.16.0 :tada:

PR #3676, opened 2026-04-09
feat: Expose cloudwatch log group skip_destroy argument

daniel-s-palmer:
## Description

Exposes the `skip_destroy` argument on the `aws_cloudwatch_log_group` resource.

Defaults the argument to `null` which is the current default when not providing the argument (no change by default to existing deployments).

Intentionally not updating the examples as enabling this feature will orphan resources.

## Motivation and Context

I would like to be able to set this to true in my environment.

## Breaking Changes

No breaking changes.

## How Has This Been Tested?

After updating one of the examples to set `cloudwatch_log_group_skip_destroy = true`, the plan has the following relevant change which shows `skip_destroy = true`:

```
# module.eks.aws_cloudwatch_log_group.this[0] will be created
  + resource "aws_cloudwatch_log_group" "this" {
      + arn                         = (known after apply)
      + deletion_protection_enabled = (known after apply)
      + id                          = (known after apply)
      + log_group_class             = (known after apply)
      + name                        = "/aws/eks/ex-eks-auto-mode/cluster"
      + name_prefix                 = (known after apply)
      + region                      = "us-gov-west-1"
      + retention_in_days           = 90
      + skip_destroy                = true
      + tags                        = {
          + "GithubOrg"  = "terraform-aws-modules"
          + "GithubRepo" = "terraform-aws-eks"
          + "Name"       = "/aws/eks/ex-eks-auto-mode/cluster"
          + "Test"       = "ex-eks-auto-mode"
        }
      + tags_all                    = {
          + "GithubOrg"  = "terraform-aws-modules"
          + "GithubRepo" = "terraform-aws-eks"
          + "Name"       = "/aws/eks/ex-eks-auto-mode/cluster"
          + "Test"       = "ex-eks-auto-mode"
        }
    }
    ```


bryantbiggs:
thank you but unfortunately this opens up a footgun for users - if you need this type of functionality, simple remove the resource from Terraform state before destroying the cluster (i.e. - `terraform state rm -target=...`)

daniel-s-palmer:
I had hoped that since it required the user to opt in, they would be considered responsible for the outcome :(

This is really a saving grace against developers who make mistakes in environments where those logs have requirements (government, for example) to be kept for extended periods of time.

Additionally, removing it from state does not solve the original problem. I would like to prevent accidents. Removing from state would imply the user knows they are making a mistake.

daniel-s-palmer:
@bryantbiggs can you please reconsider? The RDS module already exposes that variable as seen here https://github.com/terraform-aws-modules/terraform-aws-rds/blob/master/modules/db_instance/main.tf#L173.

daniel-s-palmer:
Looks like the vpc module also exposes this https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/vpc-flow-logs.tf#L73

daniel-s-palmer:
Are there other official modules that don't already expose this? I'm having a hard time finding one.

Perhaps we could consider adding it on the idea of greater consistency among the official modules?

daniel-s-palmer:
Interestingly enough, @bryantbiggs you merged it in for the vpc module. https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1009

bryantbiggs:
just because its in other modules doesn't mean it applies here - you can look back through history/issues/PRs - the common theme is that we have struggled with disabling the EKS created log group to ensure the Terraform created log group wins (and is controlled by IaC) in order to manage and destroy. This would be a step in the opposite direction

Also, the prior was asked by a lot of users - this is only being asked by one user. I have given you a way to work around this, we are not going to open up this footgun for users

daniel-s-palmer:
The workaround is not a real solution? Compliance is a real thing. If I remove it from state early to actually solve my problem, I am no longer able to terraform apply.

I am curious. Where is the footgun exactly? The proposed solution didn't change the default behavior. If a user is willing to dig through the variables and read what the variable does, I think they are wanting to, as you might put it, shoot themselves in the foot _and_ be glad they did it.

The user of the module _should_ be held responsible for the configuration they set! The fact that we are scared of allowing users to make a choice is sad.

It's unfortunate that we have had a change of heart over the last few years as to what we allow users to opt into with terraform.

> this is only being asked by one user

Every ask starts with the first user. Also, maybe if we left an issue open for more than 5 minutes, we might actually find other users that want the solution too?

daniel-s-palmer:
Perhaps we should open issues in the other modules to remove the `footgun` from those modules as well. Since we can't trust users to understand what the arguments do in those modules either.

daniel-s-palmer:
> just because its in other modules doesn't mean it applies here - you can look back through history/issues/PRs - the common theme is that we have struggled with disabling the EKS created log group to ensure the Terraform created log group wins (and is controlled by IaC) in order to manage and destroy. This would be a step in the opposite direction

How exactly is this a step in the opposite direction? This has nothing to do with how the log group is created or with ensuring that Terraform owns the log group on initial and subsequent applies of this module.

If we read what the [argument does from the provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group#skip_destroy-1), it orphans the log group instead of destroying it when the user runs `terraform destroy`.

This seems like a completely reasonable feature to be included in the module for a security/compliance conscientious user.

PR #3682, opened 2026-04-15
feat: Add support for standalone Pod Identity associations

shainberg:
## Summary

- Add `pod_identity_associations` variable for creating standalone `aws_eks_pod_identity_association` resources directly through the module
- Currently, pod identity associations are only supported as inline blocks within addon definitions — users who need pod identity for their own workloads must manage the resources outside the module
- Supports all resource arguments: `namespace`, `service_account`, `role_arn` (required), plus `disable_session_tags` and `target_role_arn` (optional) for session tag control and role chaining
- Follows existing module patterns (`access_entries` style): conditional creation via `local.create`, per-entry tags merged with module-level tags, ternary `for_each`

### Example usage

```hcl
module "eks" {

  # ... cluster config ...

  pod_identity_associations = {
    vpc_cni = {
      namespace       = "kube-system"
      service_account = "aws-node"
      role_arn        = "arn:aws:iam::123456789012:role/vpc-cni-role"
    }
    external_secrets = {
      namespace       = "external-secrets"
      service_account = "external-secrets-sa"
      role_arn        = "arn:aws:iam::123456789012:role/external-secrets-role"
    }
  }
}
```

## Test plan

- [x] `terraform fmt` passes on all modified files
- [x] Updated `self-managed-node-group` test fixture to exercise the new standalone path
- [x] `eks-managed-node-group` test fixture continues to cover inline addon pod identity associations
- [x] README updated with new variable and output in the TF_DOCS section

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,7 @@ We are grateful to the community for contributing bugfixes and improvements! Ple
 | [aws_eks_addon.before_compute](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon) | resource |
 | [aws_eks_addon.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon) | resource |
 | [aws_eks_cluster.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster) | resource |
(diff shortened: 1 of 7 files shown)


bryantbiggs:
thank you but we are not going to add this at this time. there are a number of ways this can be handled, I think we cover the most common including https://github.com/terraform-aws-modules/terraform-aws-eks-pod-identity

Issue #3696, opened 2026-04-30
`aws_iam_role_policy_attachment.this` missing `lifecycle { create_before_destroy = true }` in eks-managed-node-group


dsantanu:
## Description

When `kubernetes_groups` on an `aws_eks_access_entry` is changed, it triggers `assume_role_policy` on the node IAM role to become _known after apply_. This cascades to `policy_arn = (known after apply)` on `aws_iam_role_policy_attachment.this`, forcing a +/- destroy+recreate. The destroy half completes successfully, but the recreate silently fails to complete, leaving node roles with no core policies attached (`AmazonEKSWorkerNodePolicy`, `AmazonEC2ContainerRegistryReadOnly`, `AmazonEKS_CNI_IPv6_Policy`). This breaks the entire cluster — nodes lose ECR access and aws-cni fails.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Cluster impact
Nodes lose ECR pull access, aws-cni gRPC connection fails, new pods cannot be scheduled

## Versions

- Module version [Required]: 21.19.0
- Terraform version: v1.11.0
- Provider version(s): v6.23.0

## Reproduction Code [Required]

1. Have an EKS cluster managed by this module with at least one managed node group
2. Have aws_eks_access_entry resources with kubernetes_groups configured, e.g.:
```hcl
resource "aws_eks_access_entry" "this" {
  for_each          = var.eks_cluster_ops
  cluster_name      = module.cluster["eks"].cluster_name
  principal_arn     = each.value.principal_arn
  kubernetes_groups = each.value.k8s_groups
  type              = "STANDARD"
}
```
3. Change the value of `kubernetes_groups` on any access entry — e.g. rename a group from `cluster:admins` to `cluster:eksadmins`
4. Run terraform plan - observe the cascade:
  - `aws_eks_access_entry` → in-place update (expected)
  - `module.nodegroups[*].aws_iam_role.this → assume_role_policy` becomes _known after apply_
  - `module.nodegroups[*].aws_iam_role_policy_attachment.this[*]` → +/- destroy+recreate because `policy_arn = (known after apply)`
5. Run terraform apply - apply reports success with no errors
6. Result: aws_iam_role_policy_attachment.this destroy half completes, recreate half silently does not. Node roles are left with no core policies attached.
7. Verify with:
```hcl
aws iam list-attached-role-policies \
  --role-name <node-group-role-name> \
[...]

PR #3697, opened 2026-04-30
fix(eks-managed-node-group): add `create_before_destroy` to `iam_role_policy_attachment`

dsantanu:
## Description
Add `lifecycle { create_before_destroy = true }` to `aws_iam_role_policy_attachment.this` in `modules/eks-managed-node-group/iam.tf`, consistent with the same pattern already applied to `aws_launch_template.this` and `aws_eks_node_group.this` in the same module.

## Motivation and Context
When `kubernetes_groups` on an `aws_eks_access_entry` is modified, Terraform marks `assume_role_policy` on the node IAM role as `known after apply`. This cascades to `aws_iam_role_policy_attachment.this` where `policy_arn` also becomes `known after apply`, forcing a `+/-` destroy+recreate on all core node policy attachments.

The destroy half completes successfully but the recreate silently fails, leaving node IAM roles missing all core policies (`AmazonEKSWorkerNodePolicy`, `AmazonEC2ContainerRegistryReadOnly`, `AmazonEKS_CNI_IPv6_Policy`). The cluster breaks immediately — ECR image pulls fail with authorization errors and aws-cni loses gRPC connectivity, preventing new pods from being scheduled.

Terraform reports the apply as successful with no errors, making this particularly difficult to diagnose.

Issue: #3696 

## Breaking Changes
None. 
This is a non-breaking lifecycle improvement with no API or behaviour changes. Existing policy attachments are unaffected on a clean apply.

## How Has This Been Tested?
> **Note:** This fix was validated in a production EKS cluster (`v21.19.0`) by reproducing the failure condition (modifying `kubernetes_groups` on an `aws_eks_access_entry`) and confirming that with the `lifecycle` block in place the policy attachments are recreated before being destroyed, leaving no gap in node IAM permissions.

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -622,6 +622,10 @@ resource "aws_iam_role_policy_attachment" "this" {
 
   policy_arn = each.value
   role       = aws_iam_role.this[0].name
+
+  lifecycle {
+    create_before_destroy = true
+  }
 }
 
 resource "aws_iam_role_policy_attachment" "additional" {


bryantbiggs:
lets review in the issue first - this does not make sense

PR #3706, opened 2026-05-12
feat(examples): Add Cilium CNI + Karpenter example with ENI native ro…

sushanku:
## Description
Adds a new example `examples/karpenter-cilium` demonstrating how to run
Cilium as the sole CNI on an EKS cluster managed by Karpenter, solving
the CNI bootstrap ordering problem entirely within Terraform — no manual
kubectl steps required.

Cilium is configured in ENI native routing mode with kubeProxyReplacement
enabled, Hubble observability, and Maglev load balancing. Karpenter handles
node autoprovisioning with the Cilium not-ready taint applied automatically
to all nodes.

## Motivation and Context
When replacing the default vpc-cni with Cilium on EKS, there is a
well-known chicken-and-egg problem: nodes will not reach Ready state without
a CNI binary present, but Cilium cannot be scheduled without Ready nodes.
Existing workarounds require a manual kubectl patch of the aws-node DaemonSet,
breaking automated provisioning.

This example resolves the ordering entirely in Terraform using three mechanisms:
- helm_release.cilium depends only on the control plane with wait = false,
  so Cilium is registered before any node joins
- The managed node group applies node.cilium.io/agent-not-ready:NoExecute
  taint, preventing workload scheduling until the Cilium agent is ready
- Cilium nodeInit writes the CNI binary to the node filesystem at bootstrap,
  so nodes become Ready immediately after the agent starts
  
## Breaking Changes
None — additive example only.

## How Has This Been Tested?

Tested on EKS 1.35 / Cilium 1.16.x / Karpenter 1.6.0. Full apply/destroy
cycle completed successfully. Cilium status healthy, kube-proxy replaced,
Hubble relay running, Karpenter provisioning nodes correctly.

### DIFF
--- a/examples/cilium-with-karpenter/README.md
+++ b/examples/cilium-with-karpenter/README.md
@@ -1,0 +1,91 @@
+# Cilium + Karpenter Example
+
+Configuration in this directory creates an AWS EKS cluster with [Cilium](https://cilium.io/) as the CNI and [Karpenter](https://karpenter.sh/) for node provisioning.
+
+## Solving the CNI Bootstrap (Chicken-and-Egg) Problem
+
+EKS nodes will not reach `Ready` state until a CNI plugin is present on the node. Normally, `vpc-cni` handles this automatically. When replacing it with Cilium, a bootstrap ordering problem arises:
+
+- Cilium must be installed **before** nodes join, so nodes have a CNI binary on startup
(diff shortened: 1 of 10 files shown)


bryantbiggs:
thank you but this is not something we are planning to take on in this project - that looks more like an example use case, sometimes called a "blueprint" these days, which we do not store in this repo

PR #3714, opened 2026-05-20
feat: Add support for launch template `network_performance_options`

alien2003:
## Description

- Add support for the launch template `network_performance_options` block on `eks-managed-node-group` and `self-managed-node-group`, threaded through the root module's `eks_managed_node_groups` and `self_managed_node_groups` maps.
- The block exposes `bandwidth_weighting`, which lets users on supported instance families (M7i / R7i / C7i and successors) bias bandwidth between network and EBS at launch (`default` / `vpc-1` / `ebs-1`).
- Mirrors the existing `cpu_options` shape: typed optional object, single `dynamic` block in `aws_launch_template.this`, default `null` so the block is omitted unless set. No validation in the module since the provider already enforces the enum.

## Motivation and Context

- `aws_launch_template.network_performance_options` was added in AWS provider [v6.29.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v6.29.0). The module already requires `>= 6.42`, so no version bump is needed.
- API reference: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateNetworkPerformanceOptionsRequest.html
- Today users running EKS workloads that need to bias bandwidth on supported instance types have to drop the module and roll their own launch template. This wires the field through using the same shape the module already uses for `cpu_options`, `credit_specification`, etc.

## Breaking Changes

- No. Strictly additive: the new variable is `optional(object({ bandwidth_weighting = optional(string) }))` defaulting to `null`, so the dynamic block emits nothing when unset.

## How Has This Been Tested?

`pre-commit run -a` clean (terraform_fmt, terraform_docs, terraform_tflint, terraform_validate, end-of-file-fixer, trailing-whitespace, mixed-line-ending) across the root module and every directory under `modules/`, `examples/`, `tests/`.


bryantbiggs:
please fix the conflicts when you get a chance

alien2003:
Rebased on master and resolved the conflicts. Since master already requires AWS provider `>= 6.42`, the original `versions.tf` bump is no longer needed, so the diff is now just the variable, the `dynamic` block in each node group, and the regenerated docs.

bryantbiggs (approved):

antonbabenko:
This PR is included in version 21.23.0 :tada:

PR #3719, opened 2026-05-26
feat: Add support for launch template `secondary_interfaces`

alien2003:
## Description

Adds a `secondary_interfaces` argument to the `eks-managed-node-group` and `self-managed-node-group` sub-modules, and the corresponding `eks_managed_node_groups` / `self_managed_node_groups` maps on the root module. Each element renders a `secondary_interfaces` block on `aws_launch_template`, mirroring the existing `network_interfaces` passthrough.

## Motivation and Context

`aws_launch_template` gained the `secondary_interfaces` block in AWS provider `v6.33.0` (the module already requires `>= 6.42`). It is used to attach additional ENIs at instance launch and was not previously exposed by the module, so users needing secondary interfaces had to manage their own launch template. The argument is exposed symmetrically on both node-group types, consistent with the rest of the launch template surface.

## Breaking Changes

None. New optional argument that defaults to `[]`, so no block is rendered unless set.

## How Has This Been Tested?

`terraform_fmt`, `terraform_docs`, `terraform_tflint`, and `terraform_validate` were run for the root module and both node-group sub-modules using the pinned `pre-commit-terraform` `v1.105.0` toolchain.


bryantbiggs:

Issue #3677, opened 2026-04-10
support EC2 warm pools for EKS managed node groups


maximeroma:
## Is your request related to a new offering from AWS?

Yes ✅

This request is related to the recent AWS announcement introducing **EKS Managed Node Groups EC2 Warm Pools**:
https://aws.amazon.com/fr/about-aws/whats-new/2026/04/amazon-eks-managed-node-groups-ec2-warm-pools/

⚠️ At the time of writing, this functionality is **not yet available in the AWS Terraform provider**.
Tracking issue: https://github.com/hashicorp/terraform-provider-aws/issues/47389

This feature request is **blocked** until it is implemented in the provider.

---

## Is your request related to a problem? Please describe.

When using EKS managed node groups, scaling events can introduce latency due to instance provisioning time.

AWS recently introduced **EC2 Warm Pools for EKS Managed Node Groups**, which significantly reduces scale-up time by keeping pre-initialized instances ready.

---

## Describe the solution you'd like.

Add support for configuring **EC2 Warm Pools in EKS Managed Node Groups** within the module, once it is supported by the AWS Terraform provider.

This would likely include:

* New input variables to configure warm pool settings (e.g., min/max size, pool state, reuse policy, etc.)
* Integration within `eks_managed_node_groups`
* Proper defaults to maintain backward compatibility

---

## Additional context

This feature request is **forward-looking** and should be implemented once the AWS provider supports it.

Tracking upstream dependency:
https://github.com/hashicorp/terraform-provider-aws/issues/47389

AWS announcement:
https://aws.amazon.com/fr/about-aws/whats-new/2026/04/amazon-eks-managed-node-groups-ec2-warm-pools/

maximeroma:
not stale

skmannion:
not stale

jameslanska-du:
[v6.56.0 of the Terraform AWS Provider](https://github.com/hashicorp/terraform-provider-aws/blob/v6.56.0/CHANGELOG.md) added the necessary functionality.

jameslanska-du:
not stale

PR #3412, opened 2025-07-09
feat!: Upgrade min AWS provider and Terraform versions to `6.0` and `1.5.7` respectively

bryantbiggs:
## List of backwards incompatible changes

- Terraform `v1.5.7` is now minimum supported version
- AWS provider `v6.0.0` is now minimum supported version
- TLS provider `v4.0.0` is now minimum supported version
- The `aws-auth` sub-module has been removed. Users who wish to utilize its functionality can continue to do so by specifying a `v20.x` version, or `~> v20.0` version constraint in their module source.
- `bootstrap_self_managed_addons` is now hardcoded to `false`. This is a legacy setting and instead users should utilize the EKS addons API, which is what this module does by default. In conjunction with this change, the `bootstrap_self_managed_addons` is now ignored by the module to aid in upgrading without disruption (otherwise it would require cluster re-creation).
- When enabling `enable_efa_support` or creating placement groups within a node group, users must now specify the correct `subnet_ids`; the module no longer tries to automatically select a suitable subnet.
- EKS managed node group:
    - IMDS now default to a hop limit of 1 (previously was 2)
    - `ami_type` now defaults to `AL2023_x86_64_STANDARD`
    - `enable_monitoring` is now set to `false` by default
    - `enable_efa_only` is now set to `true` by default
    - `use_latest_ami_release_version` is now set to `true` by default
    - Support for autoscaling group schedules has been removed
- Self-managed node group:
    - IMDS now default to a hop limit of 1 (previously was 2)
    - `ami_type` now defaults to `AL2023_x86_64_STANDARD`
    - `enable_monitoring` is now set to `false` by default
    - `enable_efa_only` is now set to `true` by default
    - Support for autoscaling group schedules has been removed
- Karpenter:
    - Native support for IAM roles for service accounts (IRSA) has been removed; EKS Pod Identity is now enabled by default
    - Karpenter controller policy for prior to Karpenter `v1` have been removed (i.e. `v0.33`); the `v1` policy is now used by default
    - `create_pod_identity_association` is now set to `true` by default
- `addons.resolve_conflicts_on_create` is now set to `"NONE"` by default (was `"OVERWRITE"`).
- `addons.most_recent` is now set to `true` by default (was `false`).
- `cluster_identity_providers.issuer_url` is now required to be set by users; the prior incorrect default has been removed. See https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3055 and https://github.com/kubernetes/kubernetes/pull/123561 for more details.
[...]

PR #1310, opened 2021-04-19
feat: Add support for ASG Warm Pools

sppwf:
# PR o'clock

## Description

AWS Released lately [ASG Warm Pool](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-warm-pools.html) feature that will ensure that instances are ready to quickly start serving application traffic, accelerating the response to a scale-out event.

### Checklist

- [X] CI tests are passing
- [ ] README.md has been updated after any changes to variables and outputs. See https://github.com/terraform-aws-modules/terraform-aws-eks/#doc-generation

PR #3720, opened 2026-05-26
feat: Add support for self-managed node group `warm_pool`

alien2003:
## Description

- Re-add the `warm_pool` block on `self-managed-node-group`, threaded through the root module's `self_managed_node_groups` map, along with the companion `force_delete_warm_pool` argument.
- A warm pool keeps pre-initialized EC2 instances in a `Stopped`, `Hibernated`, or `Running` state so the Auto Scaling group scales out without waiting for instance launch and node bootstrap. The block exposes `pool_state`, `min_size`, `max_group_prepared_capacity`, and a nested `instance_reuse_policy { reuse_on_scale_in }`.
- Mirrors the existing `instance_refresh` / `mixed_instances_policy` shape: typed optional object, single `dynamic` block in `aws_autoscaling_group.this`, default `null` so the block is omitted unless set. No validation in the module since the provider already enforces the enums.

## Motivation and Context

- `warm_pool` and `force_delete_warm_pool` were supported on self-managed node groups before v21 (added in #1310). They were dropped in #3412 when the Auto Scaling group variables were converted from `any` to concrete object types: the sibling blocks (`instance_refresh`, `mixed_instances_policy`, `instance_maintenance_policy`) were converted and kept, but `warm_pool` was not. This restores the capability, typed to match those siblings.
- Warm pools apply to self-managed node groups, which own their Auto Scaling group directly. EKS managed node groups do not expose warm pools through the EKS API (see #3677), so this is independent of that request.
- API reference: https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-warm-pools.html

## Breaking Changes

- No. Strictly additive: `warm_pool` is `optional(object({...}))` defaulting to `null` (the dynamic block emits nothing when unset) and `force_delete_warm_pool` is `optional(bool)` defaulting to `null`. No resource address changes, so no state migration.

## How Has This Been Tested?

Ran the v1.105.0 hooks (terraform_fmt, terraform_docs, terraform_tflint, terraform_validate) against the root module and `modules/self-managed-node-group`; all clean. A live apply to observe warm pool instances is deferred to maintainer testing, consistent with how the sibling ASG blocks were added.


bryantbiggs:

PR #3721, opened 2026-05-26
feat: Add support for self-managed node group `default_cooldown`, `service_linked_role_arn` and `wait_for_capacity_timeout`

alien2003:
## Description

- Restore three `aws_autoscaling_group` arguments on `self-managed-node-group`, threaded through the root module's `self_managed_node_groups` map:
  - `default_cooldown` (`number`) - seconds after a scaling activity completes before another can start.
  - `service_linked_role_arn` (`string`) - service-linked role the Auto Scaling group uses to call other AWS services.
  - `wait_for_capacity_timeout` (`string`) - how long Terraform waits for instances to be healthy; `"0"` skips capacity waiting.
- All three are `optional`, default `null`, so an unset value passes through to the provider default and existing configurations are unchanged.

## Motivation and Context

- These arguments were supported on self-managed node groups before v21. They were dropped in #3412 when the Auto Scaling group variables were converted from `any` to concrete object types: most blocks were converted and kept, but these scalar arguments were removed rather than carried over. This restores them, typed and documented to match the surrounding variables.
- Scope is limited to arguments that apply to any self-managed node group. The other arguments removed in #3412 are intentionally left out: `min_elb_capacity` / `wait_for_elb_capacity` / `target_group_arns` require load balancer attachment that the module does not expose; `elastic_gpu_specifications` / `elastic_inference_accelerator` are deprecated by AWS; `platform`, `cluster_version` and `delete_timeout` were renamed/migrated, not dropped.

## Breaking Changes

- No. Strictly additive: each new variable is `optional` and defaults to `null`, so the rendered `aws_autoscaling_group` is identical to today's for any configuration that does not set them. No resource address changes, so no state migration.

## How Has This Been Tested?

Ran the v1.105.0 hooks (terraform_fmt, terraform_docs, terraform_tflint, terraform_validate) against the root module and `modules/self-managed-node-group`; all clean.


bryantbiggs:

PR #3725, opened 2026-06-23
feat: Expose before_compute addons as a separate output

vchepkov:
## Description
Add `cluster_addons_before_compute` output exposing `aws_eks_addon.before_compute` on its own.

## Motivation and Context
The only addon output today is `cluster_addons`, which merges the before-compute and post-compute addons. Consumers running VPC CNI custom networking need to anchor ENIConfig creation on vpc-cni readiness, but referencing the merged output pulls in the post-compute addons, which depend on the node groups, which under custom networking cannot become Ready until ENIConfig exists, creating a dependency cycle.

## Breaking Changes
Purely additive; no behavior change, no new resources


vchepkov:
I would appreciate any feedback if there's a better approach to solving this.
As far as I can tell, the only way to guarantee the required dependency ordering today is to manage the add-ons and node groups outside the module. That works, but it's considerably less convenient and reduces the value of the module.
The dependency chain I'm trying to enforce is:
```
cluster -> vpc-cni -> ENIConfigs -> node group -> coredns/kube-proxy/extra
```
If there's another way to achieve this within the module, I'd be very interested to learn about it.

PR #3729, opened 2026-06-25
fix: Ignore desired_capacity_type drift on self-managed node group ASG

### LINKED ISSUE #3723: self-managed-node-group: desired_capacity_type drift can break apply (terraform-provider-aws#37523)
Hi! Thanks for maintaining this module — filing this so other folks who hit it can find a pointer here.

## Summary

The `aws_autoscaling_group` in `modules/self-managed-node-group` can get caught by [hashicorp/terraform-provider-aws#37523](https://github.com/hashicorp/terraform-provider-aws/issues/37523): when `desired_size_type` (→ `desired_capacity_type`) is left unset and AWS populates the field server-side (e.g. after a console-driven capacity change, or when `instance_requirements` is in use), subsequent applies plan `desired_capacity_type = "units" -> null` and the API rejects the update with:

```
api error ValidationError: 1 validation error detected:
Value '' at 'desiredCapacityType' failed to satisfy constraint:
Member must have length greater than or equal to 1
```

The module already does `ignore_changes = [desired_capacity]` on the ASG (lifecycle block on `aws_autoscaling_group.this` in `modules/self-managed-node-group/main.tf`), so the precedent for ignoring server-side capacity drift is already there.

## Reproduction

A minimal, no-modules reproduction on just the AWS resource is in [hashicorp/terraform-provider-aws#37523](https://github.com/hashicorp/terraform-provider-aws/issues/37523) (confirmed on AWS provider 5.49.0 and 6.42.0). Triggered via this module by:

1. Apply a `self-managed-node-group` config without setting `desired_size_type`.
2. Out-of-band, change desired capacity in the AWS console (or set `DesiredCapacityType` via the CLI).
3. Re-apply — fails with the validation error above.

## Possible fixes

Both would resolve it; happy to defer to whichever you prefer:

1. Add `desired_capacity_type` to the existing `ignore_changes` on the ASG. Broader than strictly needed but matches the precedent set for `desired_capacity`.
2. Default `desired_size_type` to `"units"` when unset, matching the AWS API default for ASGs that don't use attribute-based instance selection.

Glad to open a PR for whichever direction you'd prefer — just let me know. Thanks again!

somaz94:
## Description

When `desired_size_type` (→ `desired_capacity_type`) is left unset, AWS populates it
server-side (e.g. after a console-driven capacity change, or when `instance_requirements`
is in use). The next apply then plans `desired_capacity_type = "units" -> null`, which the
API rejects:

    ValidationError: Value '' at 'desiredCapacityType' failed to satisfy constraint:
    Member must have length greater than or equal to 1

This adds `desired_capacity_type` to the existing `ignore_changes` on
`aws_autoscaling_group.this` in `self-managed-node-group`, matching the precedent already
set for `desired_capacity` directly above it.

Fixes #3723. The root cause is the provider bug hashicorp/terraform-provider-aws#37523;
this is a module-level mitigation that does not depend on the provider fix landing.

The issue also suggested option 2 (defaulting `desired_size_type` to "units" when unset).
This PR takes option 1 because it follows the in-module `desired_capacity` precedent and is
backward compatible. Happy to switch if you prefer option 2.

## Breaking Changes

No.

## How Has This Been Tested?

- [x] `terraform fmt` and `terraform validate` pass
- [x] `tflint` clean

### DIFF
--- a/modules/self-managed-node-group/main.tf
+++ b/modules/self-managed-node-group/main.tf
@@ -807,7 +807,8 @@ resource "aws_autoscaling_group" "this" {
   lifecycle {
     create_before_destroy = true
     ignore_changes = [
-      desired_capacity
+      desired_capacity,
+      desired_capacity_type,
     ]
   }
 }


somaz94:
Still relevant — this fixes #3723, where `desired_capacity_type` drift causes a perpetual diff on self-managed node group ASGs. CI is green and there are no conflicts; it's waiting on maintainer review only.

bryantbiggs:
this is a provider issue and the fix will come from there

PR #3730, opened 2026-07-03
feat: Add enable_efa_placement_group to decouple EFA from Cluster Placement Group

flozonn:
## Description

When `enable_efa_support = true`, the module unconditionally creates a Cluster Placement Group via:

```hcl
create_placement_group = var.create && (local.enable_efa_support || var.create_placement_group)
```

This causes `InsufficientInstanceCapacity` errors when launching instances against **On-Demand Capacity Reservations (ODCRs)** or **Capacity Blocks** that were not provisioned with a matching placement group. The ODCR is active and allocated, but the RunInstances request includes a CPG constraint that doesn't match the reservation, so EC2 cannot guarantee capacity delivery.

## Problem

Customers using GPU instances (e.g., p6-b200.48xlarge) with:
- EFA enabled (for high-bandwidth networking)
- ODCR or Capacity Blocks (for guaranteed GPU capacity)

Hit ICE errors because the module forces a CPG that the reservation doesn't have.

## Solution

Add a new variable `enable_efa_placement_group` (default: `true` for backward compatibility) that allows users to enable EFA without forcing a Cluster Placement Group:

```hcl
create_placement_group = var.create && ((local.enable_efa_support && var.enable_efa_placement_group) || var.create_placement_group)
```

### Usage

```hcl
module "eks_managed_node_group" {
  # Enable EFA without Cluster Placement Group (for ODCR compatibility)
  enable_efa_support         = true
  enable_efa_placement_group = false
}
```

## Changes

- `modules/eks-managed-node-group/main.tf` — Decouple EFA from CPG creation
- `modules/eks-managed-node-group/variables.tf` — Add `enable_efa_placement_group` variable
- `modules/self-managed-node-group/main.tf` — Same fix
- `modules/self-managed-node-group/variables.tf` — Same variable
- `node_groups.tf` — Passthrough for both module types
- `variables.tf` — Add to root module type definitions

## Backward Compatibility

✅ Fully backward compatible. Default is `true`, preserving existing behavior. Only customers who explicitly set `enable_efa_placement_group = false` will see changed behavior.

## Real-World Use Case

Customer with 4x p6-b200.48xlarge ODCR in us-east-2b couldn't launch their 4th GPU instance because the EKS Terraform module automatically added a CPG that didn't match the ODCR. EC2 on-call confirmed: "User has a placement group constraint, only reservations for placement group X are guaranteed. Insufficient capacity within placement group constraint."

### DIFF
--- a/modules/eks-managed-node-group/main.tf
+++ b/modules/eks-managed-node-group/main.tf
@@ -707,7 +707,7 @@ resource "aws_iam_role_policy" "this" {
 ################################################################################
 
 locals {
-  create_placement_group = var.create && (local.enable_efa_support || var.create_placement_group)
+  create_placement_group = var.create && ((local.enable_efa_support && var.enable_efa_placement_group) || var.create_placement_group)
 }
 
 resource "aws_placement_group" "this" {
--- a/modules/eks-managed-node-group/variables.tf
+++ b/modules/eks-managed-node-group/variables.tf
@@ -420,6 +420,13 @@ variable "create_placement_group" {
   nullable    = false
 }
 
+variable "enable_efa_placement_group" {
+  description = "Determines whether to automatically create a Cluster Placement Group when EFA is enabled. Set to `false` when using On-Demand Capacity Reservations (ODCRs) or Capacity Blocks that are not configured with a placement group"
+  type        = bool
+  default     = true
+  nullable    = false
+}
+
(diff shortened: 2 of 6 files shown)


dwo:
relates to: #3661

dwo:
not stale

dwo:
not stale

PR #3748, opened 2026-09-18
fix: Document known Terraform/OpenTofu limitations in README

antonbabenko:
## Description

Adds a "Known limitations (Terraform/OpenTofu, not this module)" section above Examples.
It covers requests that come up repeatedly but cannot be implemented by any module,
because `lifecycle` arguments must be literal values inside the resource block
(hashicorp/terraform#18367, hashicorp/terraform#22544, opentofu/opentofu#1329), and names
the native workaround for each one first.

The section sits between `<!-- BEGIN_KNOWN_LIMITATIONS -->` and
`<!-- END_KNOWN_LIMITATIONS -->` so it can be updated in place later.

## Breaking Changes

None. README only.

Disclosure: I maintain this module and also build compliance.tf, which the section links to.


antonbabenko:
This PR is included in version 21.25.1 :tada:

PR #3749, opened 2026-09-19
fix: Honor self-managed launch template network bandwidth requirements

nbaliyan260:
## Description

Pass `instance_requirements.network_bandwidth_gbps` through to the self-managed node group's `aws_launch_template`, preserving its optional `min` and `max` values. Extend the existing attribute-based instance selection fixture in `tests/self-managed-node-group` with both bounds.

## Motivation and Context

Both the root module and the self-managed node-group submodule already accept this option, but the launch template omits the nested block. For example, `network_bandwidth_gbps = { min = 1, max = 10 }` currently produces a launch-template plan with no bandwidth requirement. The same option is already handled for mixed-instance-policy overrides.

This change makes the launch-template behavior match the existing input contract and [AWS provider schema](https://registry.terraform.io/providers/hashicorp/aws/6.59.0/docs/resources/launch_template#instance_requirements).

## Breaking Changes

No input or default changes. When bandwidth requirements are already configured, the next plan will correctly include them in a new launch-template version.

## How Has This Been Tested?

The existing `tests/self-managed-node-group` fixture was updated. No AWS deployment was performed.

- Ran credential-free, local Terraform plans with AWS provider 6.59.0 against the unmodified baseline and the patched module. The baseline dropped all supplied bandwidth blocks; the fixed module passes six cases: both bounds, minimum only, maximum only, an empty block, an omitted block, and omitted instance requirements. Existing memory, vCPU, and network-interface requirements remain unchanged.
- Repeated all six passing cases with Terraform 1.16.3 and the minimum supported Terraform 1.5.7.
- `terraform validate` passes for the local regression fixture on both Terraform versions.
- Formatting and targeted TFLint checks pass.
- All repository pre-commit hooks pass. The first all-files run passed formatting, docs, TFLint and whitespace checks, but encountered TLS-provider startup failures during parallel initialization. Rerunning `terraform_validate` across all files with parallelism limited to one passed without code changes.


bryantbiggs (approved):

antonbabenko:
This PR is included in version 21.25.2 :tada:

